PR #1, opened 2018-01-15
Split modm into many submodules and refactor header includes

salkinium:
I've rebased some of your unittest commits and merged xpcc/2017q4 and updated the copyright headers. This is now the develop branch.

However, this change adds almost 100 submodules by splitting up various large modules into small ones. Then I updated all dependecies by grepping the header files and updated header include where they were wrong, or unecessary. This is a fairly straight forward, mechanical step.

However, there are a few special cases:
1. I removed the dependency of all `:platform:**` modules on `:platform:core`, and instead replaced it with `:cmsis:device`. Nothing depends on `:platform:core` anymore now, which is intentional, so that in the future, it becomes easier to merge xpcc with other libraries (by using `cubemx:platform:core` for example).
2. The `:architecture:**` modules closely map onto the `modm/architecture/interface` on purpose, so that `:driver:**` modules can specify more precisely what they need. This prevents the `:architecture` from dragging in all other modules, due to the interfaces dependencies.
3. Some interfaces delegate implementation details to `:platform:core`. This is not modelled as a dependency on purpose, to allow the user to choose the `:platform:core` implementation. Instead the `:platform:core` implementation depends on most `:architecture:**` modules, specifically those that it implements.
4. Holy cow, do we have a shit ton of device drivers.
5. Unittests have been updated and compile (on my machine!1!!).
6. Examples are still WIP, and expect lotsa `fixup!` commits.

This change depends on having multiple `*.lb` files in the same directory (as [enabled here](https://github.com/modm-io/modm/commit/e17b379f00e47ed32421220f0378f55da219a3a9#diff-752cbade133a050392dd10c7626fac08R50)), which depends on [this change to library-builder](https://github.com/dergraaf/library-builder/pull/21).

This PR is for you, @dergraaf, to get an overview of what I'm doing to modm. Please review and comment.


dergraaf:
Very nice!

> I removed the dependency of all :platform:** modules on :platform:core, and instead replaced it with :cmsis:device. Nothing depends on :platform:core anymore now, which is intentional, so that in the future, it becomes easier to merge xpcc with other libraries (by using cubemx:platform:core for example).

But `cmsis:device` only works for Cortex devices? I don't understand how that makes merging with other libraries easier.

> Unittests have been updated and compile (on my machine!1!!).

On mine too :smile:

salkinium:
> But `cmsis:device` only works for Cortex devices?

Yes, and only `:platform:**` modules on Cortex-M devices depend on `:cmsis:device` (ie. all STM32 modules, but none of the AVR/Hosted ones). This dependency is actually more correct than `:platform:core`, since it generates `<modm/platform/device.hpp>` which all peripheral drivers include for the IO definitions.

> I don't understand how that makes merging with other libraries easier.

One use case is using the CubeHALs Ethernet drivers, because we don't have any. AFAIK you can still use your own code that accesses the peripherals directly _as long as_ you don't also use CubeHAL in parallel. So the idea is to provide a "shim" `cubemx:platform:core` module that boots up in a way that makes both modm and CubeHAL work together. I have no idea if that's really possible, but at least you can try without having to rewrite half the modules dependencies.

salkinium:
Fixed up some issues during the merge that broke `tools/system_design`. Now all examples compile even with SCons 3! Apparently the scons tools from DLR work fine with SCons 3.

salkinium:
The last really important features I want to add to modm are these:
- [x] Update Readme
- [ ] Get Circle CI running all examples again for modm
- [ ] Port xpcc.io website to modm.io and update instructions

See also [the complete list here](https://github.com/orgs/modm-io/teams/maintainers/discussions).

PR #5, opened 2018-04-07
[xpcc] Acknowledge responses based on entry list

salkinium:
When using the DynamicPostman for sending requests but without ever
registering an action with it, `Postman::isComponentAvailable()` always
returns `false`. The Dispatcher then cannot acknowledge any responses,
and they time out on the remote component.
This has been changed to also check if the response is expected in the entry list.

### DIFF
--- a/src/modm/communication/xpcc/dispatcher.cpp
+++ b/src/modm/communication/xpcc/dispatcher.cpp
@@ -44,13 +44,12 @@ xpcc::Dispatcher::update()
 		}
 		else
 		{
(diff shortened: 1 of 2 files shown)


salkinium:
@dergraaf, @georgi-g do you ever remember issues with DynamicPostman regarding unnecessary resends? If a component only registers Events and also sends Requests, then those requests will silently timeout, even if the request was successfully received.

The unit tests are a bit weird, they find it ok, that a response without a matching request should also be acknowledged?

PR #6, opened 2018-04-08
[system_design] Implement array types for C++

salkinium:
This adds a C++ implementation for arrays in packet definitions.
The constructor uses a `memcpy` analog to how all other constructors copy the data.
Maybe I need to implement the copy constructor as well for when a packet with arrays is used within another packet?

cc @dergraaf @georgi-g


salkinium:
I'm just going to merge this even if the Java generators aren't implemented. I use this for ELVA. In the long run, I want to replace XPCC with [UAVCAN](http://uavcan.org).

PR #17, opened 2018-05-18
Add a GPIO Group to STM32 GPIO

strongly-typed:
How and where should a test to the CI be added?

### DIFF
--- a/examples/generic/blinky/project.xml
+++ b/examples/generic/blinky/project.xml
@@ -1,6 +1,27 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <library>
+  <!-- <extends>../../../src/modm/board/al_avreb_can/board.xml</extends> -->
+  <!-- <extends>../../../src/modm/board/arduino_uno/board.xml</extends> -->
+  <!-- <extends>../../../src/modm/board/blue_pill/board.xml</extends> -->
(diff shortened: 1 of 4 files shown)


salkinium:
See http://modm.io/guide/testing/.
You can just add it in the generic folder and it'll just get compiled, just like before. Manual call:
```sh
 cd examples
 ../tools/scripts/examples_compile.sh generic
```
I plan to compile the generic examples for all boards at least, but that's not been implemented yet.

salkinium:
I don't agree with this PR on the point that the canonical interface of this is `modm::platform::SoftwareGpioPort`.  I would rather see that improved to merge port access in a generic way than introduce such a specific new interface and implementation.
I'd recommend you keep this in your own code base/fork for now while I check how difficult it is to implement this generically in `SoftwareGpioPort`.

salkinium:
> while I check how difficult it is to implement this generically in SoftwareGpioPort.

This is fairly easy to implement with C++17 fold expressions, which in combination with the relaxed constexpr function rules, flattens the recursive requirement of variadic arguments. constexpr if makes the resulting code gen very efficient. Is using C++17 an issue for your use-case?

salkinium:
I'm closing this in favor of #19.

PR #24, opened 2018-05-24
[CI] Run Gitlab CI in parallel.

strongly-typed:
Project is mirrored from Github to Gitlab to use Gitlab's free CI.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -68,66 +68,50 @@ build:linux-examples:
     - (cd examples && ../tools/scripts/examples_compile.py linux zmq)
 
 
-build:avr-compile-all:
+.build:compile-all: &job_definition
   stage: build
   script:
-    - (cd test/all && python3 run_all.py at)
+    - (cd test/all && python3 run_all.py $SERIES)
   artifacts:
     paths:
       - test/all/log
+
+build:avr-compile-all:
+  variables:
+    SERIES: at
+  <<: *job_definition
 
 build:stm32f0-compile-all:
-  stage: build
-  script:
-    - (cd test/all && python3 run_all.py stm32f0)
-  artifacts:
-    paths:
-      - test/all/log
+  variables:
+    SERIES: stm32f0
+  <<: *job_definition
 
 build:stm32f1-compile-all:
-  stage: build
-  script:
-    - (cd test/all && python3 run_all.py stm32f1)
-  artifacts:
-    paths:
-      - test/all/log
+  variables:
+    SERIES: stm32f1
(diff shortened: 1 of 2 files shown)


salkinium:
But why?

salkinium:
> But why?

I'm serious, I don't want to maintain this too, if there isn't some good benefit from it.

strongly-typed:
I see your point. Previous CIs caused headaches on a regular basis. For now, I think there is not much maintenance effort as all the magic happens inside the docker container by the Python scripts. No more dependency tracking in the CI. I will take care of it.

I / we / RCA run a private Gitlab instance with its CI enabled and for our internal builds this is very convenient to run on the same platform for internal PRs and branches.

salkinium:
> I / we / RCA run a private Gitlab instance with its CI enabled and for our internal builds this is very convenient to run on the same platform for internal PRs and branches.

Ok, so modm providing a "DIY GitLab CI"-template is a completely new feature then. We've never made the CI part explicitly ship with modm, only the tests.

What kind of guarantees do we want to give for this? Do we want to keep the two in sync? New features may be added to CircleCI but not to GitLab CI? Do we have a final CircleCI job that compares its own output with that of GitLabCI?

I don't understand how $human can maintain this with reasonable effort.
We have a GitLab template in xpcc too, it sits there just unmaintained.
I'm certainly not manually comparing the CircleCI output with the GitLab CI output, not even if I get a cute 🐈 for every comparison 😛.

strongly-typed:
I just would drop CircleCI ;-). To me, the Gitlab CI is capable of much more fun things.

Like having a well defined pipeline, with dependencies and parallel jobs. This is very cool for proper staging: lint, build, test, push to staging. And a magical manual button to push it to production if the pipeline is all green.

The guarantees are: you get what you pay for 😂
My expectations is that the code from develop always compiles cleanly with the CI and the tests are green in the CI. If it does not run there, how would it run on my machine? This is currently enforced by the merge settings of Github.

I do not see the need to compare the outputs of both CIs. They are self-contained. What would you compare?

salkinium:
Hm, ok, in the long run, if we want to at least run the STM32 unit test on actual hardware, then I'd have to run a GitLab instance either in the RCA or at home *anyways*. So maybe dropping CircleCI on favor or GitLab CI would be the best option.

I don't want to have a duplicate CIs visible in this project, I think it's just confusing to see two identical jobs and then not knowing which one leads to CircleCI or GitLab. So for now I'd like to stick with CircleCI (over which I have full control too), until I (or someone else) can figure out how to integrate local hardware with the CI.

PR #25, opened 2018-05-24
EXTI warning, and improve bitmask readability.

strongly-typed:
Maybe `connect` would be an even better interface for that?

### DIFF
--- a/src/modm/platform/gpio/stm32/pin.hpp.in
+++ b/src/modm/platform/gpio/stm32/pin.hpp.in
@@ -137,6 +137,9 @@ public:
 		setInput();
 	}
 	// External Interrupts
+	// Warning: This will disable any previously enabled interrupt which is
+	// routed to the same interupt line, e.g. PA3 will disable PB3.
+	// This is a hardware limitation by the STM32 EXTI.
 	inline static void
 	enableExternalInterrupt()
 	{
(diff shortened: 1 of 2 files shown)


salkinium:
> Maybe connect would be an even better interface for that?

Probably something like this.
```cpp
Exti::connect<GpioB3::Exti, GpioB4::Exti>();
Exti::connect<GpioB3::Exti, GpioC3::Exti>(); // static assertion failure
```

It may not prevent programming mistakes, since there is no way to keep state at compile time in C++ TMP. This is perfectly valid code, and there's nothing I can do about it.
```cpp
Exti::connect<GpioB3::Exti>(); // fine
Exti::connect<GpioC3::Exti>(); // also fine, but will write over previous assignment
```

This is an inherent design issue, the compile time obviously cannot enfore runtime behavior.
However, we can read out the `SYSCFG->EXTICR` register and `modm_assert_debug()` that it's not being overwritten (for debug builds at least). That's the best I can do in-code.
An external tool that parses the code (or gets configured like CubeMX) and therefore knows global state can do a lot more of course.

salkinium:
Of course you can always use this compiler specific hack for keeping state at compile time:
https://blog.tartanllama.xyz/integral_variable/

salkinium (approved):
Thanks!

strongly-typed:
> Probably something like this.

Maybe more like this

```c++
class MyIrqHandler : public ExtInterruptHandler
{
   static void
   do() // naming by convention, not by configuration
   {
      // do something

      // Acknowledgement of flags is done automatically
   };
};

GpioB3::setInput(Gpio::InputType::PullUp);
GpioC4::setInput(Gpio::InputType::PullUp);

// Connect external events to the call of a method.
MyIrqHandler::connect<GpioB3::Exti::RisingEdge, GpioC4::Exti::BothEdges>();
```

This involves many magic, e.g. adding a proxy between the `EXTI3` and `EXTI4` to call `MyIrqHandler::do` as `GpioB3` and `GpioC4` are not connected to the same interrupt line.
That's all very hardware-specific to STM32 and I did not look into other Cortex-Chips how there external interrupt hardware was designed.

I know that there were some efforts but they were not successful because of the lack of a good model of interrupts in the C++ language.

PR #26, opened 2018-05-28
[test] SpiMasterMock for device driver unittests

rleh:
To be able to test drivers for external peripherals connected via SPI to the microcontroller I propose `SpiMasterMock`.

`SpiMasterMock` utilizes the same interface as all *SpiMaster*-implementations (`BitBangSpiMaster`, `SpiMasterX`, ...), has no hardware dependencies and buffers the data sent from the device driver.

Unittests can access the buffer using
```C++
std::size_t length = SpiMasterMock::getTxBufferLength();

uint8_t buffer[length];
SpiMasterMock::popTxBuffer(buffer);
```
and compare it to the expected byte-sequence.

Additionally the unittest may inject data that the device driver reads from *SpiMasterMock*:
```C++
uint8_t buffer[] = {0x80, 0xFF, 0x00, 0x42};
SpiMasterMock::appendRxBuffer(buffer, 4);
```
---
As an example the unittest for the Ltc2984 driver is extended.

### DIFF
--- a/test/modm/platform/spi/mock/module.lb
+++ b/test/modm/platform/spi/mock/module.lb
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018, Raphael Lehmann
+#
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = "spi"
+    module.parent = "test:platform"
+
+def prepare(module, options):
+    module.depends(
+		"modm:architecture:spi",
+        "modm:container")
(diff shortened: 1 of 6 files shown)


test/modm/platform/spi/mock/module.lb
```diff
@@ -0,0 +1,26 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = "spi"
+    module.parent = "test:platform"
+
+def prepare(module, options):
+    module.depends(
+        "modm:container")
```
salkinium: And `modm:architecture:spi`.

salkinium (approved):
Very nice!

rleh:
I discovered one issue: The second unittest that uses `SpiMasterMock` might get data from the first unittest instead of empty buffers.

Two possible solutions:
* Add a dummy template parameter to the class.
* Add a `init()`/`clear()` function to the class.
Other ideas?

cc @salkinium

salkinium:
Use a clear method and put into the teardown of the test.

PR #27, opened 2018-05-29
Add DRV832x driver

rleh:
[DRV832xS](http://www.ti.com/product/drv8323): 6 to 60-V Three-Phase Smart Gate Driver With Three Current Shunt Amplifiers

This driver only covers the gate driver configuration accessible via SPI interface. (Devices with S suffix.)

[Datasheet](http://www.ti.com/lit/ds/symlink/drv8320.pdf)


salkinium (approved):
Thanks!

PR #28, opened 2018-06-11
Provide top-level includes like <modm/platform.hpp>

salkinium:
Instead of `<modm/platform/platform.hpp>`, which is redundant and annoying to use.
This restores the behavior as it was in xpcc, and thus makes it a little simpler to port from xpcc to modm.

I also moved the boards into their own submodule. The reasoning for using the `board.name` scheme was that this was a hint to the user that this is an exclusive module, only one board can be used, or the generated code would clash (all board modules generate at least the `board.hpp`).
However, it would be better to declare this to lbuild and let it manage this explictly.
cc @dergraaf 

Is there a need for a top top-level include, like `<modm.hpp>`? I believe @strongly-typed wanted this. It would only be available for user code, not for within modm, to allow the module system to continue to work.

This is a breaking change, sorry about this. cc @siy @rleh
- `<modm/architecture/architecture.hpp>` -> `<modm/architecture.hpp>`
- `<modm/board/board.hpp>` -> `<modm/board.hpp>`
- `<modm/platform/platform.hpp>` -> `<modm/platform.hpp>`
- `<modm/container/container.hpp>` -> `<modm/container.hpp>`
- `<modm/io/io.hpp>` -> `<modm/io.hpp>`
- `<modm/processing/processing.hpp>` -> `<modm/processing.hpp>`
- `<modm/utils/utils.hpp>` -> `<modm/utils.hpp>`

If you manually include any board module, you need to also do `modm:board.name` -> `modm:board:name`.


rleh (approved):

salkinium:
We can add a top-level include `<modm.hpp>` later if truly needed.

PR #40, opened 2018-07-12
[ui] Generate led tables via module options

salkinium:
Generates the product of table options.

The `SetOption` can only take enumerations, not number ranges, this should be fixed in lbuild in the future.

- [x] Generate tables via lbuild.
- ~~[ ] Add module documentation.~~

I'll add module documentation in the `feature/docs` branch, since I'm not sure what the right formatting should be for the yet-to-be-built documentation system.

cc @strongly-typed @dergraaf


salkinium:
Adding these options
```xml
<option name=":ui:led:gamma">2.2,2.0</option>
<option name=":ui:led:bit">8,14,16</option>
<option name=":ui:led:range">10,100,256</option>
```
will then create all these tables:
```cpp
namespace modm
{
namespace ui
{
	/// Provides 8-bit resolution (0 to 255) in 10 steps, 10 bytes of flash.
	extern modm::accessor::Flash<uint8_t> table22_8_10;
	/// Provides 8-bit resolution (0 to 255) in 100 steps, 100 bytes of flash.
	extern modm::accessor::Flash<uint8_t> table22_8_100;
	/// Provides 8-bit resolution (0 to 255) in 256 steps, 256 bytes of flash.
	extern modm::accessor::Flash<uint8_t> table22_8_256;
	/// Provides 14-bit resolution (0 to 16383) in 10 steps, 20 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table22_14_10;
	/// Provides 14-bit resolution (0 to 16383) in 100 steps, 200 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table22_14_100;
	/// Provides 14-bit resolution (0 to 16383) in 256 steps, 512 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table22_14_256;
	/// Provides 16-bit resolution (0 to 65535) in 10 steps, 20 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table22_16_10;
	/// Provides 16-bit resolution (0 to 65535) in 100 steps, 200 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table22_16_100;
	/// Provides 16-bit resolution (0 to 65535) in 256 steps, 512 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table22_16_256;
	/// Provides 8-bit resolution (0 to 255) in 10 steps, 10 bytes of flash.
	extern modm::accessor::Flash<uint8_t> table20_8_10;
	/// Provides 8-bit resolution (0 to 255) in 100 steps, 100 bytes of flash.
	extern modm::accessor::Flash<uint8_t> table20_8_100;
	/// Provides 8-bit resolution (0 to 255) in 256 steps, 256 bytes of flash.
	extern modm::accessor::Flash<uint8_t> table20_8_256;
	/// Provides 14-bit resolution (0 to 16383) in 10 steps, 20 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table20_14_10;
	/// Provides 14-bit resolution (0 to 16383) in 100 steps, 200 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table20_14_100;
	/// Provides 14-bit resolution (0 to 16383) in 256 steps, 512 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table20_14_256;
	/// Provides 16-bit resolution (0 to 65535) in 10 steps, 20 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table20_16_10;
	/// Provides 16-bit resolution (0 to 65535) in 100 steps, 200 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table20_16_100;
	/// Provides 16-bit resolution (0 to 65535) in 256 steps, 512 bytes of flash.
	extern modm::accessor::Flash<uint16_t> table20_16_256;
}
}
```

rleh:
I would prefer to explicitly select combinations of {gamma, bit, range}.

For example, I might want to create `table18_8_64` and `table22_32_65536`, but the following configuration creates eight tables and wastes >100kB flash memory.
```
<option name=":ui:led:gamma">1.8,2.2</option>
<option name=":ui:led:bit">8,32</option>
<option name=":ui:led:range">64,65536</option>
```
---
Also: Is there a reason to not allow *bit* values greater than 16? STM32 has 32bit timers...

salkinium:
> I would prefer to explicitly select combinations of {gamma, bit, range}.

Currently not possible with `SetOption`, which expects you to choose from an enumeration.
I could do it with a custom parsing of a `StringOption`, but that may be a bit weird and non-standard.
The `StringOption` really only exists to allow the linkerscript injections in the code module for the boards.

```xml
<option name=":ui:led:tables">{1.8,8,64},{2.2,32,65536}</option>
```

>  wastes >100kB flash memory.

Hm, doesn't the linker throw out the unused tables?

> Also: Is there a reason to not allow bit values greater than 16? STM32 has 32bit timers...

No, it just seems a little overkill to map a 256 (or even 1024) range onto 32-bits. It just wouldn't give you better resolution after a certain point, but always double the storage requirements (since there is no `uint24_t`).
Do you have a specific use-case in mind for 32-bit output values?

PR #42, opened 2018-07-24
[build.scons] Add CMake wrapper SCons tool

salkinium:
Brings back the `scons cmakewrapper` tool for making modm projects importable to CLion etc.

cc @rleh


rleh (approved):
:+1:

salkinium:
(We need to wrap SCons in CMake, since not all the tools we need are ported to CMake).

The goal is to have an IDE with usable debug functionality. Current state-of-the-art in xpcc is `arm-none-eabi-gdb -tui` which is better than nothing, but definitely not convenient, especially not for beginner to intermediate developers.

CLion seems to have better support for embedded than Qt Creator:
https://blog.jetbrains.com/clion/2017/12/clion-for-embedded-development-part-ii

cc @dergraaf @strongly-typed @siy

PR #47, opened 2018-07-31
added module file for SAB2

cajt:
module.lb was missing, added one derived from SAB.

### DIFF
--- a/examples/stm32f4_discovery/sab2/main.cpp
+++ b/examples/stm32f4_discovery/sab2/main.cpp
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2018, Carl Treudler
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/communication/sab.hpp>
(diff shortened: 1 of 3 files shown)


salkinium:
I didn't add this on purpose, since there wasn't an example to compile and verify its functionality. Do you want to add one? (I assume you're using this in another project).

cajt:
Well, of course, there is another project using it. Do you need an compile-able code example or a some type of test?

salkinium:
> Do you need an compile-able code example or a some type of test?

Example I think is enough for now, unit test would be excellent. But SABx is not used by the RCA anymore, hence low maintenance from my side.

examples/stm32f4_discovery/sab2/main.cpp
```diff
@@ -0,0 +1,68 @@
...
+modm::sab2::Interface<Usart1, 20> sabA;
+modm::sab2::Interface<Uart4, 20> sabB;
+
+int
+main()
+{
+	Board::initialize();
+	Board::LedBlue::set();
+
+	Usart1::connect<GpioB6::Tx>();
```
rleh: You ~can~ should connect Tx and Rx in a single line:
```cpp
Usart1::connect<GpioB6::Tx, GpioB7::Rx>(); 
```

salkinium: Not just *can*, you *should*, so that the API can check for pin group remap violations.

cajt: I don't know if this is a request or general commentary. I interpreted it as a request and there is the fix.

salkinium (dismissed):
Thanks!

PR #55, opened 2018-08-05
[ui] Fix build of KeyFrameAnimation

chris-durand:
key_frame.hpp and thus the F4 Discovery timer example does not build with a recent GCC because in KeyFrameAnimationBase::getMode() it is attempted to return a uint8_t. The required return type is KeyFrameAnimationMode. A static_cast is added to fix this.


salkinium (approved):
Thanks!

salkinium:
Moved into #54

PR #57, opened 2018-08-08
Various fixes and improvements to repo and build modules

salkinium:
Not much to see here, just some cleanup.

### DIFF
--- a/repo.lb
+++ b/repo.lb
@@ -15,11 +15,20 @@ import os
 import sys
 import glob
 
-rootpath = localpath("ext", "modm-devices", "tools", "device")
+rootpath = localpath("ext/modm-devices/tools/device")
 sys.path.append(rootpath)
 
-import modm.pkg
-import modm.parser
+try:
+    import modm.pkg
+    import modm.parser
+except Exception as e:
+    print(e, "\n")
+    print("You might be missing the git submodules in `ext/`.\n"
+          "Try:\n"
+          "    cd modm\n"
+          "    git submodule update --init\n"
+          "then build again.")
+    exit(1)
 
 def init(repo):
     repo.name = "modm"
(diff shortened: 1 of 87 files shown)


repo.lb
```diff
@@ -15,11 +15,20 @@ import os
...
+    import modm.pkg
+    import modm.parser
+except Exception as e:
+    print(e, "\n")
+    print("You might be missing the git submodules in `ext/`.\n"
+          "Try:\n"
+          "    cd modm\n"
+          "    git submodule update --init\n"
+          "then build again.")
+    exit(1)
```
salkinium, line 31: This should solve the "Oh, I've not added the `--recursive` flag to my `git clone` command", that happens fairly often.

test/all/run_all.py
```diff
@@ -65,14 +65,19 @@ def __init__(self, device):
...
-        lbuild_command.append("-D:target={}".format(self.device))
 
         with tempfile.TemporaryDirectory() as tempdir:
-            lbuild_command.extend(["-p", str(tempdir), "build", "--no-log"])
+            if self.device.startswith("at"):
+                lbuild_command = ["-c", "avr.xml"]
+                shutil.copyfile("avr.cpp", os.path.join(tempdir, "main.cpp"))
+            elif self.device.startswith("stm32"):
+                lbuild_command = ["-c", "stm32.xml"]
+                shutil.copyfile("stm32.cpp", os.path.join(tempdir, "main.cpp"))
```
salkinium, line 75: I feel like this should be done via lbuild too, maybe in a `:test:all` module, but then the unittest which do `:test:**` would also generate an additional `main.cpp` file. I'll fix it if it become annoying enough, this is fine.

salkinium (commented):
GitHub won't let me approve myself. How… depressing.

PR #59, opened 2018-08-09
Move git/build information into it's own file

salkinium:
This refactors the `info` SCons tool to create a `<info_git.h>` and `<info_build.h>` header file containing the information, instead of passing the information via the command line, to prevent complete rebuilds whenever it changed.

The generated files are .gitignored, so that if you commit the modm/ directory, you don't commit these files. 

Preliminary documentation can be found here: https://modm.io/reference/module/modm-build-scons/#information-tool

cc @rleh @chris-durand @strongly-typed

### DIFF
--- a/tools/build_script_generator/scons/site_tools/info.py
+++ b/tools/build_script_generator/scons/site_tools/info.py
@@ -19,9 +19,14 @@ import getpass
 import locale
 import re
 from collections import defaultdict
+from os.path import join, dirname, abspath
+
+TEMPLATE_SOURCE = join(dirname(abspath(__file__)), "info.h.in")
 
 def is_git_available():
-	return (subprocess.call(['which', 'git'], cwd=os.getcwd(), stdout=open(os.devnull, 'wb')) is 0)
+	git_exists = subprocess.call(['which', 'git'], cwd=os.getcwd(), stdout=open(os.devnull, 'wb')) is 0
+	is_git_repo = subprocess.call(['git', 'status'], cwd=os.getcwd(), stdout=open(os.devnull, 'wb')) is 0
+	return git_exists and is_git_repo
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


salkinium (commented):
I'm ok with merging this.

tools/build_script_generator/scons/site_tools/info.py
```diff
@@ -78,39 +85,42 @@ def git_info_defines(env, with_status=False):
...
-	defines['MODM_BUILD_MACHINE'] = platform.node()
-	defines['MODM_BUILD_USER'] = getpass.getuser()
+	subs = {}
+	defines['MODM_BUILD_INFO'] = "1"
+	subs['MODM_BUILD_PROJECT_NAME'] = env.get('CONFIG_PROJECT_NAME', 'Unknown')
+	subs['MODM_BUILD_MACHINE'] = platform.node()
+	subs['MODM_BUILD_USER'] = getpass.getuser()
 	# Generate OS String
 	if platform.system() == 'Linux':
 		os = " ".join(platform.linux_distribution())
```
chris-durand: `platform.linux_distribution()` is deprecated since Python 3.5 and will be removed in Python 3.8. It's not deprecated in python 2 though. On arch Linux SCons uses Python 2 by default for which `platform.linux_distribution()` returns `('', '', '')` for me. If we decide to keep it, we should probably add a check for empty string and just put `Linux` into MODM_BUILD_OS if no distro name is available.

salkinium: No wonder, this code was taken from xpcc. I’ll fix this, it’ll only be an issue later.

salkinium: So… the alternative is to `pip install distro` and use that. But I don't want to have to install another package just for this. I'll probably just revert to `platform.platform()`, which isn't the same, but close enough I guess. We'll fix it for real if someone needs it.

PR #60, opened 2018-08-10
[build] [scons] Add command-line arguments to manipulate CCFLAGS, CFLAGS, etc.

rleh:

---
```
```
---

Todo: I would like to add a option to permanently modify the flags.

What do you think about _modm:build:build_ options in the project.xml file? Could be used for both Scons and Cmake.

### DIFF
--- a/tools/build_script_generator/scons/resources/SConscript.in
+++ b/tools/build_script_generator/scons/resources/SConscript.in
@@ -117,4 +117,20 @@ env.Append(CCFLAGS="$ARCHFLAGS")
 env.Append(ASFLAGS="$ARCHFLAGS")
 env.Append(LINKFLAGS="$ARCHFLAGS")
 
+for flags in ["CCFLAGS", "CFLAGS", "CXXFLAGS", "ASFLAGS", "ARCHFLAGS", "LINKFLAGS"]:
+	flags_str = ARGUMENTS.get(flags)
+	if flags_str is not None:
+		flags_list = flags_str.split(",")
+		for flag in flags_list:
+			if len(flag) > 1 and flag[0] == "~":
+				try:
+					env[flags].remove(flag[1:])
+				except ValueError:
+					print("'" + flag[1:] + "' does not exist in " + flags +
+						" and therefore can not be removed.")
+					print("Info: " + flags + ": " + ", ".join(env[flags]))
+					exit(1)
+			else:
+				env[flags].append(flag)
+
 Return("library")


salkinium:
> What do you think about `modm:build:build` options in the project.xml file?

> I do not like to modify the SConstruct file because it might be overwritten by lbuild.

😒 from here: https://modm.io/reference/module/modm-build-scons/

> We do not intend to serve every possible use-case with this module. If you need something special, just write your own SConstruct file, maybe starting by modifying ours. Remember to set `modm:build:scons:include_sconstruct` to `False`, so that your custom `SConstruct` does not get overwritten.

salkinium:
Maybe this would work better as a SCons tool? And make it generic, so that you can modify any Environment variable, not just the flags.

In general, I'm really not convinced of this idea.
In my mind I already see people write specific command line arguments in the Readme to get their project to work. If people use IDEs via CMakeWrapper or manually call scons, then modifying the command line calls is annoying too.

Use a custom SConstruct and commit it. Makes it instantly reproducible on other machines.

salkinium:
I'm closing this in favor of adding the changes to the SConstruct file.

rleh:
> In my mind I already see people write specific command line arguments in the Readme to get their project to work. If people use IDEs via CMakeWrapper or manually call scons, then modifying the command line calls is annoying too.
> Use a custom SConstruct and commit it. Makes it instantly reproducible on other machines.

That is a completely different use-case and solution.

rleh:
> I'm closing this in favor of adding the changes to the SConstruct file.

Where can I find a example `SConstruct` file for that?

salkinium:
> Where can I find a example SConstruct file for that?

Ah, I remember what we talked about. I wanted to put a commented out example in the default `SConstruct` file, didn't I? Hm, I forgot, sorry.

Essentially you can modify the flags like a python list:
```py
env["CCFLAGS"].remove("-finline-limit=10000")
env["CCFLAGS"].append("-finline-limit=100")

# or the 'SCons' way of `.append` (there is no 'SCons' way for `.remove` tho)
env.Append(CCFLAGS="-finline-limit=100")
```

salkinium:
> Essentially you can modify the flags like a python list:
> ```py
> env["CCFLAGS"].remove("-finline-limit=10000")
> env["CCFLAGS"].append("-finline-limit=100")
> ```

I wonder if there is some metadata on GCC flags. Like the `-finline-limit` flag can only be set once, so they'll likely overwrite each other and only the last value is used.
But other flags are different, some cancel each other out, or there may be priorities of cancellations.
I'd like to have a `env.setFlag(CCFLAGS="-finline-limit=100")` and it'll warn me of the consequences if there are any.

salkinium:
Yes there is ([via Twitter/matt_dz](https://twitter.com/matt_dz/status/1033762571910881281)):

> Generally - option specification files: https://gcc.gnu.org/onlinedocs/gccint/Options.html
> Common: https://github.com/gcc-mirror/gcc/blob/gcc-8-branch/gcc/common.opt
> x86: https://github.com/gcc-mirror/gcc/blob/gcc-8-branch/gcc/config/i386/i386.opt
> AArch64: https://github.com/gcc-mirror/gcc/blob/gcc-8-branch/gcc/config/aarch64/aarch64.opt, https://github.com/gcc-mirror/gcc/blob/gcc-8-branch/gcc/config/aarch64/aarch64-option-extensions.def
> There are other files, too; e.g., optimizations:
> https://github.com/gcc-mirror/gcc/blob/gcc-8-branch/gcc/opts.c#L438
> https://github.com/gcc-mirror/gcc/blob/gcc-8-branch/gcc/common/common-target.h

And [about warnings](https://twitter.com/matt_dz/status/1033773671326920712):

> I'm only aware of a similar project for the warnings: https://github.com/Barro/compiler-warnings.

PR #61, opened 2018-08-11
Use relative, configurable paths for openocd.cfg and gdbinit

salkinium:
The initial idea was to define an `openocd.cfg` and `.gdbinit` file in the user directory, so that you'd only have to type `openocd`, which would then automatically load the `openocd.cfg`, same with `arm-none-eabi-gdb`.

However, this doesn't work when passing anything else to openocd, and GDB on some distros doesn't  automatically load `.gdbinit` files outside the home dir for security reasons.
So, both are called explicitly with `-f openocd.cfg` and `-x gdbinit`.

There are two problems with this:

1. This path is hardcoded, which required the files to be in `./`. Sometimes we need to place the generated `modm/` in subfolders.
2. They unnecessarily overwrite the user's files in `./`.

This PR moves the `openocd.cfg` and `gdbinit` files into the generated `modm/` folder, and makes their path relative to the SConscript using the `MODM_OPENOCD_CONFIGFILES` and `MODM_GDBINIT` environment variables, so that only the path to the `modm/SConscript` needs to be known.

For programming, openocd now calls `modm_program $SOURCE`, instead of `program_{profile}`, since for most of the RCA projects with custom SConstructs do not set the `modm:build:build.path` variable, and therefore the `modm:build` module generates wrong paths (ie. the default `build/{profile}/{project.name}.elf`). This change makes `scons program` work also in that scenario.

@rleh @chris-durand

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,3 @@ examples/**/modm
 examples/**/SConstruct
 examples/**/Makefile
 examples/**/generated
-examples/**/gdbinit
-examples/**/openocd.cfg
--- a/tools/build_script_generator/module.lb
+++ b/tools/build_script_generator/module.lb
@@ -46,12 +46,6 @@ def prepare(module, options):
 
     if options[":target"].has_driver("core:cortex-m*"):
         module.add_option(
-            BooleanOption(name="include_openocd", default=True,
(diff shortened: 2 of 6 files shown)


salkinium (commented):
👍

PR #69, opened 2018-08-18
Fixing the backslash issue in the generated files.

amarokmclion:
This will fix the backslash in the generated files allowing windows to be able to find the .elf files for openocd programming and will prevent an encoding issue of the file paths with scons in the SConscript file.

### DIFF
--- a/tools/build_script_generator/module.lb
+++ b/tools/build_script_generator/module.lb
@@ -11,6 +11,17 @@
 # -----------------------------------------------------------------------------
 
 import os
+import platform
+
+#filter to windowsify paths separators
+def windowsify(path):
+	if 'Windows' in platform.platform():
+	    path = path.replace('/','\\')
+	    path = path.replace("\\", "\\\\")
+	    return path
+	else:
+	    return path
+
 
 def init(module):
     module.name = "build"
@@ -64,6 +75,7 @@ def build(env):
     env.append_metadata_unique("openocd.search", "modm/openocd")
 
 
+
 def post_build(env, buildlog):
     target = env[":target"]
     env.substitutions = {
@@ -75,5 +87,5 @@ def post_build(env, buildlog):
         env.template("gitignore.in", ".gitignore")
 
     if env[":target"].has_driver("core:cortex-m*"):
-        env.template("openocd.cfg.in")
-        env.template("gdbinit.in")
+        env.template("openocd.cfg.in", filters={"windowsify": windowsify})
+        env.template("gdbinit.in", filters={"windowsify": windowsify})
--- a/src/modm/platform/clock/stm32/clock.hpp.in
+++ b/src/modm/platform/clock/stm32/clock.hpp.in
@@ -292,8 +292,8 @@ public:
 	 * Enable PLL.
 	 *
 	 * \code
-	 * VCO input frequency = PLL input clock frequency / PLLM [with 2 ≤ PLLM ≤ 63]
-	 * VCO output frequency = VCO input frequency × PLLN [with 64 ≤ PLLN ≤ 432]
+	 * VCO input frequency = PLL input clock frequency / PLLM [with 2 <= PLLM <= 63]
+	 * VCO output frequency = VCO input frequency × PLLN [with 64 <= PLLN <= 432]
 	 * \endcode
 	 *
 	 * \param	source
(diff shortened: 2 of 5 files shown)


rleh:
> This made me remember that windows 10 supports both forward and back slash as path separators even in the command line [...]

So this will changes will only work for Windows 10? What about Windows 7?

amarokmclion:
I'm sorry but I don't know if everything will work on Windows 7. I'll have to try, but I'm pretty sure that the modification to the env.Files will work as python almost always uses the right path separator. Only the OpenOCD parameter might not work as it's a direct command line call, but well you can manually change the openocd.cfg file. 
On the other hand the Jinja2 template modification will always work for each version of windows.

tools/build_script_generator/cmake/module.lb
```diff
@@ -35,8 +35,8 @@ def prepare(module, options):
 def build(env):
     project_name = env[":build:project.name"]
     build_path = env[":build:build.path"]
-    env.append_metadata_unique("elf.release", join(build_path, "cmake-build-release", project_name + ".elf"))
-    env.append_metadata_unique("elf.debug",   join(build_path, "cmake-build-debug",   project_name + ".elf"))
+    env.append_metadata_unique("elf.release", join(build_path, "cmake-build-release", project_name + ".elf").replace('\\','/'))
+    env.append_metadata_unique("elf.debug",   join(build_path, "cmake-build-debug",   project_name + ".elf").replace('\\','/'))
```
salkinium: I'd rather make this a part of the `openocd.cfg.in`/`gdbinit` template code, with a jinja2 filter (maybe `windowsify`), because I'm pretty sure I'll forget this `.replace` when adding new paths or refactoring this module.

Inside the `module.lb`:
```py
def windowsify(path):
    return path.replace("\\", "/")

env.template(..., filter={"windowsify": windowsify})
```

Then inside the template:
```
{{ path | windowsify }}
```

amarokmclion:
I did implement the "windowsification" as template filters, also it should work also with Windows 7 now.

tools/build_script_generator/module.lb
```diff
@@ -11,6 +11,17 @@
 # -----------------------------------------------------------------------------
 
 import os
+import platform
+
+#filter to windowsify paths separators
+def windowsify(path):
+	if 'Windows' in platform.platform():
+	    path = path.replace('/','\\')
```
salkinium, line 19: Is this a stray line? Cos above you only did this: `.replace('\\','\\\\')) #windows path compatibility hack`

amarokmclion: That line is needed because the path will be something like "path/to\\release.elf", because a normpath call is missing for the "path/to" part of the path. If you think it will be more clear I'll change it in a os.path.normpath(path).

salkinium (approved):
Thanks!

lmoesch:
Wouldn't it be more elegant to use [Pathlib](https://docs.python.org/3/library/pathlib.html) to handle paths for different operating systems, instead of hacking strings of slashs together?

amarokmclion:
From what I've seen the paths are already well formed using os.path (there are a few normpath calls missing but they are an easy fix when they come up), using Pathlib will need a complete refactor of the codebase. 
The issue here is that Jinja2 correctly writes a double backslash (i.e. 'path\\\\to\\\\foo.bar') to a single backslash in the template files, this creates a problem with how OpenOCD handles the .elf files for programming and debugging, and how Scons handles the env.Files strings (don't ask me why). 
This is more a workaround to get all things working together rather than a deep problem with how the paths are managed.

salkinium (dismissed):
I’m happy to merge this, I think we can add more Windows compatibility fixes when they come up. Thanks everyone!

salkinium:
I'm getting so annoyed with CircleCI. 🙄 
```
  Error pulling image modm/modm-build:latest: Error response from daemon: received unexpected HTTP status: 503 Service Unavailable... retrying
  image cache not found on this host, downloading modm/modm-build:latest
```

salkinium:
Ok, it's not CircleCI's fault this time, but Docker's. Still 🙄 

> On August 25th at 11am PT (Saturday), we're performing major infrastructure upgrades to Docker Hub, Store and Cloud. We anticipate this upgrade will take between 15 and 45 minutes to perform.

PR #73, opened 2018-08-22
Refactor and document examples

salkinium:
The examples contain a lot of duplicate code, it would be nice if the examples that work on multiple boards are placed in one location and compiled with different targets.

For this purpose the `tools/script/example_compile.py` script will parse the `project.xml` file and execute a modifier function for modifying itself.

The `generic/blinky` example has been modified this way to compile all boards with SCons and if available with CMake.


examples/generic/blinky/project.xml
```diff
@@ -1,5 +1,12 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <library>
+  <!--
+    ..ci:
+        for board in boards:
+            new_config = config.replace("/blinky", "/blinky/" + board) \
+                               .replace("nucleo_l476rg", board)
+            yield (new_config, "*", board)
+  -->
```
salkinium, line 9: This is the actual generator for modifying the `project.xml`.

salkinium:
cc @rleh @chris-durand @strongly-typed

examples/generic/can/main.cpp
```diff
@@ -26,82 +35,67 @@ displayMessage(const modm::can::Message& message)
...
+	Board::Leds::setOutput();
+#ifndef MODM_BOARD_HAS_LOGGER
+    Generic::Uart::connect();
+#endif
 	MODM_LOG_INFO << "CAN Test Program" << modm::endl;
 
-	MODM_LOG_INFO << "Dividing filter bank..." << modm::endl;
-	CanFilter::setStartFilterBankForCan2(14);
-
 	MODM_LOG_INFO << "Initializing Can1..." << modm::endl;
```
rleh, line 73: `Generic::Can` is not necessarily `Can1`.

examples/generic/connect.hpp
```diff
@@ -0,0 +1,86 @@
...
+{
+
+using Tx = GpioOutputA9;
+using Instance = Usart1;
+
+void
+connect()
+{
+	Instance::connect<Tx::Tx>();
+    Instance::initialize<Board::systemClock, 115200>();
```
rleh, line 33: Tab?

examples/generic/connect.hpp
```diff
@@ -0,0 +1,86 @@
...
+using Rx = Board::D67;
+#elif defined(MODM_BOARD_NUCLEO_64)
+using Tx = Board::D14;
+using Rx = Board::D15;
+#else
+using Tx = GpioOutputB9;
+using Rx = GpioInputB8;
+#endif
+
+// using Tx2 = GpioOutputB13;
```
rleh, line 55: ?

examples/generic/can/main.cpp
salkinium: I know, I'm not finished refactoring. The CAN example in general is pretty messy, but I'd like to add Can2 as well, when it's available.

examples/generic/connect.hpp
salkinium: Adding Can2 loopback mode or something. I'm not sure anymore how much of this is example and how much is an integration test. 😒 hwut?

rleh: Ok

rleh:
The _examples_ in moved to `examples/generic` are not really examples anymore. (e.g. for newbies to start using modm)
I think `compile-test/` would be a better name than `examples/generic`...

salkinium:
> The examples in moved to examples/generic are not really examples anymore. (e.g. for newbies to start using modm)

There's also `examples/stm32/nucleo` which works the same way, but has a bit more variety to it.

I'm not sure how useful these examples are in general, most of them are basically integration tests or require external hardware. There are only a few really self-contained ones, like the accel/gyro to LED ones or the game of life one.
Like is a blinky example really that relevant? I guess it shows off the API for a really dead-simple thing.

rleh:
> I'm not sure how useful these examples are in general [...]

Very useful to absolutely indispensable in my opinion.

To start with "modm" I would hate to have to search for documentation (if it exists at all) and then adapt a "generic" example to my dev board.
People might actually want to use modm (only) to do tasks with microcontrollers. :wink: 

A lot of boards (F042. L476, F746, F0-DISCO, F030, ...) only have/had one blink example. With their blink examples deleted there is no possibility to just do `cd examples/.../blink && lbuild build && scons`.

The generic examples add more 'magic'. In my opinion the `board.hpp` file is almost too much magic.
Going from a dev board supported in modm to a custom board is not that easy for new modm users (because it is completely undocumented).

chris-durand:
> Like is a blinky example really that relevant? I guess it shows off the API for a really dead-simple thing.

The relevant part of the blinky example is not the blinking API. You get a minimum working example for your specific board.
Having to find out how the build system works might deter people from trying modm at all. It will be easier for beginners if you have something that works on your board right away instead of having to find out how to write a proper `project.xml`. While the blink examples are not very useful on their own they offer a useful starting point.

salkinium:
> With their blink examples deleted there is no possibility to just do `cd examples/.../blink && lbuild build && scons`.
> While the blink examples are not very useful on their own they offer a useful starting point.

Ok, I agree. I think this can be solved with ✨ EVEN ✨ MORE ✨ CODE ✨ GENERATION ✨!1!!
I wanted a way to make *all* boards compile the generic examples so I can test the BSPs.
This way when you add a new BSP you get all these tests for free.

So instead of the `examples/generic/*`, I'll put them into a template and generate them into their right place. But what about the `examples/stm32/nucleo`?

rleh:
https://github.com/modm-io/modm/blob/9d9667c57f79ac46e275d0ad1f1e0a2bda8f64f3/examples/stm32/nucleo/comp/main.cpp#L35
I am pretty sure this will not work on all (nucleo) boards...
Not every example is scalable across every dev board.

---
> :sparkles: EVEN :sparkles: MORE :sparkles: CODE :sparkles: GENERATION :sparkles: !1!!

My proposal:
* Move all generic examples to `example-templates/` (and maybe convert them into templates?).
  * Generate the examples to `example/generated/`.
  * Every example has a white- or preferably black-list of dev boards to be generated for.
E.g. the nucleo-generic examples (`examples/stm32/nucleo`) filter `board:nucleo-*`.
  * The generated examples are committed and not git-ignored. A CI job re-generates them and assures there are no (manual) changes.
* Move remaining (board specific, ...) examples to `example/hand-crafted/`.

chris-durand:
> Every example has a white- or preferably black-list of dev boards to be generated for.
E.g. the nucleo-generic examples (examples/stm32/nucleo) filter board:nucleo-*.
The generated examples are committed and not git-ignored. A CI job re-generates them and assures there are no (manual) changes.

Do you think it is useful to generate every sensor example for a large number of boards by default? This will increase CI build times which are already an issue. Maybe generating them on demand is a better option. I would prefer to have the examples for those boards for which we actually did a test in hardware.

The examples serve mainly two purposes: to test successful compilation of device drivers and to demonstrate how to use certain APIs. For both a single example should be sufficient. For the HAL we already have compilation tests for all targets.

rleh:
In the long term with hwut (hardware unit tests) I would like to test every example on every compatible dev board.

Currently there is no huge benefit of compiling more examples.
To satisfy the `cd examples/.../blink && lbuild build && scons` use case I would like all examples to be generated and committed.
Until we have a faster CI, we can just manually select some of the generated examples for CI.

salkinium:
> Do you think it is useful to generate every sensor example for a large number of boards by default? This will increase CI build times which are already an issue.

And it will scale "every example" times "every board": N * M, which is pretty bad, if you consider how many Nucleo boards there are that we can add with actually reasonably little effort.

The issue is that the examples are supposed to be hand-crafted stuff, but lately they're being abused for bring-up tests or as (bad) manual hwut substitutions. Like, there literally is a STM32F7 example that just multiplies two doubles to show the double precision FPU. That's not even a test, how could I tell by running the example in hardware that it's using the FPU?!? You need to compile one cpp file to asm and grep for some floating point instructions. There is a lot of useless shit there, like [super basic Uart](https://github.com/modm-io/modm/blob/develop/examples/stm32f4_discovery/uart/main.cpp) or [SPI examples](https://github.com/modm-io/modm/blob/develop/examples/stm32f4_discovery/spi/main.cpp) which actually belong in the documentation ([tested of course](http://www.roboterclub.rwth-aachen.de/projekt/modm-dokumentation.html), [perhaps annotated like this](https://github.com/modm-io/modm/blame/feature/docs/src/modm/ui/led/module.md#L150-L213)).

> Maybe generating them on demand is a better option. I would prefer to have the examples for those boards for which we actually did a test in hardware.

I want to have a `modm` python tool, that you can do `modm new -b nucleo_f103rb` or something that will generate a project structure perhaps with blinky for you. It would be a light wrapper around lbuild (and it's dependencies), perhaps similiar in spirit to [`mbed-cli`](https://github.com/ARMmbed/mbed-cli).

I mean the basics of such a tool are already in modm anyways, we have a script for [generating the unittests](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/site_tools/unittest.py), for [compiling all target HALs](https://github.com/modm-io/modm/blob/develop/test/all/run_all.py), for [compiling the examples](https://github.com/salkinium/modm/blob/feature/compactify_examples/tools/scripts/examples_compile.py), for [gathering and compiling the documentation](https://github.com/modm-io/modm/blob/feature/docs/tools/doc_generator/module.lb), they all integrate lbuild in various inconsitent ways anyway.

salkinium:
> In the long term with hwut (hardware unit tests) I would like to test every example on every compatible dev board.

I don't even think hwut is required in the short term, a unittest that either [directly accesses the hardware](https://github.com/modm-io/modm/tree/develop/test/modm/platform/gpio) or [uses a mockup](https://github.com/modm-io/modm/tree/develop/test/modm/platform/spi/mock) would be pretty great.

amarokmclion:
> I want to have a modm python tool, that you can do modm new -b nucleo_f103rb or something that will generate a project structure perhaps with blinky for you. It would be a light wrapper around lbuild (and it's dependencies), perhaps similiar in spirit to mbed-cli.

I'd really love to have something like that. If there is something to do just ask.

salkinium:
> I'd really love to have something like that.

At the moment I'm refactoring lbuild since it has grown a little wild while we were adding the features we discovered we needed during the xpcc-to-modm porting, so this'll have to wait just a little longer.

> If there is something to do just ask.

Yes, maybe you can describe your expectations of such a tool. What use-cases do you want covered, how would the CLI look like? The more info I have on this, the better the design.

amarokmclion:
> > I'd really love to have something like that.
> 
> At the moment I'm refactoring lbuild since it has grown a little wild while we were adding the features we discovered we needed during the xpcc-to-modm porting, so this'll have to wait just a little longer.

Okay I'll also wait to push the bsp I was preparing for the NUCLEO-F446RE that I'm using.

> > If there is something to do just ask.
> 
> Yes, maybe you can describe your expectations of such a tool. What use-cases do you want covered, how would the CLI look like? The more info I have on this, the better the design.

What I'd like to see is some kind of barebone project creator, for example:

`modm new --board nucleo-f446re --demo blinky --module rtos timer1 cmake --ide vscode`

This will create a small project with a blinky demo the rtos, timer1 and cmake module and the nucleo-f446re bsp  already enabled in project.xml. In addition it will create the basic .json file to enable the build and debug commands in visual studio code (I'm using vscode just as an example it could be any other ide or text editor).

After having implemented this basic functionality it will be possible to build upon it with a simple automatic clock configuration generator (the default will be the internal oscillator at maximum frequency for example) and a parametric blinky project (taking the intended blinking GPIO as a parameter) that will help building working projects without a BSP.

salkinium:
> Okay I'll also wait to push the bsp I was preparing for the NUCLEO-F446RE that I'm using.

You can open a PR for that, the lbuild changes are just internal, the interface will remain the same.

rleh:
> `modm new --board nucleo-f446re --demo blinky --module rtos timer1 cmake --ide vscode`

I my opinion the `--ide` argument is not useful because editor/IDE configuration depends on the build system. E.g. the `cmakewrapper` and `qtcreator` are submodules of the scons module.

rleh:
My vision of the `modm-cli`:
* `modm-cli init [--repo /path/to/modm [--no-clone]]` clones the modm git repository recursively
  * to `~/.modm/repository`
  * or to `/path/to/modm` if `--repo` is given
  * If the repository is already cloned `--no-clone` option can be used
  * The repository path is stored in the config file `~/.modm/config`
* Maybe something like a `modm-cli update` to update the git repository recursively?
* `modm project create (--board nucleo-f429 | --mcu stm32f429zit) [--path /project/path] [--modules module1,module2] [--list-available-modules]`
* `modm project update` (`lbuild build`)
* `modm project clean` (`lbuild clean`)
* `modm example (--board nucleo-f429 | --mcu stm32f429zit) --example uart` like `modm project create` but without `--modules` argument and additionally generates example source code files.

`modm-cli` wraps (or replaces) the library-builder cli (`lbuild discover-modules, discover-option, dependencies, ...`).

salkinium:
Thanks for the feedback! I generally agree, some details:

> a simple automatic clock configuration generator

See https://github.com/salkinium/save-the-clocktrees and https://github.com/roboterclubaachen/xpcc/pull/39 for some background to this.

> modm-cli wraps (or replaces) the library-builder cli

Just wraps, lbuild needs to be a separate tool, it it very useful for a whole bunch of things unrelated to our specific code gen requirements. I may add some features to lbuild and wrap then in modm-cli, also see https://github.com/dergraaf/library-builder/issues/24.

> `modm-cli init [--repo /path/to/modm [--no-clone]]` clones the modm git repository recursively

We don't have version management in lbuild and I also don't want to go down that rabbit hole, so this would add modm as a git submodule (+ `--no-submodule` flags to disable). If the `--path` is not given, it would navigate to the top of the git repo and add it to `ext/modm`. You are then responsible for consciously upgrading modm and dealing with the API changes (via our changelog, like in xpcc).
I would by default only choose the latest *release*, and then add a `--head` for using the `develop` branch.

Maybe I'm overthinking this tho, so I will first prototype these workflows before implementing anything.

> The repository path is stored in the config file `~/.modm/config`

Yes! I was planning on adding this to lbuild for some time. The file would be apartial config file which would be inherited by your project file. One thing I want in lbuild is the ability to name alias config files.
So instead the full path `<extends></extends>` the repo.lb can declare a mapping from name to path `<extends>modm:board:disco-f469ni</extends>`. Then you can move these configs around without breaking the interface to projects.

rleh:
> We don't have version management in lbuild and I also don't want to go down that rabbit hole, so this would add modm as a git submodule (+ --no-submodule flags to disable). If the --path is not given, it would navigate to the top of the git repo and add it to ext/modm.

* The user might not use _git_ or another VCS, so it is not good to depend on a _git_.
* If the user generates his modm-environment (and puts it into his VCS) there is no need to put the modm source repository in his project.

---

> You are then responsible for consciously upgrading modm and dealing with the API changes (via our changelog, like in xpcc).
I would by default only choose the latest release, and then add a --head for using the develop branch.

:+1: 

---

> The file would be a partial config file which would be inherited by your project file.

KiCad does something with their library management and it is horrible, because projects don't work out of the box on other people computers.
A `~/.modm/config` file should only store user preferences and computer-specific settings (like the path to a modm repository that is needed by `modm-cli`)

salkinium:
> The user might not use git or another VCS, so it is not good to depend on a git.
> If the user generates his modm-environment (and puts it into his VCS) there is no need to put the modm source repository in his project.

Yeah, good points! I agree using git submodules that way is a RCA thing, not how we recommend using modm.

> KiCad does something with their library management and it is horrible, because projects don't work out of the box on other people computers.
> A ~/.modm/config file should only store user preferences and computer-specific settings (like the path to a modm repository that is needed by modm-cli)

See #76 on how I did this: I chose a plain `examples/lbuild.xml`, because I don't like hidden .dotfiles which you will forget to copy and then stuff won't work anymore. I think this works pretty good.

salkinium:
Closing this in favor of #359.

PR #75, opened 2018-08-25
Improve linkerscript and add `-fwrapv` flag

salkinium:
This refactors the debug linkerscript sections for DWARF-4 format.

Additionally, the `-fwrapv` flag is added to make signed integers overflow as expected and not leave it undefined for the questionable benefit of some compiler engineers to help doctor their benchmarks with.

> This option instructs the compiler to assume that signed arithmetic overflow of addition, subtraction and multiplication wraps around using twos-complement representation. This flag enables some optimizations and disables others.

### DIFF
--- a/src/modm/platform/core/cortex/linkerscript/linker.macros
+++ b/src/modm/platform/core/cortex/linkerscript/linker.macros
@@ -322,44 +322,20 @@ TOTAL_STACK_SIZE = MAIN_STACK_SIZE + PROCESS_STACK_SIZE;
 
 
 %% macro section_debug()
-	/* Stabs debugging sections.  */
-	.stab          0 : { *(.stab) }
-	.stabstr       0 : { *(.stabstr) }
-	.stab.excl     0 : { *(.stab.excl) }
-	.stab.exclstr  0 : { *(.stab.exclstr) }
-	.stab.index    0 : { *(.stab.index) }
-	.stab.indexstr 0 : { *(.stab.indexstr) }
-	.comment       0 : { *(.comment) }
-
-	/* DWARF debug sections.
-	 Symbols in the DWARF debugging sections are relative to the beginning
-	 of the section so we begin them at 0.  */
-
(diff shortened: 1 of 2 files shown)


rleh:
What happened to the `-fwrapv-pointer` flag?

chris-durand:
I'm actually not sure if `-fwrapv` is a good idea. It can suppress optimizations that users expect to happen. Imagine you have an arithmetic expression that depends on multiple compile time constants, like `variable * a / b` where a = 10 and b = 5. With `-fwrapv` the compiler cannot optimize the expression to `variable * 2` and [emits multiplication instructions](https://gcc.godbolt.org/z/4EmKHa) instead of a single left shift. This can matter in practice.

salkinium:
> What happened to the -fwrapv-pointer flag?

arm-none-eabi-gcc doesn't accept it. Don't know why.

> arm-none-eabi-g++: error: unrecognized command line option '-fwrapv-pointer'; did you mean '-fcray-pointer'?

salkinium:
> `variable * a / b` where a = 10 and b = 5. With `-fwrapv` the compiler cannot optimize the expression to `variable * 2` and emits multiplication instructions instead of a single left shift.

Cos technically this is grouped left to right so `(variable * 10) / 5`, which can obviously overflow.
If you do `variable * (10 / 5)` everything's fine. This is behavior that I'd expect.
Your example is actually quite funny, check out the difference in assembly between `-Og -mcpu=cortex-m4 -fwrapv` and `-Os -mcpu=cortex-m4 -fwrapv`. Quite terrifying.
With `(10 / 5)` the assembly is always the same.

In general, I'd claim that if you write math in C(++), always use braces to force evaluation order and don't rely on compiler stuff like this.

salkinium:
I'll merge this and then just blame you guys for the great fwrapv fallout of 2018 when we compile our robot code and nothing works anymore. Deal?

PR #76, opened 2018-09-02
Use new lbuild repository features

salkinium:
This adds several lbuild features introduced in https://github.com/modm-io/lbuild v1.x:

- repo formatters for modm descriptions: removing HTML comments and some markdown formatting.
- repo ignore patterns: `*.lb` and `*/board.xml` cut down on a lot of duplicate code.
- repo configuration aliases: used for extending board configurations.
- lbuild base config loaded from filesystem.
- option dependencies

cc @dergraaf @rleh @chris-durand

### DIFF
--- a/examples/lbuild.xml
+++ b/examples/lbuild.xml
@@ -1,0 +1,10 @@
+<library>
+  <!-- This is the default lbuild configuration file for every
+       example in this folder. It only sets the repository path,
+       so that this isn't duplicated in every single example.
+       When you write your own application, you must set this
+       path yourself! -->
+  <repositories>
+  	<repository><path>../repo.lb</path></repository>
+  </repositories>
+</library>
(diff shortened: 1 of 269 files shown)


src/modm/board/al_avreb_can/module.lb
```diff
@@ -29,7 +29,7 @@ def prepare(module, options):
 
 def build(env):
     env.outbasepath = "modm/src/modm/board"
-    env.copy('.', ignore=env.ignore_patterns("*.lb", "*.cfg", "*.xml"))
```
rleh, line 32: `*.cfg` is no longer ignored?

salkinium: It's not needed, since the only .cfg files in modm are in `tools/openocd/modm`.

examples/lbuild.xml
```diff
@@ -0,0 +1,7 @@
+<library>
+  <repositories>
+    <repository><path>../repo.lb</path></repository>
+  </repositories>
+  <options></options>
+  <modules></modules>
```
salkinium: One could also put `<module>modm:build:scons</module>` in here, since it's used in every example, but I would add this in #73 instead.

I need to add a comment to this file, describing what it does, otherwise people will copy the examples over and it won't work anymore.

Issue #18, opened 2018-05-18
Use C++17 language features


salkinium:
We're already using C++14 by default, which requires at least GCC 5.
This is an issue for the official avr-gcc from Atmel which ships with Ubuntu in version 4.9.3.
Most other linux distributions, as well as [macOS via Homebrew](https://github.com/osx-cross/homebrew-avr) (Q: what about Windows?) compile their own avr-gcc from upstream, which is missing support for several AVR architectures and devices.
For the sake of moving forwards, we've accepted this caveat and are using a [self-compiled avr-gcc 7.3.0](https://github.com/salkinium/docker-avr-gcc-7/) for our CI.
ARM already ships arm-none-eabi-gcc for v7.2.0 since the 2017q4 release.

The [C++ standards support page for GCC](https://gcc.gnu.org/projects/cxx-status.html) shows that all language features of C++17 (except fixes to template argument deduction for class templates) are supported by GCC 7.
I would therefore like to enable C++17 as the default for modm to already make use of very useful new features such as `constexpr if`, enhancements to variadic templates, selection statements with initializers, constexpr lambdas and many more.

cc @dergraaf @strongly-typed @rleh @chris-durand @daniel-k @ekiwi @georgi-g @mhthies

siy:
It was the thing which I was about to ask for.

rleh:
## `arm-none-eabi-gcc`
* ARM toolchain: [GCC 7.1](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads)
    * [PPA for Ubuntu](https://launchpad.net/~team-gcc-arm-embedded/+archive/ubuntu/ppa) :+1:
    * Download (precompiled) for Windows :+1:
    * Download (precompiled or source) for *Linux 64-bit* :+1:
* Ubuntu (since bionic): [GCC 6.3.1](https://packages.ubuntu.com/bionic/gcc-arm-none-eabi) :-1: (but PPA see above)
* Debian: [GCC 5.4.1](https://packages.debian.org/stretch/gcc-arm-none-eabi) :-1:
* Fedora: [GCC 7.1.0](https://apps.fedoraproject.org/packages/arm-none-eabi-gcc-cs) with CodeSourcery patches :+1:
* Arch Linux: [GCC 8.1.0](https://www.archlinux.org/packages/community/x86_64/arm-none-eabi-gcc/) :+1:
[...]

PR #77, opened 2018-09-06
[WIP] libstdc++ for AVR

chris-durand:
This PR ports a subset of libstdc++ to the AVR platform. It will allow for portably using large parts of the C++ standard library inside modm. Almost the complete implementation is copied from libstdc++ sources released with GCC 8.2.0. Most headers do not require any modifications.

Some parts have been deliberately excluded because they are unavaliable on embedded platforms or the provided implementation is inherently unsuitable for AVR and requires considerable porting efforts. For instance, the string conversion functions contain kilobytes of static data that would be put into RAM.

Iostreams and std::string are also not available. All stream-related code is removed from the supplied headers. 

The following parts are included:

- Containers
- Algorithms (`<algorithm>`)
- Iterators library ( `<iterator>`, except for iostream iterators)
- Numerics library (`<complex>`,  `<numeric>`, `<ratio>`, `<valarray>`)
- Utilities library (`<type_traits>`, `<any>`, `<optional>`, `<variant>`, `<tuple>`, `<utility>`, `<chrono>`, …)
- Numeric limits (`<limits>`)
- C headers except for C99/C11 additions not supported by AVR libc (`<c*>`)
- Memory management (`<new>`, `<memory>`, `<scoped_allocator>`), except for `std::shared_ptr`
- String view (`<string_view>`)

A complete list of headers can be found [here](https://github.com/modm-io/modm/blob/1d3051307de860c06a9753270d5bc5d319c102f4/test/stdc%2B%2B/compile_test.cpp).

**Not** included (list may be incomplete):
- `std::string` and string conversions, regex
- Anything that requires an OS to work (threads, filesystem, locales, ...)
- Atomics
- iostreams and related headers
- Random number generation (`<random>`)
- `std::shared_ptr` (build on atomics, high porting effort)
- `std::bitset`
- Polymorphic allocators (`<memory_resource>`), experimental in libstdc++
- Exception headers
- RTTI headers

Most of the unit tests are yet missing. There is one for vector and one to check if all headers compile. The vector test runs successfully on an Arduino Nano clone (with avr-g++ 8.2.0).

@salkinium I tried to use `modm_assert` for error handling [inside the standard library](https://github.com/modm-io/modm/commit/94930707352296901814cca47eae06c1dd28a0f8#diff-2653880f1ad58695f325c0396ed589a6R18). It increased the flash usage of an empty example by >200 bytes. Is this avoidable?  

TODO:
   - [x] Make it work with avr-gcc 7
   - [ ] Reasonable unit test coverage
   - [x] Provide a hash function implementation for AVR (16-bit `size_t`), libstdc++ only has [a dummy implementation](https://github.com/modm-io/modm/commit/ace79f92da4d8d5950a65e6001af228d7db93399#diff-d55adb1988a673f28ff31e7ae7679366R180)
   - [x] Use modm_assert for error handling


chris-durand:
It does not compile with avr-gcc 7 :unamused:

salkinium:
First of all: THANK YOU. This makes AVR support so much less painful!

> I tried to use modm_assert for error handling inside the standard library. It increased the flash usage of an empty example by >200 bytes. Is this avoidable?

Probably not, but you could use `modm_assert_debug`, which is only enabled in debug builds. Or make it an lbuild option for [disable, debug, debug+release]?

I assume you had to remove/comment out some things. Is there an intention to provide patches? I'm thinking that this may not really be maintainable without patches. Preferrably with a script of some kind that makes this repoducible. What do you think?

salkinium:
I'm checking if I can compile avr-gcc 8.2 for the CI, but I'm not sure how easily you can get hold of an up-to-date avr-gcc for the various OSes.

rleh:
`avr-gcc 8.x` seems to be only available on Arch Linux.

https://github.com/modm-io/modm/issues/18#issuecomment-390430223

salkinium:
Well, my misnamed repository can also build avr-gcc 8.2 so we could have it in the CI.
https://github.com/salkinium/docker-avr-gcc-7/releases/tag/v8.2.0

> avr-gcc 8.x seems to be only available ohn Arch Linux.

This tho.

chris-durand:
> I assume you had to remove/comment out some things. Is there an intention to provide patches? I'm thinking that this may not really be maintainable without patches. Preferrably with a script of some kind that makes this repoducible. What do you think?

Commit  [ace79f9](https://github.com/modm-io/modm/pull/77/commits/ace79f92da4d8d5950a65e6001af228d7db93399) imports completely unmodified files distributed with gcc 8.2.0. To reproduce it you just have to copy files from gcc sources to `modm/src/stdc++`. It is possible to have a script to do that.

`c++config.h` and `c++allocator.h` from [4c06490](https://github.com/modm-io/modm/pull/77/commits/4c06490a2324ed3c38e3c5e0e1d3ce629db0dc70) are platform specific headers that are somehow generated during the gcc build process. `c++allocator.h` is an unmodified file from arm-none-eabi-gcc. `c++config.h` is basically the arm version with some unavailable C headers disabled.

The following commits are more or less the patches you are looking for. All modifications of libstdc++ files are visible in git.

chris-durand:
> I'm checking if I can compile avr-gcc 8.2 for the CI, but I'm not sure how easily you can get hold of an up-to-date avr-gcc for the various OSes.

It will work with avr-gcc 7.x, too. Gcc 8 adds the two intrinsics `__is_assignable` and `__is_constructible` to improve the implementation of some type traits. I'll fix the compile error in the CI.

chris-durand:
Now all headers compile with avr-gcc 7.2.0 and 7.3.0.

salkinium:
What do you think about moving this code to a separate repository and integrating it as a submodule via `ext/`? Cos +91,163 −677 is a lot of code, and it'll likely change a lot on an eventual version upgrade. I would preserve the history by extracting the entire folder from your fork.

salkinium:
Btw, strongly related work: libc++ freestanding proposal: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0829r1.html

chris-durand:
> What do you think about moving this code to a separate repository and integrating it as a submodule via ext/? Cos +91,163 −677 is a lot of code, and it'll likely change a lot on an eventual version upgrade.

I would actually prefer to have it inside modm. One reason is that I like to have a CI for it. If we don't have it someone will make a change in the external repo that breaks something and we will see it fail in the next unrelated modm PR. This will become even more important when we adopt real hardware unit tests.

Building it already depends on modm and the unit tests have to be in modm anyway. I wouldn't be afraid of too large changes. Apart from smaller bug fixes it won't change much until C++20 gets standardized, we adopt it and drop support for all GCC versions without C++20 support (maybe ~2021?).

If you like it or not, this will be an integral part of modm (unless you drop AVR support :wink:). Yes, it's a lot of code, but putting it into an external repo does not relieve us from caring for it.
In my opinion we don't gain much from moving it to an external repo beside tweaking some code line statistics counters.

rleh:
The CI job in the external repo can clone modm recursively, update the submodule `ext/libstdc++` to the necessary commit and then run the unittests.

But since this avr port of libstdc++ is not usable without modm I am not sure about the benefit of using a submodule.

salkinium:
We should definitely move the files to `ext/stdlibc++` since the code even with your changes is overly not ours anymore.

> this avr port of libstdc++ is not usable without modm

Are you sure? Looking at the commits it seems that the work is mostly AVRs specific, not modm-specific, with the exception of modm_assert and the modm unittests. I think this could be made into its own repository with its own minimal examples and modm repo tests. What do we need for minimal support? Dynamic memory, build system, some cxxabi?

chris-durand:
> Dynamic memory, build system, some cxxabi?

Probably. We would need some kind of build system for minimum examples and an integration with the modm ci. If we move it to an external repo, it should be usable without python, jinja, scons ...

We could elegantly solve the modm_assert dependency by making the error handlers weakly linked in the library and replace them with handlers calling modm_assert inside modm.

salkinium:
>  If we move it to an external repo, it should be usable without python, jinja, scons ...

Just checked, the `modm:platform:core` for AVR doesn't depend on any data, only the dynamic memory implementation has options for block sizes (which we can hardcode for our test devices tho).
So, I think that's perfectly doable, we'll just copy out the bare-minimum parts from modm we need.

> We would need some kind of build system for minimum examples and an integration with the modm ci.

Agreed, I would only provide a minimal set of non-modm examples for AVRs, and then do the unittest *only* in modm, like @rleh described, by force updating the submodule explicitly for the CI.
I think anyone who wants to use libstdc++ on AVRs is probably robust enough to use this.

I'll figure out how to do the minimal buildable examples as a Makefile, do you want to figure out how to move this to a separate repo and run the modm CI in that?

chris-durand:
> I'll figure out how to do the minimal buildable examples as a Makefile, do you want to figure out how to move this to a separate repo and run the modm CI in that?

I'll look into it tomorrow.

salkinium:
Preferences for repository naming?

- gcc-libcpp-avr
- avr-libcpp-gcc

rleh:
`gcc-libstdc++-avr`?

salkinium:
> gcc-libstdc++-avr

can't have `+` in repo names, so `gcc-libstdcpp-avr`?

chris-durand:
Hm, + really is not allowed? I called my local copy `libstdc++-avr`.

salkinium:
> Hm, + really is not allowed?

screen shot 2018-09-11 at 17 51 36

rleh:
grafik

salkinium:
Moving this to here for now: https://github.com/modm-io/avr-libstdcpp

salkinium:
Moved to #82

PR #85, opened 2018-09-18
Improve VL53L0

strongly-typed:
The naming error was probably introduced when porting from xpcc to modm.


src/modm/driver/position/vl53l0.hpp
```diff
@@ -438,7 +438,7 @@ class Vl53l0 : public vl53l0, public modm::I2cDevice< I2cMaster, 5 >
 	/// the predicate must take a uint8_t argument and return a bool
 	template < typename Predicate >
 	modm::ResumableResult<bool>
-	poll(Register reg, Predicate pred, uint16_t timeoutMs = 20, uint16_t stepMs = 4);
+	poll(Register reg, Predicate pred, const uint16_t timeoutMs = 20, const uint16_t stepMs = 4);
```
salkinium: These are passed by copy anyways, this isn't that useful.

salkinium (commented):
> The naming error was probably introduced when porting from xpcc to modm.

😳 😨 😱
Yeah, I was a little overzealous there.

src/modm/driver/position/vl53l0.hpp
strongly-typed: Yes, it is mostly cosmetic, but it makes sure that the value is not accidentally changed in the course of the method, e.g. `if (timeoutMs = 0)`. (That should trigger another compiler warning, btw.)

salkinium (dismissed):

src/modm/driver/position/vl53l0.hpp
chris-durand: Actually, `const` isn't useful here because you (accidentally) added it to the declaration not the definition :wink: According to C++ language rules this will have exactly no effects on anything. You probably meant to make the parameters const in the implementation file ([here](https://github.com/modm-io/modm/blob/e365959a4f52eb412f8a2eb128fdba8f6d825d91/src/modm/driver/position/vl53l0_impl.hpp#L640)).

salkinium: > top-level const and volatile on argument are not part of the signature, according to the standard

Wow, didn't know that. So you can't overload `func(const int)` and `func(int)` or `func(int *const)` and `func(int*)`, but you can use const in the implementation, but not the definition signature.

strongly-typed: Ok, fixed it and moved it to definition.

chris-durand (approved):

salkinium (approved):

PR #92, opened 2018-10-01
[wip] Add MPU9250 and Invensense embedded Motion Driver

salkinium:
Ports and integrates the MPU9250 IMU with the Invensense embedded Motion Driver 6.12 library.


salkinium:
Moved entirely to https://github.com/modm-io/Invensense-emd.

PR #93, opened 2018-10-02
Better support for generating modm:build:scons with multiple repos

salkinium:
Refactors to be able to better support generating build scripts for multiple repositories via `modm:build:scons`.

### DIFF
--- a/ext/arm/core.lb
+++ b/ext/arm/core.lb
@@ -31,4 +31,4 @@ def build(env):
     if "m0" not in core:
         env.copy("cmsis/CMSIS/Core/Include/mpu_armv7.h", "mpu_armv7.h")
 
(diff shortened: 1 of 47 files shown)


salkinium:
The `:build:scons` module now generates separate SConscripts per generated top level folder (which must map to repository name). It correctly partitions the sources and metadata between these files to build a separate static library for each repo, set include paths, linked libraries and packages.
The SConstruct is now aware of the top level output directories and ignores them for source searching and calls the SConscripts in the right order for building.

This should make it easier to build multiple lbuild repos with the modm build system. (cc @strongly-typed). This has been built for and tested with [modm-io/Invensense-eMD](https://github.com/modm-io/Invensense-eMD).

PR #102, opened 2018-10-14
Fix Doxygen Documentation Generator

salkinium:
This is a massive PR with a lot of small artisan changes to the documentation ~~followed by a huge whacking of trailing whitespace removals for good measure.~~

This makes the `modm:docs` module generate the doxygen module structure by defining its groups and including the module description in it. This way the module description can be shared between the `lbuild discover` commands and doxygen.

To support this however, all modm modules had to be refactored to name the right `@ingroup modm_module_submodule`. I've also added a lot of missing classes into their respective groups.
Every lbuild module now also has a description, or at least a description title.

With this PR you can `cd modm/docs; doxygen doxygen.cfg` and the fully featured documentation will appear magically under `modm/docs/html/` for your specific target.

cc @strongly-typed @daniel-k @mhthies @rleh @chris-durand


salkinium:
Here are all of the modules for the F469 blink example:

screenshot 2018-10-14 at 23 15 43 screenshot 2018-10-14 at 23 15 56 
screenshot 2018-10-14 at 23 16 07screenshot 2018-10-14 at 23 16 16

salkinium:
I'm removing the whitespace changes from this PR and putting them in a separate one, otherwise this becomes very messy.

daniel-k:
\o/

salkinium:
> \o/

Yeah, this makes modm much more accessible for third parties. I thought you'd like this ;-P
Now we only need to address the *level of detail* of documentation not its complete lack of. \*phew\*

salkinium:
Unfortunately, while applying whitespace removal I accidentally my whole `.git/objects/*` folder recursively.
So I lost some important scripts that I need to generate the homepage documentation. I plan to brute force the missing bits (space and tab values preceding newline, ie. trailing whitespace) to make the object files match their sha again. Wish me luck.

PR #103, opened 2018-10-14
[utils] Implement ArithmeticTraits<T> without hardcoded constants

chris-durand:
Implements `ArithmeticTraits<T>` using C++ standard library facilities instead of magic constants. This also fixes the incorrect assumption that `uint64_t` is a 32 bit type on AVR. 

`ArithmeticTraits<T>::WideType` is a bit tricky and requires some ugly template magic for a portable implementation. It relies on the fact that C++ guarantees `sizeof(char) <= sizeof(short) <= sizeof(int) <= sizeof(long) <= sizeof(long long)`.

### DIFF
--- a/src/modm/utils/arithmetic_traits.hpp
+++ b/src/modm/utils/arithmetic_traits.hpp
@@ -4,6 +4,7 @@
  * Copyright (c) 2012, 2014, Niklas Hauser
  * Copyright (c) 2013, 2015, Sascha Schade
  * Copyright (c) 2015, Kevin Läufer
+ * Copyright (c) 2018, Christopher Durand
  *
  * This file is part of the modm project.
  *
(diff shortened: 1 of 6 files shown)


salkinium:
Would you be willing to add a unittest for this? There wasn‘t one before, but it would give me more confidence in this.

salkinium:
Please also put the refactored ~~file~~ class into the correct doxygen group and perhaps add a little documentation if you want.

Please don't use the preprocessor to *exclude* stuff from doxygen, use `/// @cond` and `/// @endcond`.

chris-durand:
Now with test, less bugs and less ugly template stuff.

src/modm/utils/arithmetic_traits.hpp
```diff
@@ -56,243 +57,149 @@ namespace modm
...
-	typedef float SignedType;
-	typedef float UnsignedType;
+	static constexpr bool isFloatingPoint = std::is_floating_point_v<T>;
+	static constexpr bool isInteger = std::is_integral_v<T>
+		&& !std::is_same_v<std::decay_t<T>, bool>;
 
-	static constexpr bool isSigned = true;
-	static constexpr bool isFloatingPoint = true;
-	static constexpr bool isInteger = false;
+	using WideType = modm::WideType<T>;
```
salkinium, line 192: Ohh, nice, I like less typing for `modm::WideType`.

salkinium (dismissed):
Excellent work!

chris-durand:
Two test cases were commented out ...

salkinium:
Adding the AVR stdlibc++ was a great idea, I really love how this is working out!

PR #107, opened 2018-10-15
Remove utils/template_metaprogramming.hpp

chris-durand:
For all classes implemented in `utils/template_metaprogramming.hpp` a more mature implementation in the C++ standard library exists. This PR replaces all usages of `modm::SameType`, `modm::Select`, … with the corresponding implementations from the standard library. Thanks to C++14/17 `_t`, `_v` aliases for std type traits it will remove a lot of `typename`, `::value`, `::Result`, `::type`.

Also `utils/algorithms.hpp` is removed, which contains a basic version of a C++98 `std::find`.


salkinium (dismissed):
Very nice!

PR #109, opened 2018-10-17
RTOS: Added missing queue constructor

nesos:
The typesafe wrapper of the queue in the RTOS component is missing a constructor.

This fix will call the constructor of the underlying QueueBase class with the provided length parameter and the size of the templated type

### DIFF
--- a/src/modm/processing/rtos/freertos/queue.hpp
+++ b/src/modm/processing/rtos/freertos/queue.hpp
@@ -98,7 +98,9 @@ namespace modm
 		class Queue : private QueueBase
 		{
 		public:
-			Queue(unsigned portBASE_TYPE length);
+			inline Queue(unsigned portBASE_TYPE length):
+			    QueueBase(length,sizeof(T))
+			{};
 
 			using QueueBase::getSize;
 


src/modm/processing/rtos/freertos/queue.hpp
```diff
@@ -98,7 +98,9 @@ namespace modm
 		class Queue : private QueueBase
 		{
 		public:
-			Queue(unsigned portBASE_TYPE length);
+			Queue(unsigned portBASE_TYPE length):
```
salkinium: needs `inline`.

salkinium:
I saw you're using our FreeRTOS integration for your roswell project, just wanted to let you know that it's not really up-to-date (v7.x), and I've been meaning to update that for some time now.
I've started the [integration in this branch](https://github.com/salkinium/modm/tree/freertos), but haven't gotten through all the configuration yet.
Just in case you have some free time ;-P

salkinium (dismissed):
Thanks!

salkinium:
Squashed and rebased on develop.

PR #118, opened 2018-10-31
Enable STM32 extended I2c on many targets

chris-durand:
This renames and enables the L4 I2C implementation for many more targets. A prescaler calculator is implemented based on ST's beautiful excel sheet with VBA macros. This calculator deliberately violates conditions stated in the reference manual to simply do what ST does (except that it does not use half-french class names).

CAUTION: This is work in progress and not ready. On my machine trying to compile the Nucleo L476 i2c example eats up all ram. ST's brute-force approach of trying to check all possible `16*255*255` combinations seems to be too much for GCC's constexpr evaluation logic. Any ideas?


src/modm/platform/i2c/stm32-extended/i2c_timing_calculator.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+		for(int prescaler = 15; prescaler >= 0; --prescaler)
+		{
+			if(not (prescalerMask & (1 << prescaler))) {
+				continue;
+			}
+
+			for(uint16_t sclLow = 0; sclLow <= 255; ++sclLow) {
+				for(uint16_t sclHigh = 0; sclHigh <= 255; ++sclHigh) {
+					std::optional<float> speed = calculateSpeed(prescaler, sclLow, sclHigh);
```
salkinium: Can you guesstimate a sclLow/sclHigh combination and search upwards or downwards from there? Are there any bounds on the error? Is there a gradient of the error, you can then use for bisection?

salkinium:
Is the compile time still bad now with your improvements?

chris-durand:
> Is the compile time still bad now with your improvements?

No, that is solved. 

I compared the results of the calculator to CubeMX and it seemed to be correct. Sometimes the register values look vastly different but the timing will be nearly identical. In many cases there are a lot of near optimal solutions.

salkinium:
> No, that is solved

Very nice!

Can you rebase on top of develop, I think you need the new STM32 CMSIS headers for the CI?

chris-durand:
The I2C driver now compiles successfully on all F0, F3, F7, L0, L4.

I would really appreciate it, if someone could test it on F0/F3/F7/L0 hardware

The remaining errors in the CI are unrelated problems with the ADC driver on L41x/42x targets.
@salkinium Would it be ok for you if I fix them in this PR?

salkinium:
> Would it be ok for you if I fix them in this PR?

Absolutely fine with me.

I‘ll test some devices from my magic STM32 box and will post the results.

salkinium:
I tested the F3 disco accelerometer example with success, but since we don‘t have any off-the-shelf I2C examples for F0/F7, I didn‘t continue testing. Will do so tomorrow.

salkinium:
Works on F072 Discovery. Checking on F7 Disco next.

salkinium:
Working also on F7. I'm merging this once the CI passes.

salkinium (approved):
Fantastic work!

chris-durand:
Thanks for testing!

PR #123, opened 2018-11-11
Fix cmake builds on windows

fb39ca4:
Sets the compiler before the PROJECT call in cmake ensuring it tests the embedded toolchain and not the system compiler, which might be unavailable, especially on Windows.

Writes paths with forward slashes into CMakeLists.txt, because cmake cannot handle backslashes even on Windows.

### DIFF
--- a/tools/build_script_generator/cmake/cmake_scripts/configure-stm32-gcc.cmake.in
+++ b/tools/build_script_generator/cmake/cmake_scripts/configure-stm32-gcc.cmake.in
@@ -1,5 +1,3 @@
-INCLUDE(./modm/cmake/report-build-options.cmake)
-
 SET(CMAKE_SYSTEM_NAME Generic)
 SET(CMAKE_SYSTEM_PROCESSOR arm)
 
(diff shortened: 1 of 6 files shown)


salkinium:
#69 tried to fix a similar issue on Windows, can you check why you have different solutions to this problem? Does this require a different solution for different Windows releases?
Can you check if you can use the `windowsify` Jinja2 filter for this? I would like to ~~contain~~ maintain the crazy in one place 😄.

salkinium:
Cc @amarokmclion @daniel-k @siy

fb39ca4:
From what I understand, Windows accepts forward slashes in paths, but its command processor can confuse them for command parameters. I would recommend outputting forward slashes everywhere by default and only converting to backslash where things break on Windows.

salkinium:
> I would recommend outputting forward slashes everywhere by default and only converting to backslash where things break on Windows.

I've made the windowsify filter global (for `:build` and its submodules), so there is only one location that changes paths.

Can you check if changing that one filter to forward slashes fixes everything? I suspect that since the `openocd.cfg`/`gdbinit` files are probably passed via command line, the paths needs to be different than for CMake and Python?

salkinium:
This has been fixed in #204.

PR #124, opened 2018-11-11
 [build] [scons] Add command-line arguments to manipulate CCFLAGS, CFLAGS, etc.

rleh:

```
```

### DIFF
--- a/tools/build_script_generator/scons/resources/SConscript.in
+++ b/tools/build_script_generator/scons/resources/SConscript.in
@@ -154,4 +154,20 @@ env.Append(LINKFLAGS="$ARCHFLAGS")
 %% endif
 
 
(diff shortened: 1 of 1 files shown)


salkinium:
Same issues as in #60, plus in addition:

1. This will not work anymore when building multiple repositories with SCons, since the build flags are now split into their respective `repo/SConscript` and the modm SConscript is called first (and thus any changes done in it may be overwritten/undone).
2. This needs to go into a scons tool, perhaps just into the `env.BuildTarget()` tool, that already exists.

salkinium:
I'm sorry, but this isn't happening, we need to instead document how to modify the SCons environment via the SConstruct file inside the module documentation (and inside the generated SConstruct file).

PR #130, opened 2018-11-24
Use fully qualified names in lbuild configuration

dergraaf:
Use fully qualified names for modules and options. This is critical for the board files and the integration with other lbuild based libraries.

E.g. if the board file defines a value for ":target" and another library (other than modm) also defines a "target" option, then lbuild will fail. Using fully qualified names in the configuration files avoids this problem.

### DIFF
--- a/test/config/hosted.xml
+++ b/test/config/hosted.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <library>
   <modules>
-    <module>:platform:core</module>
-    <module>:test:**</module>
+    <module>modm:platform:core</module>
+    <module>modm-test:test:**</module>
   </modules>
 </library>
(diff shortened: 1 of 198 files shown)


test/config/hosted.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <library>
   <modules>
-    <module>:platform:core</module>
-    <module>:test:**</module>
+    <module>modm:platform:core</module>
+    <module>modm:test:**</module>
```
salkinium: The tests have all been moved to the modm-test repository in #110.

dergraaf: I've updated the commit.

salkinium (approved):
Thanks!

PR #135, opened 2018-12-19
Fix Un/SignedType for Vector classes

salkinium:
This is a followup of #103, which exposes `modm::SignedType<T>` and `modm::UnsignedType<T>` for use with `modm::Vector` since `ArithmeticTratis<Vector<T, N>>` won't work due to missing `std::numeric_limits<T>` overloads and it is the wrong abstraction for this anyways. (A vector doesn't have digits, or an epsilon, only its base type).

This is also fixing an issue in the RCA internal code base.

cc @rleh @se-bi @chris-durand


salkinium:
@chris-durand maybe you can have a look at this, I tried to specialize `modm::SignedType<Vector<T, N>>`, but that doesn't work the way I want it too.

src/modm/math/geometry/vector2.hpp
```diff
@@ -57,10 +57,6 @@ namespace modm
 	{
 		friend class Location2D<T>;
 
-	public:
-		typedef typename GeometricTraits<T>::WideType WideType;
```
chris-durand: This line is the issue. You removed the type alias which is used [here](https://github.com/modm-io/modm/blob/a33ee9d055b3ca1e53306bf67e4da9296bfebbf1/src/modm/math/geometry/vector2.hpp#L107). Thus, the compiler looks up the name `WideType` in an outer scope and finds the unrelated template `modm::WideType<T>`. It is trying to use it without template arguments and fails.

chris-durand (dismissed):

salkinium:
This fix escalated a little… do you think we should just completely eliminate the ArithmeticTraits class and just provide stand-alone definitions for digits and isInteger, like with SignedType? I think that would be cleaner, as the std library does this too.

salkinium:
> This fix escalated a little… do you think we should just completely eliminate the ArithmeticTraits class and just provide stand-alone definitions for digits and isInteger, like with SignedType? I think that would be cleaner, as the std library does this too.

I'm going to leave this up to future-self to fix and will merge this as is.

PR #136, opened 2019-01-03
Driver and example for PAT9125EL motion sensor

chris-durand:
Driver for PAT9125EL motion sensor. Only the I2C version is supported yet because I don't have access to the SPI version.

### DIFF
--- a/examples/nucleo_f429zi/pat9125el/main.cpp
+++ b/examples/nucleo_f429zi/pat9125el/main.cpp
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2018 Christopher Durand
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+
+#include <modm/processing/timer.hpp>
+#include <modm/processing/protothread.hpp>
+#include <modm/driver/motion/pat9125el.hpp>
+
+using I2c = I2cMaster1;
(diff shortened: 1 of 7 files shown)


chris-durand:
@salkinium Circle CI seems to be broken. The nucleo f429 blink example fails because of a missing file:

```scons: building terminated because of errors.arm-none-eabi-ar: /root/project/build/nucleo_f429zi/blink/release/modm/src/modm/platform/clock/latency.o: No such file or directory```

File: examples/nucleo_f429zi/pat9125el/main.cpp [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+Thread thread;
+
+// ----------------------------------------------------------------------------
+int
+main()
+{
+	Board::initialize();
+	Board::Leds::setOutput();
+
+	MODM_LOG_INFO << "\n\nPAM9125EL I2C example\n\n";
```
salkinium: PAM or PAT?

salkinium:
Hm, I recently upgraded the Docker container to gcc v8 for both avr and arm. Grrr.

examples/nucleo_f429zi/pat9125el/main.cpp
chris-durand: PAT is correct.

examples/nucleo_f429zi/pat9125el/project.xml
```diff
@@ -0,0 +1,13 @@
+<library>
+  <extends>modm:board:nucleo-f429zi</extends>
+  <options>
+    <option name="modm:build:build.path">../../../build/nucleo_f429zi/blink</option>
```
salkinium: This is the reason the CI is failing, this path needs to be unique, since we’re compiling in parallel and this overwrites the blinky project.

salkinium (approved):
Beautiful code as usual!

PR #139, opened 2019-01-07
[wip] Add support for STM32H7 family

salkinium:
Just an initial exploratory PR to see how much work it is to support the STM32H7 family, which has significant changes to its clock and power architecture.

cc @rleh @chris-durand


salkinium:
Closing this in favor of #379, which is more advanced except for the board module, which is however completely generic in this PR.

PR #129, opened 2018-11-21
Implement separate compile flags per module/file

salkinium:
This implements adding separate compile flags for specific files by attaching metadata to lbuild operations. This allows to compile CMSIS-DSP with its own flags (specifically `__FPU_PRESENT=1` is required), without using these settings globally (which triggers a warning since `<stm32xx.h>` device headers usually define this macro).

Implemented for:
- [x] SCons
- [ ] CMake

See #125 for discussion.
Depends on https://github.com/modm-io/lbuild/pull/16.
cc @dergraaf @rleh @se-bi @chris-durand

PR #144, opened 2019-01-16
[build] Generate SConscripts only for repositories

salkinium:
A repository may generate more that one folder, but the SConscript may
only be generated once, in the folder named like the repository name.


salkinium:
Moved into #129 instead.

PR #146, opened 2019-01-16
[scons] Fix MODM_GIT_SUBJECT escaping " quotes

salkinium:
What a silly bug, I wonder who wrote that code… 🤔


salkinium:
>  I wonder who wrote that code… 🤔

Unsurprisingly, it was me.
image

PR #147, opened 2019-01-16
[scons] Add instructions to add compiler options to SConstruct

salkinium:
Instead of #60 and #124, I've documented how to customize the default SConstruct to modify compile options.

cc @rleh


rleh (commented):
Great :+1:, but for a completely different use-case compared to #60 and #124.

rleh (approved):

PR #160, opened 2019-02-09
Fixes Hd44780 initialization issue #159

### LINKED ISSUE #159: hd44780 initialization fails because of writeHighNibble(uint8_t) call with RW set to RW::read
I tried to use a 16x2 a HD44780 display with modm/driver/display/hd44780.hpp in 4-bit mode and it would only display garbage after a call to initialize(). It seems that the problem was that, during the `inititalize()` method, `isBusy()` would indirectly call `RW::set(RW_Read)`  and, afterwards, a call to `Bus<DATA, E, DATA::width>::writeHighNibble(Set4BitBus)` would behave unexpectedly, since it requires `RW` set to `RW_Write`.

It is implemented in modm/driver/display/hd44780_base_impl.hpp

pedrofza:
Hd44780_base 's `initialize`call wouldn't work because `isBusy()` would call `RW::set(RW_Read)` and the following `Bus<DATA, E, DATA::width>::writeHighNibble(Set4BitBus)` call would require `RW` set to `RW_Write`.

### DIFF
--- a/src/modm/driver/display/hd44780_base_impl.hpp
+++ b/src/modm/driver/display/hd44780_base_impl.hpp
@@ -40,6 +40,7 @@ modm::Hd44780Base<DATA, RW, RS, E>::initialize(LineMode lineMode)
 	if (DATA::width == 4) {
 		while(isBusy())
 			;
+		RW::set(RW_Write);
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Good catch!

PR #161, opened 2019-02-09
Add support for qch generation

asmfreak:
This adds support to internally-used doxygen to create qch files, which, in turn, enables to use generated docs in QtCreator natively.
See example:
documentation usage example
In this example I've just tapped `F1` while having cursor on line with:
```cpp
Adc::setReference(Adc::Reference::InternalVcc);
```


salkinium (dismissed):
Oh wow, this is really cool, I had no idea this was possible!

salkinium:
(Rebased on top of develop)

PR #162, opened 2019-02-10
Fixed issue #156 regarding incorrect datatype in delay declarations

### LINKED ISSUE #156: Mismatched declaration and implementation of _delay_ms/us/ns on cortex
In src/modm/platform/clock/cortex/delay.hpp, these functions are declared as accepting uint32_t arguments, but the implementation in delay.cpp uses uint16_t. I noticed this when I enabled link-time-optimization and GCC warned it might not produce correct code. Was there a reason behind this or can it be changed?

ptrks:
Simple change to fix #156 and ensure declaration is correctly referencing uint16_t rather than uint32_t.

### DIFF
--- a/src/modm/platform/clock/cortex/delay.hpp
+++ b/src/modm/platform/clock/cortex/delay.hpp
@@ -20,9 +20,9 @@
 #include <stdint.h>
 #include <modm/architecture/utils.hpp>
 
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Excellent, thanks!

Issue #148, opened 2019-01-18
TODOs and Ideas


salkinium:
Some rough ideas about what needs to be done without having to open an individual issue for each one of them. Please feel free to comment or add more features.

## Library

- [ ] Run unittests in hardware on AVR and STM32 targets
- [ ] Fix AVR unittests
- [x] Automatic code formatting (via Clang-Format?)
- [x] Integrate [gdbgui](https://www.gdbgui.com)
- [ ] Move XPCC protocol out of modm into its own `xpcc-communication` repository.
- [ ] Refactor xpcc-generator into it's own proper python library
- [ ] Replace `modm:fatfs` with [ufat](https://github.com/dlbeer/ufat)
- [ ] Integrate [`upainter` graphics algorithms](https://github.com/salkinium/upainter)
- [ ] Integrate [fixed point math](https://github.com/MikeLankamp/fpm)
- [ ] Let user choose which memory regions to use as heap
- [x] assign linkerscript sections via memory attributes
- [ ] call same function on a set of pins (ie. access-optimized `setAlternateFunction()`)
- [ ] Implement Semihost logging for boards without serial or SWO output
- [ ] Port fmtlib (aka. `std::format` in C++20)
- [ ] Port [defmt offline formatting library from Rust](https://github.com/knurling-rs/defmt)
- [x] **Windows CI**
- [ ] Use-case for Peripheral::connect() methods: e.g. combinations of SPI signals
  - only MO, no SCK: WS2812 accelerated bit-bang
  - MI and SCK: read-in 74HC165
  - MO and SCK: write-out 74HC595
  - Some SPI-slaves have Open-Drain pins -> optional PullUp
- [ ] Hook [hardware events directly via generated ISR multiplexors](https://github.com/modm-io/modm/issues/248#issuecomment-511884427) (attach functions to GPIOs directly)
- [x] Add USB stack by integrating one of these projects:
  - https://github.com/hathach/tinyusb
  - https://github.com/dmitrystu/libusb_stm32
  - https://github.com/IntergatedCircuits/USBDevice
  - https://github.com/kevinmehall/usb
- [ ] Integrate MRI: https://github.com/adamgreen/mri
- [ ] Better [default configuration structs for drivers](https://github.com/roboterclubaachen/xpcc/pull/86#issuecomment-186578136)
- [x] Generate a peripheral support matrix for all target families from lbuild data to replace the non-sensical "★" table in the main README.
- [x] Write CI test to check if all examples are run by the CI
[...]

PR #168, opened 2019-02-17
[scons] Add Black Magic Probe Programming method (again)

strongly-typed:
This was added back in 2016 in 7f0d2ef, but disappeared when migrated to modm.

This shall solve #148 Add "programming capabilities/scons/Black Magic Probe"

- [x] Tested with STM32F1 blue pill
- [x] Add documentation to module.md
- [x] Add example in documentation
- [ ] ~~Add debugging method like `scons openocd-debug` / `scons bmp-debug`~~.
Postponed to a rework of the `upload` feature of modm.

### DIFF
--- a/tools/build_script_generator/module.md
+++ b/tools/build_script_generator/module.md
@@ -4,7 +4,7 @@ This parent module defines a common set of functionality that is independent of
 the specific build system generator implementation.
 This includes straight-forward options like project name and build path
 but also more complicated configuration for programming your target via
-AvrDude/OpenOCD and debugging via GDB.
+AvrDude/OpenOCD/Black Magic Probe and debugging via GDB.
 
 **Note that this module does not compile your project, you will need to choose
 the `modm:build:scons` or `modm:build:cmake` submodule for that, or provide
@@ -12,7 +12,7 @@ your own build system.**
 
 ## Using OpenOCD
 
-For accessing your ARM Cortex-M based device, we use OpenOCD and generate a
+For accessing your ARM Cortex-M based device, we use OpenOCD by default and generate a
 `modm/openocd.cfg` file with the target specific configuration:
 
 - Search directories passed via the `openocd.search` metadata key.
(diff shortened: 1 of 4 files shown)


salkinium:
> This was added back in 2016 in [7f0d2ef](https://github.com/modm-io/modm/commit/7f0d2efe80d4cf626fd7f2981c398e476920cda8), but disappeared when migrated to modm.

The issue is fundamentally that the SCons tools in xpcc didn't really deal with the user's programmer/debugger stack very well and I removed it in modm since there still wasn't a unified concept with how to deal with accessing targets.

I would like to see a concept for retargeting the existing semantics of the CLI onto the user's specific targets:

- `scons program` transparently uses either OpenOCD or BMP or DFU or STLink
- `scons gdb` transparently runs both GDB and programmer

This needs to work *WITHOUT* hardcoding information as an lbuild option, since the `black_magic_probe.port` is obviously going to be different for other users and OSes (the `:build:avrdude*` options are also questionable here).

Ideally I would like to see an automated guess as to what programmer to use, probably using pyUSB to check the connected programmers and before choosing one.
Alternatively we need to find a different solution that doesn't hardcode the BMP port in lbuild.

> Add documentation @salkinium: where to put?

Add it to `module.md` and use a better option description like the rest.

PS: Something with the CI is wrong, it seems to show an incomplete output from your own CI fork.

salkinium:
Alternatively to the automated detection, I'd also be ok with using arguments to specify the programmer and port:

```sh
scons program type=bmp  # automatically choose port for BMP
scons program type=bmp port=/dev/usbserial.foobar  # use specific BMP
scons program type=openocd usbid=serial  # in case your have multiple openocd programmers connected
scons program  # use OpenOCD
scons program  # in future: automatically choose programmer and port
```

This would just default `type` to openocd for now and can implement the automatic programmer choice later. This would be my preferred solution for now.
(Maybe `type` is too generic? I want something short to type.)

strongly-typed:
openocd != ST-Link, but so far this is the only combination we use. openocd offers many other [debug adapters](http://openocd.org/doc/html/Debug-Adapter-Configuration.html#Debug-Adapter-Configuration).

arm-none-eabi-gdb + Black Magic Probe is the other combination I am using.

Also to be considered:
* Deal with serial and CAN bootloaders 
  * at least the ones which are preprogrammed to STM devices
  * optionally, all 'official' bootloader protocols like app note AVR109 (which is commonly used by Arduino I think)

> I would like to see a concept for retargeting the existing semantics of the CLI onto the user's specific targets:
> * `scons program` transparently uses either OpenOCD or BMP or DFU or STLink
> * `scons gdb` transparently runs both GDB and programmer

I am in favour of
 * `scons upload` (or even `flash` if target is flash.)
 * `scons debug`

And `debug` can also be something else, not tied to the `gdb` tui forever. Saw some nice web-based tool.

How to proceed?

salkinium:
> * `scons upload` (or even `flash` if target is flash.)
> * `scons debug`
> And `debug` can also be something else, not tied to the `gdb` tui forever. Saw some nice web-based tool.

Ok, agreed. I never like `scons program` too much myself, `scons upload` is much better.

> Deal with serial and CAN bootloaders

Yes, good point! The intergrated STM32 bootloaders would be very nice to have.
There is also the programming of external Flash modules, for example the F469 Discovery board has an external QuadSPI Flash connected, which OpenOCD is still reviewing a patch for (since 2016!): http://openocd.zylin.com/#/c/4321/. There's also a generic bit-bang version: http://openocd.zylin.com/#/c/4760/.

I would very much prefer to delegate this to other projects (like OpenOCD or BMP) though, instead of duplicating this effort somehow in modm.

> How to proceed?

A "proper" solution requires larger changes that need to be planned and prototyped. I'm opening a new issue to keep track of this discussion.

For this PR I would the BMP functionality as the `scons bmp port=/dev/usbserial` format.
I wouldn't add the scons gdb functionality since it probably requires too many changes for now (like setting the correct port in `gbdinit.in` using the tools).

strongly-typed:
> For this PR I would the BMP functionality as the `scons bmp port=/dev/usbserial` format.

What is the best way to read the command line parameter in the script? Is there any example that uses command line parameters already?

salkinium:
See the user guide:
https://scons.org/doc/3.0.4/HTML/scons-user.html#sect-command-line-variables

strongly-typed:
```
tools/build_script_generator/scons/resources/SConscript.in
14:profile = ARGUMENTS.get("profile", "release")
```
Yeah ;-)

tools/build_script_generator/module.md
```diff
@@ -120,6 +120,70 @@ Program received signal SIGINT, Interrupt.
...
+Loading section .table.zero.intern, size 0x8 lma 0x80013e4
+Loading section .table.copy.intern, size 0xc lma 0x80013ec
+Loading section .table.section_heap, size 0xc lma 0x80013f8
+Start address 0x8000e6c, load size 5120
+Transfer rate: 10 KB/sec, 365 bytes/write.
+Detaching from program: , Remote target
+[Inferior 1 (Remote target) detached]
+scons: done building targets.
+
+```
```
salkinium, line 163: Please reduce the verbosity of this output and remove personal information like paths.
I did something very similiar for the OpenOCD output here: https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/module.md#using-openocd

salkinium (dismissed):
Thanks!

PR #173, opened 2019-02-21
Spi: remove incorrect assertions

rleh:
These assumptions are incorrect, here are some use-cases for not connecting *all* SPI signals:
* only MO, no SCK: WS2812 accelerated bit-bang
* MI and SCK: read-in 74HC165
* MO and SCK: write-out 74HC595
* Some SPI-slaves have Open-Drain pins -> optional PullUp


salkinium (approved):
Thanks!

salkinium (changes requested):
Hang on, there is another assertion in the STM32 UartSpiMaster ;-P

salkinium (approved):

PR #179, opened 2019-03-01
Add option to define an offset at the beginning of the flash memory for linking

dergraaf:
This useful for using bootloaders like the stm32duino bootloader which require the program to be linked to address 0x08002000 (instead of 0x08000000).

### DIFF
--- a/src/modm/platform/core/cortex/module.lb
+++ b/src/modm/platform/core/cortex/module.lb
@@ -11,8 +11,10 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # -----------------------------------------------------------------------------
 
+
 def common_vector_table_location(env):
     return env.get(":platform:cortex-m:vector_table_location", "rom")
+
 
 def common_vector_table(env):
     """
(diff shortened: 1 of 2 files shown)


src/modm/platform/core/cortex/module.lb
```diff
@@ -73,11 +75,34 @@ def common_memories(env):
...
+            # in order for the vector table relocation to work. Every interrupt
+            # requires 4 bytes, and therefore two additional bits of alignment.
+            number_of_interrupts = env.query(":::vector_table")["highest_irq"] + 16
+            bit_alignment = number_of_interrupts.bit_length() + 2
+            mask = (1 << bit_alignment) - 1
+            if flash_offset & mask:
+                raise ValidateException("Invalid flash offset in option '{}'. "
+                                        "The offset needs to be {} bit aligned."
+                                        .format("linkerscript.flash_offset",
+                                                bit_alignment))
```
salkinium: You may want to add a validate function and call the query there, so that this is raised in the right context. This is enough:
```py
def validate(env):
    env.query("vector_table")
```

salkinium (changes requested):
Nice!

src/modm/platform/core/cortex/module.lb
dergraaf: I've moved the complete validation to the `validate` step.

salkinium (approved):
Thanks!

PR #181, opened 2019-03-07
[literals] Remove bit/baudrate, tolerance enums

salkinium:
More aggressively deprecates bitrate, baudrate and tolerance enums in favor of user-defined literals.

@se-bi @chris-durand @rleh

### DIFF
--- a/src/modm/architecture/interface/i2c_master.hpp
+++ b/src/modm/architecture/interface/i2c_master.hpp
@@ -48,17 +48,6 @@ public:
 		Unknown				///< Unknown error condition
 	};
 
-	/// Baudrate of the I2C bus. Most slaves only work in Standard or Fast mode.
-	enum
-	Baudrate : uint32_t
-	{
-		LowSpeed =   10000,	///< Low-Speed datarate of 10kHz
-		Standard =  100000,	///< Standard datarate of 100kHz
-		Fast     =  400000,	///< Fast datarate of 400kHz
-		FastPlus = 1000000,	///< Fast Plus datarate of 1.0MHz (rarely supported)
-		High     = 3400000	///< Super datarate of 3.4MHz (rarely supported)
-	};
-
 	enum class
 	PullUps
 	{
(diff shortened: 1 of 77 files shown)


rleh (dismissed):

src/modm/architecture/interface/i2c_master.hpp
```diff
@@ -48,17 +48,6 @@ class I2cMaster : public ::modm::PeripheralDriver, public ::modm::I2c
...
-	enum
-	Baudrate : uint32_t
-	{
-		LowSpeed =   10000,	///< Low-Speed datarate of 10kHz
-		Standard =  100000,	///< Standard datarate of 100kHz
-		Fast     =  400000,	///< Fast datarate of 400kHz
-		FastPlus = 1000000,	///< Fast Plus datarate of 1.0MHz (rarely supported)
-		High     = 3400000	///< Super datarate of 3.4MHz (rarely supported)
-	};
-
```
salkinium, line 61: I've considered keeping this mapping from name to frequency, however, these names are mostly marketing names (like in USB) and in most datasheets the frequency is explicitly written in kHz. We also never used anything else but `Standard` in the examples.

chris-durand (dismissed):
Nice work :+1:

src/modm/math/units.hpp
```diff
@@ -0,0 +1,123 @@
...
+namespace modm {
+enum class percent_t : uint16_t {};
+template<typename T> constexpr percent_t pct(T value) { return percent_t(uint16_t(value * 600ul)); }
+constexpr float pct2f(percent_t value) { return uint16_t(value) / 60'000.f; }
+namespace literals
+{
+	constexpr percent_t operator""_pct(long double value) { return pct(value); }
+	constexpr percent_t operator""_pct(unsigned long long int value) { return pct(value); }
+}
+}
```
salkinium, line 93: So… about the `enum class percent_t`: C++ prohibits using a class as a type for template parameters, only integrals and enums are allowed as non-type parameters and `float` is explicitly forbidden (how would you specialize a template on a float value?).

However, percent is not passed as a unit of one, like frequency/baudrate/bitrate is, so integers or floats must first be converted, requiring a strong type, and enum class is the only type that can be used as a template parameter type. I've added the `pct2f` (percent-to-float) function as a compromise to normalize the unit as required for tolerance checking.

In [C++20 using non-type template parameters](https://en.cppreference.com/w/cpp/language/template_parameters) is made possible with the `<=>` operator, and thus I plan to replace this hack with a proper class type, which provides implicit conversions from/to float.

PR #184, opened 2019-03-16
Add support for 4x MAX7219 modules

dergraaf:
You can get very cheap 32x8 pixel LED matrices (less than 5€) based on 4x MAX7219 chips. This-pull request adds support for this arrangement of drivers. 

To avoid too many bit operations this also required a new buffered display class to compress the pixels in y-direction. The existing buffered display class has also been renamed to show that it only supports monochrome displays.

Example for a MAX7219 module:
Matrix


src/modm/driver/display/ks0108.hpp
```diff
@@ -41,7 +40,7 @@ namespace modm
 	 */
 	template < typename E, typename RW, typename RS,
 			   typename PIN_CS1, typename PIN_CS2, typename PORT >
-	class Ks0108 : public BufferedGraphicDisplay<128, 64>
+	class Ks0108 : public MonochromeGraphicDisplayBufferedVertical<128, 64>
```
salkinium, line 43:
```suggestion
	class Ks0108 : public MonochromeGraphicDisplayBufferedVerticalFactory<128, 64>
```
2wbvgn

src/modm/driver/display/max7219.hpp
```diff
@@ -0,0 +1,130 @@
...
+    static constexpr uint8_t MAX7219_REG_DIGIT3 = 0x4;
+    static constexpr uint8_t MAX7219_REG_DIGIT4 = 0x5;
+    static constexpr uint8_t MAX7219_REG_DIGIT5 = 0x6;
+    static constexpr uint8_t MAX7219_REG_DIGIT6 = 0x7;
+    static constexpr uint8_t MAX7219_REG_DIGIT7 = 0x8;
+    static constexpr uint8_t MAX7219_REG_DECODEMODE = 0x9;
+    static constexpr uint8_t MAX7219_REG_INTENSITY = 0xA;
+    static constexpr uint8_t MAX7219_REG_SCANLIMIT = 0xB;
+    static constexpr uint8_t MAX7219_REG_SHUTDOWN = 0xC;
+    static constexpr uint8_t MAX7219_REG_DISPLAYTEST = 0xF;
```
salkinium: Don’t you want to put this into an enum class called Register?

src/modm/driver/display/max7219_matrix.hpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+template <typename SPI, typename CS, uint8_t COLUMNS, uint8_t ROWS>
+void
+modm::Max7219Matrix<SPI, CS, COLUMNS, ROWS>::update()
+{
+    // Iterate column 0 to 7 of MAX LED driver
+    for (uint8_t ledCol = 0; ledCol < 8; ++ledCol)
+    {
+        uint8_t buf[COLUMNS * ROWS];
```
salkinium, line 93: this will blow out the stack quite easily. It would be interesting to compute this on the fly, I guess with an iterator or some kind.

salkinium (dismissed):
It’s ok I guess. my comments are just nitpicking.

salkinium:
Our UI system has never sparked any joy in me, I want to get rid of it and instead integrate [littlevgl](
https://littlevgl.com/) into modm. But of course that is completely outside the scope of this PR.

src/modm/driver/display/max7219_matrix.hpp
dergraaf: Done

src/modm/driver/display/ks0108.hpp
dergraaf: That are only 40 characters for that name, there is still room to grow :smile:

src/modm/driver/display/max7219_matrix.hpp
dergraaf: True, but those displays won't typically be very big. Something like 1..2 columns and 4..8 rows (each with 8 pixels). If you need larger displays, there are better ways to go.

With 2 columns and 8 rows, that would only be 16 bytes of RAM which should not overflow the stack.

salkinium (approved):
Thanks!

PR #187, opened 2019-03-26
[math] fix matrix vector multiplication & add unittest

se-bi:
Thanks to Markus Schmidt.

(The unittest may have some more tests ... )


salkinium (approved):
Thanks!

PR #191, opened 2019-04-09
Add minimal BNO055 driver

salkinium:
Adds a minimal implementation of the BNO055 driver, which only reads the results array in full NDOF mode.

### DIFF
--- a/src/modm/driver/inertial/bno055.hpp
+++ b/src/modm/driver/inertial/bno055.hpp
@@ -1,0 +1,660 @@
+/*
+ * Copyright (c) 2019, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_BNO055_HPP
+#define MODM_BNO055_HPP
+
+#include <modm/architecture/interface/register.hpp>
+#include <modm/architecture/interface/i2c_device.hpp>
+#include <modm/processing/resumable.hpp>
+#include <modm/math/geometry/vector3.hpp>
+#include <modm/math/geometry/quaternion.hpp>
+
+namespace modm
+{
+
+// forward declaration for friending with bno055::Data
+template < class Transport >
+class Bno055;
+
+/// @ingroup modm_driver_bno055
+struct bno055
+{
+	static constexpr uint8_t
+	addr(uint8_t offset = 1)
+	{
+		return 0x28 + (offset & 0b1);
+	};
+
+	enum class
+	Identifier : uint8_t
+	{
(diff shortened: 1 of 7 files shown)


src/modm/platform/i2c/stm32/module.lb
```diff
@@ -38,6 +38,7 @@ class Instance(Module):
         properties = device.properties
         properties["target"] = target = device.identifier
         properties["id"] = self.instance
+        properties["error_handler"] = False
```
dergraaf: Is it useful to always disable this? Either this should be configurable or the code should be removed from the driver. Otherwise it is just dead code with a layer of obfuscation.

src/modm/driver/inertial/bno055.hpp
```diff
@@ -0,0 +1,660 @@
...
+ * @author	Niklas Hauser
+ */
+template < class I2cMaster >
+class Bno055 : public bno055, public modm::I2cDevice<I2cMaster, 2>
+{
+public:
+	/// Constructor, requires a bno055::Data object.
+	/// For I2c this also sets the address to 0b110101 (alternative: 0x1C).
+	inline Bno055(Data &data, uint8_t address=addr()):
+		I2cDevice<I2cMaster,2>(address), data(data) {}
```
dergraaf, line 588: `data(data)` is very error prune, as the compiler will still be happy if you rename the argument but don't change this line. I suggest either an `m` prefix for the member (what I've been using lately) or an `_` suffix for the parameter, that seem to be the most popular options. Or rename either of the two to something different.

src/modm/driver/inertial/bno055.hpp
```diff
@@ -0,0 +1,660 @@
...
+
+		buffer[0] = reg;
+		this->transaction.configureWriteRead(buffer, 1, (uint8_t*)&data, Data::size);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+	inline modm::ResumableResult<bool>
+	updateRegister(Register regi, uint8_t setMask, uint8_t clearMask = 0xff)
+	{
+		uint8_t reg = uint8_t(regi);
```
dergraaf: This should be `const`. If it would be changed anywhere, the data flow would be become very fragile with the `RF` macros.

src/modm/platform/i2c/stm32/module.lb
salkinium: I’m not sure yet, I just needed a quick fix. Our error handling is wrong though, we stop on a bus error, which isn’t what the manual says you should do (it says ignore).

src/modm/driver/inertial/bno055.hpp
salkinium: Yeah, I’d like to use the Qt coding conventions in general: `m_` prefix and just `obj.data()` as getter. But, yeah, conventions…

src/modm/driver/inertial/bno055.hpp
salkinium: This API isn’t done yet. But the datasheet is a bit terrible (typical for Bosch).

PR #195, opened 2019-04-24
[ext] Fix missing define in our FreeRTOS port

salkinium:
Our FreeRTOS port needs to be nuked from orbit. Repeatedly!

This adds the required macro to get the port running on STM32F4 for our hacked together custom port.

cc @rleh @dergraaf


salkinium:
I've started work on porting a:FreeRTOS: https://github.com/salkinium/modm/commit/4e70d71dc3d27bfdedb1d6d272819fd155fe9446

But it's very difficult to wade through the jungle of configurations and myriads of porting files. It's almost like they don't want you to use this. 🤔

PR #198, opened 2019-04-26
Update to a:FreeRTOS

salkinium:
As a follow-up to #195 and long-standing issues, this upgrade the `modm:freertos` module to FreeRTOS v10 with MIT license from AWS.

~~The submodule will change to be slimmer, since it's downloading about 100MB of stuff right now 🙄.~~
Added [modm-ext/freertos-partial](https://github.com/modm-ext/freertos-partial) repository.


salkinium:
We only have one RTOS example on STM32F4 Discovery and it works fine with these changes.

salkinium:
Added examples for the Cortex-M0(+), Cortex-M3, Cortex-M4F, Cortex-M7F ports and tested them in hardware.

PR #203, opened 2019-05-05
[lbuild] Update for lbuild v1.11.0

salkinium:
Main change is just the unification of module naming, since using `module.parent` is deprecated now.

For NumericOptions, you can put Python code to simplify your values (Like 16MB -> `2**24` instead of `16777216`).

Filters must be prefixed with the repository name now.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -53,8 +53,6 @@ markdown_extensions:
   - pymdownx.caret
   - pymdownx.critic
   - pymdownx.details
-  - pymdownx.emoji:
-      emoji_generator: !!python/name:pymdownx.emoji.to_svg
   - pymdownx.inlinehilite
   - pymdownx.keys
   - pymdownx.magiclink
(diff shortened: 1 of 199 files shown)


src/modm/board/disco_f469ni/board.xml
```diff
@@ -9,7 +9,7 @@
     <option name="modm:target">stm32f469nih</option>
     <option name="modm:platform:uart:3:buffer.tx">2048</option>
     <option name="modm:platform:cortex-m:allocator">tlsf</option>
-    <option name="modm:tlsf:minimum_pool_size">16777216</option>
+    <option name="modm:tlsf:minimum_pool_size">2**24</option>
```
dergraaf: I'm not sure that this is much more readable to be honest. Something like `16MiB` would be better.

salkinium: Agreed, SI prefix (plus binary prefix) support would be better!
Maybe we can "just" do a replace on the input string? Like `input.replace("Mi", "*2**20")`? and then for every prefix we want?

salkinium: Removed this for now, I think there needs to be a better solution.

PR #205, opened 2019-05-05
[scons] Fix avrdude programming when no port is configured

Sh4rK:
Using an AVRISP mkII, on Windows, I don't need to set the port for the programming to work.


salkinium (approved):
Thanks!

salkinium:
This works on OS X too. Didn't know that!

PR #213, opened 2019-05-11
[platform] Fix USART interrupts on a few AVRs

### LINKED ISSUE #211: UART RX interrupt name not correct for ATmega8
Hi again!

A problem has come up with the UART on ATmega8. [The UART template](https://github.com/modm-io/modm/blob/66c0868dcf8a4b05105ebffacb9acc63af6ecbac/src/modm/platform/uart/at90_tiny_mega/uart_rx.cpp.in#L32) assumes that the RX interrupt name is `USART{{ id }}_RX`, but on the mega8 it is `USART_RXC`.

According to the [avr-libc docs on interrupts](https://www.nongnu.org/avr-libc/user-manual/group__avr__interrupts.html) the devices that use this interrupt name are ATmega16, ATmega32, ATmega323, ATmega8.

What's the correct way to fix this?

Sh4rK:
Fixes #211.

### DIFF
--- a/src/modm/platform/uart/at90_tiny_mega/uart_defines.h
+++ b/src/modm/platform/uart/at90_tiny_mega/uart_defines.h
@@ -85,12 +85,19 @@
 #	define USART0_TX_vect		USART_TX_vect
 #endif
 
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Thanks!

PR #214, opened 2019-05-13
[scons] Set default `tools` to empty

Sh4rK:
Sets the default environment to not include any tools automatically. The
correct tools are manually added in the generated SConscript files.

Without this, scons will try to find and set up a default toolchain, and
this takes quite a long time on Windows with MSVC.


salkinium (approved):
Very nice! Even shaves off 0.2s on my machine.

PR #215, opened 2019-05-13
Use lowest priority for undefined IRQs and disable them

salkinium:
This allows them to be preempted by other IRQs especially by the UART interrupts used by the buffered IODevice in modm_abandon to report the abondonment.
This also disables the IRQ so that it doesn't spam the assertion handlers.

cc @rleh

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -94,7 +94,7 @@ make gdb
 
 ## Interesting Examples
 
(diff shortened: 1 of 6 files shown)


salkinium:
Won't work for HardFault_Handler, since it has a hardcoded priority of -1. You need to use `:platform:fault` for that.

PR #153, opened 2019-01-23
Add driver for AB (incremental) encoder output

rleh:
This driver generates a AB (incremental) encoder signal to emulate a hardware encoder with a microcontroller.

---

- [x] Example
- [x] Test on real hardware
- [x] Add a higher-resolution (than 1ms) `Timeout`
- [ ] Unittest?
- [x] Rename to `EncoderOutput`.

PR #217, opened 2019-05-15
Refactor modm:processing to be std::chrono compatible

### LINKED ISSUE #188: std::chrono literals with periodic timers
It would be nice to do something like this: 

```cpp
#include <modm/board.hpp>
#include <modm/debug/logger.hpp>
#include <modm/processing/timer.hpp>
#include <chrono>
using namespace modm::literals;
using namespace std::literals::chrono_literals;

int
main(){
    auto messageTimer = modm::ShortPeriodicTimer{10ms};
    while(True){
        if(messageTimer.execute()){
            // do stuff
        }
    }
}
```

Is it possible?

For now I'm using 
```cpp
template<typename timestamp_t= modm::ShortTimestamp, typename...T>
constexpr timestamp_t d2t(std::chrono::duration<T...> c){
    return {
        static_cast<typename timestamp_t::Type>(
                    std::chrono::milliseconds(c).count())};
}
```

salkinium:
This extracts the std::chrono effort from #153 and optimizes it a bit.
Closes #188.

Unfortunately std::chrono uses signed 64-bit for all durations and clocks, which makes wrapping on overflow a messy operation. So I've kept the unsigned 16-bit and 32-bit timestamps, clocks and timers. However since both clocks are not steady, I've named them `modm::chrono::milli_clock` and `modm::chrono::micro_clock` (and aliased them to `modm::Clock` and `modm::PreciseClock`) to distinguish them from the `std::chrono::steady_clock` and `std::chrono::high_resolution_clock` and their use-cases.

Microsecond clocks are implemented via timer overflow counters, and division is avoided by shifting a fixed constant multiplication, resulting in a very accurate fractional division that keeps error under 1us on both Cortex-M (via single cycle 32x32=32-bit (ARMv6-M) or 32x32=64-bit (ARMv7-M) multiplication) and AVR (via single cycle 16x16=16bit multiplication).

On Cortex-M the 24-bit SysTick counter is used to implement a 250ms interrupt, instead of a 1ms interrupt, and instead both milli- and microseconds are computed via the above mechanism. This reduces the number of interrupts to only 4Hz, without reducing accuracy. This 250ms period works for a clock up to 536MHz.
Since ARMv6-M only has 32x32=32bit multiplication, the interrupt is kept at 1kHz and the microsecond computation is adapted to fit into just 32bit for max 1000us, which works just fine.

On AVR a default clock implementation using the 8-bit TimerCounter0 provides a 1kHz milliseconds clock and via the overflow a microsecond clock, which however does not have 1us step size, due to only being a 8-bit counter, thus having less than 256 steps per 1000us.
On a 16MHz clock, there is a 4us resolution with a counter overflow of 250 ticks and prescaler of 64.

All clock implementation have been tested especially for the property of only ever jumping forward (except during overflow of course), which had to be done without atomic lock and opportunistic polling, since the timers can still roll over even during an atomic lock, leading to interlacing.

The new clock implementations of `now()` are linked weak, so that their implementation can be overwritten. This is used in the unittests, so that the `modm:platform:core` modules do not need to be aware of the unittests and that the user can overwrite the implementation to their needs (for example using a Low-Power Timer instead of Systick, so the clock continues running during sleep).

The `modm:processing:timer` classes are refactored to use the new clocks as well as accept `std::chrono::duration` values. Passing Integers is deprecated (should it be?) and all uses inside modm have been refactored.

~~Here it is more difficult to deprecate the interger constructors and `restart(interval)` functions, since the signatures are the same unlike `modm::delay(std::chrono::{nano, micro, milli}seconds` vs `modm::delay_{ns, us, ms}(uint32_t)`.  Here the constructor only has a type overload `modm::Timeout(std::chrono::{micro, milli}seconds)` vs `modm::Timeout(uint32_t)`.
So this *requires* a cast in user code, like so: `modm::Timeout(std::chrono::milliseconds(uint32_t))`, which is quite verbose and I found it annoying to use when not having chrono literals at hand (like in all modm header-only drivers).~~
Update: I added the `using namespace std::chrono_literals` to the modm namespace, so any drivers inside modm can simply use chrono literals (and modm literals) without hassle.

TODO:
- [x] Milli- and Microsecond clocks
- [x] Implemented for Cortex-M SysTick
- [x] Implemented for AVR
- [x] Implemented for Hosted
- [x] Using (unsigned) 32-bit for everything
- [x] Adapt Timeout, PeriodicTimer and Duration/Timestamp
- [x] Any way to get rid of the division? ARMv6-M doesn't have `udiv` instruction.
- [x] Unit tests (particularly for overflow of µs)
- [x] Check microsecond clock uses for AMNB and Radio
- [x] Check The Logger thingy
- [x] Better documentation

cc @rleh @ASMfreaK @chris-durand @strongly-typed @se-bi


salkinium:
This is pretty broken due to replacing unsigned arithmetics with signed arithmetics which messes up the checks for expiration after overflow (due to comparing to "negative time").

PR #219, opened 2019-05-30
Add GNU Build ID to identify firmware

salkinium:
This uses the GNU Build ID to properly identify firmware and uses this mechanism instead of CRC32 for the coredump functionality added in #52.

Adds 32 bytes containing a 160-bit SHA1 to the firmware.
Inspired by https://interrupt.memfault.com/blog/gnu-build-id-for-firmware.

cc @dergraaf: This solves my dependence on `objcopy` and reads the data directly from the ELF file.

### DIFF
--- a/examples/nucleo_f103rb/hard_fault/main.cpp
+++ b/examples/nucleo_f103rb/hard_fault/main.cpp
@@ -54,12 +54,11 @@ main()
 	if (FaultReporter::hasReport())
 	{
 		MODM_LOG_ERROR << "\n\nHardFault! Copy the data into a 'coredump.txt' file, ";
-		MODM_LOG_ERROR << "then execute 'scons postmortem firmware=";
-		MODM_LOG_ERROR << modm::hex << FaultReporter::firmware() << "'.\n\n";
-		for (const uint8_t data : FaultReporter())
-		{
+		MODM_LOG_ERROR << "then execute\n\n\tscons postmortem firmware=" << modm::hex;
+		for (const auto data : FaultReporter::buildId()) MODM_LOG_ERROR << data;
+		MODM_LOG_ERROR << "\n\n";
+		for (const auto data : FaultReporter())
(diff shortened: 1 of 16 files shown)


salkinium:
The build id should also be added for AVRs

strongly-typed:
Are you aware that for the NXP chips there shall be a checksum in the hexfile?
openocd complains (and fixes it before writing to the flash). Verification fails

```
** Programming Started **
Warn : Verification will fail since checksum in image (0x000004e9) to be written to flash is different from calculated vector checksum (0xefffd63e).
Warn : To remove this warning modify build tools on developer PC to inject correct LPC vector checksum.
** Programming Finished **
```

Could the calculation of this checksum already be added here?

salkinium:
What [fresh hell is this?!?](https://community.nxp.com/thread/389046)
This can’t work of course since both checksums would change each other.
Maybe putting the GNU ID section to the very end could help? Checking the source code for GNU LD for NXP targets could help, since this checksum must be explicitly coded?

salkinium:
> Note: This checksum is calculated from the contents of the vector table, not the whole image. For more details please see the user manual for the MCU that you are using.

But hang on, the GNU ID doesn’t change the vector table, the 0x4e9 value is probably the address to the Undefined_Handler. The checksum is probably never computed, so this seems like a separate issue.

salkinium:
> The build id should also be added for AVRs

I didn't get this to work. The `.note.gnu.build-id` section doesn't contain any program headers, so avrdude doesn't load this section. Unfortunately all AVR linkerscripts predefine this section outside of the flash memory, so I cannot overwrite that with our custom linkerscript addition.
For `hosted-darwin` the linker does not support GNU LD flags anyways, and for `hosted-linux` I cannot reference the section, since the assembler does not accept the hyphen in `build-id` from the section name 🙄.

I'll leave this unimplemented for AVR and Hosted (for now).

Issue #125, opened 2018-11-12
Support separate compile flags per module/file in CMake


salkinium:
Some external code may not adhere to the modm coding standards and may generate warnings which may lead to errors due to our general compile flags. This is especially annoying when the external source code cannot be edited due to loosing upstream compatiblity or violating licencing terms.

It would be useful to compile parts of modm and other repositories with different compile flags.
This needs to be done carefully, since some flags are required for the platform (like `flags.archflags`) and must be used for all files (even in other repositories) and some flags may even contain conflicting definitions for separate files (like `flags.cppdefines`).
It is unclear how to define the semantics for choosing what build flags to use for user code, since currently all files in SCons's scope are build with the *union* of all repos/modules of all compile flags.

This will require a new feature in lbuild, for defining `metadata` per file too, not just per module.

#### Update 2019-01-16

SCons support has been implemented in #129.  
However CMake support is lacking!

salkinium:
cc @rleh @strongly-typed @daniel-k @dergraaf @chris-durand @mhthies

dergraaf:
> This will require a new feature in lbuild, for defining metadata per file too, not just per module.

A big objection from my side. If we add separate compile flags per file, we end up where we started with the xpcc build system. This is out of scope for lbuild.

> This is especially annoying when the external source code cannot be edited due to loosing upstream compatiblity or violating licencing terms.

The better solution is then to try to fix the issue upstream or wrap the code into a compatible wrapper.
My suggestion would be to separate problematic libraries as much as possible (different folders) and only have different compile flags at that scope (and try to avoid it as much as possible).

salkinium:
>> This will require a new feature in lbuild, for defining metadata per file too, not just per module.

> A big objection from my side. If we add separate compile flags per file, we end up where we started with the xpcc build system. This is out of scope for lbuild.

The API would just extend the file operations and attach the metadata to a buildlog operation:

```py
env.copy(..., metadata={"key":"value"})
env.template(..., metadata={"key":"value"})
env.archive(..., metadata={"key":"value"})
```

lbuild already [stores metadata by module](https://github.com/modm-io/lbuild/blob/develop/lbuild/buildlog.py#L87-L117) and provides accessors for global, repository or module level. This would just extend the concept to file operations as well.

[...]

PR #220, opened 2019-05-30
Add option to generate compilation db

asmfreak:
CLion cannot use CMakeLists.txt, generated by modm due to custom build locations.
To enable ourselves to use CLion in projects this little addition is needed.
See [this example project](https://github.com/ASMfreaK/modm-example-clion-project) to understand how it is used.


salkinium (dismissed):
Thanks!

salkinium:
Compilation Database is something I would like to generate directly from a new submodule of `modm:build`, but didn’t find example code for it. I guess I have to check out the CMake or SCons source code. See also:  https://github.com/modm-io/modm/issues/125#issuecomment-442077244

asmfreak:
As for SCons see how [mongo](https://github.com/mongodb/mongo/blob/eea0ef850ef7f07e354a466e011b39ab518a5254/site_scons/site_tools/compilation_db.py) implemented this. Maybe we can import that code directly (if license allows)?

salkinium:
I’m mobile-only due to GPN19, so I’m not expecting any further features for this PR, just need my laptop. 😛 (just in case you were wondering why I’m not merging this)

asmfreak:
I seriously needed comp_db in scons builds, so I pulled mongo's design I linked earlier. I also wanted to use `modm:unittest` to test my code, which have lead to some backward-compatible changes in `scons/module.lb` and associated files.
If it's too much, I'll revert it back.
Also licensing question - can we use mongo's solution? I've left copyright intact and added my copyright as I made changes to their code to prevent scons from crashing with `NotImplementedError`.

salkinium:
Hm, the GPL is technically compatible with MPL, I just don't really want it in the codebase. But I do see the value this adds to SCons.

I like the unittest changes, however, the SCons unittest tool is just a glorified Python script with a few regexes, that I would like to extract into a stand-alone small Python script.

I'm going to have a look at these two issues. Is it ok for you to keep these changes a little longer in your fork?

asmfreak:
It's ok with me.
I'd be happy to see it merged some day.

чт, 6 июн. 2019 г. в 22:56, Niklas Hauser <notifications@github.com>:

asmfreak:
The `compilation_db` tool included in this PR for scons no longer works. Some updates in modm and/or scons rendered it unuseable.

salkinium:
Ok, lemme check what I broke this time.

asmfreak:
I don't know scons well enough. I blame the script.
I even tried [an updated version from mongo](https://github.com/mongodb/mongo/blob/master/site_scons/site_tools/compilation_db.py).
I doesn't break, but it does not output a single thing, just `[]`
It generated *some* output, once I've put it like this
```
...
sources.append(env.InfoGit(with_status=True))
sources.append(env.InfoBuild())
# Long comment by lbuild
# ...
env.Tool("compilation_db")
env.BuildTarget(sources)
env.Alias("compilation_db", env.CompilationDatabase("compile_commands.json"))
```

It outputted only lines for main and files with buildinfo.

salkinium:
I'm trying it in the `example/nucleo_f411re/blink` and it only outputs how to compile `main.cpp`, nothing about modm.

salkinium:
>  it only outputs how to compile main.cpp, nothing about modm.

Cos I'm stupid and this needs to be placed at the end of the tools in the modm/SConscript, but BEFORE the static library is compiled. Ugh, this is horrible.

salkinium:
Specifically the `env.Tool("compilation_db")` needs to be added after the `settings_buildpath` tool to work.

salkinium:
Just to clarify: Are you able to compile your code with CompilationDB? Cos it's missing not only assembly files (`.s`, `.sx`, `.S`), but also the linker steps with our custom linkerscripts and linkflags.

asmfreak:
I don't need ConpilationDB to compile code. I only use it to get code completion and code insight for IDE. It was working ok  since I created this pull request, but stopped working after I updated modm. I even thought of running scons under strace, logging execve's and reformatting this log into CompilationDB json. I use scons for hosted linux compilations and hosted unit tests. I might just migrate that code to use cmake

salkinium:
Ok, so here's the deal. I'll create a seperate `modm:build:compilation_db` module, that generates the CompilationDB manually. The MongoDB tool isn't dealing with Assembly files and Linking, and I would like to see this as a proper build system, especially since many more IDE start to support it.

In the meantime, I've added a simple commit on this branch, that re-adds this tool correctly: https://github.com/salkinium/modm/tree/feature/compilation_db

salkinium:
I'm closing this in favor of #293.

PR #227, opened 2019-06-07
Fix ADC on Blue-Pill

asmfreak:
This fixes the issue of using ADC on boards like Blue Pill, which have `Adc1`, `Adc2` and `Adc3` in `SystemClock`. The ADC driver uses `SystemClock::Adc`.

I don't undestand which one (`Adc{{id}}` or just `Adc`) is the normal behavior, so this may not be the right place to fix this.

### DIFF
--- a/src/modm/board/black_pill/board.hpp
+++ b/src/modm/board/black_pill/board.hpp
@@ -31,9 +31,7 @@ struct SystemClock {
 	static constexpr uint32_t Apb1 = Frequency / 2;
 	static constexpr uint32_t Apb2 = Frequency;
 
-	static constexpr uint32_t Adc1 = Apb2;
-	static constexpr uint32_t Adc2 = Apb2;
-	static constexpr uint32_t Adc3 = Apb2;
+	static constexpr uint32_t Adc  = Apb2;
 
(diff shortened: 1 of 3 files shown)


salkinium:
It depends on whether the ADC instances share the same clock source or difference ones. This may be specific to the device too. I‘ll check this out. [Here are a few STM32 clock trees](https://salkinium.com/stm32/clock/).

salkinium:
Hm, all ADC instances share the same prescaler. It‘s a bit weird that the initialize method is part of every ADC instance, it should be a global  Otherwise initializing Adc1 and then Adc2 will overwrite the previous settings. That API is not thought out well 😞

salkinium:
The quickfix for this is to just have one `SystemClock::Adc` value in the blue pill BSP file.

salkinium:
I've changed your commit to fix the `SystemClock::AdcX` -> `SystemClock::Adc` clock source name instead.

salkinium:
Can you check if this fixes your issues?

asmfreak:
I'll check it today

asmfreak:
I confirm this fixing the issue

salkinium (approved):
\o/

PR #229, opened 2019-06-23
[ci] Redesign deployment of docs using workspaces

strongly-typed:
That shall push the new docs to modm-ext on every CI run on develop. It will always build the docs for all branches. Redundant code of docs creation was removed.

The workflow for develop now looks like so:

image

The final step failed as this was on my repository without the keys to push to modm-ext

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,6 +254,7 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Update submodules and install lbuild
           command: |
             (git submodule sync && git submodule update --init --jobs 8) & pip3 install -U lbuild & wait
       - run:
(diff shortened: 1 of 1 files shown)


strongly-typed:
Ah, deleted the wrong branch ...

.circleci/config.yml
```diff
@@ -263,37 +264,35 @@ jobs:
...
             - "8c:e7:b6:51:2a:40:66:e6:e9:50:ea:fe:38:49:1d:3f"
       - run:
+          name: Clone Old Docs
           command: |
-            (git submodule sync && git submodule update --init --jobs 8) &
-            pip3 install -U lbuild &
-            git clone git@github.com:modm-ext/modm.io.git docs/modm.io &
-            wait
+            git clone --depth 1 git@github.com:modm-ext/modm.io.git docs/modm.io
+            cp -R /tmp/docs/modm.io/ docs/
```
salkinium, line 286: Does this delete files?

salkinium (commented):
Hm, not super convinced that this is indeed an improvement…

.circleci/config.yml
strongly-typed: You are right. We should delete the directory (except the .git subdir) before, like
```
find -not -path "./.git/*" -not -name ".git" -delete
```
Or do the checkout step in the `build-docs` step.

strongly-typed: The easy way is the `--no-checkout` option of git ...

strongly-typed:
> Hm, not super convinced that this is indeed an improvement…

Ok, let me try to convince you: The old code:
* Had duplicated code in the CI, which is difficult to maintain.
* As a result, might have pushed different docs than were created in the step before and reviewed in the PR.
* Needed some magic to distinguish the branches (`only: /^pull\/.*$/`)
* as it relied on a naming scheme for the non-develop branches
* Was testing if building for develop twice

The new code:
* Clearly separated build and delivery step into `build-docs` and `upload-docs` as a CI/CD pipeline should do
* Moves the output of one build step to the other by the designated mechanism of `workspaces`
* There could be step in between, like spell checker on the generated docs
* No duplicated code
* `build-docs` is done for every build (which was the desired behaviour)
* `upload-docs` is only done for `develop`

A downside still is that the dependencies of `upload-docs` must be manually edited when another job is added to the CI.

What are your concerns?

salkinium:
Ok, that's fine. Just seemed overly complex with the workspace, but it's not that bad on detailed look. Pls rebase.

strongly-typed:
Thanks for your feedback, much appreciated. I just rebased and hope it will work fine on develop here when you merge. 🤞

PR #235, opened 2019-07-02
Fixes stlink version

### LINKED ISSUE #234: Bluepill cannot be programmed with stlink v2.1
The config, generated for bluepill specifies v2, which, for some very bizarre reason, worked on Windows with stlink v2.1.
It didn't work on Linux though.

asmfreak:
This fixes #234 by introducing a new configuration option to respectful boards

### DIFF
--- a/src/modm/board/black_pill/module.lb
+++ b/src/modm/board/black_pill/module.lb
@@ -33,6 +33,13 @@ def prepare(module, options):
         ":platform:clock",
         ":platform:core",
         ":platform:gpio")
+
+    module.add_option(
+        EnumerationOption(
+            name="stlink_version", enumeration=["2", "2-1"],
+            default="2",
+            description="# Version of stlink to use"
+        ))
     return True
 
 def build(env):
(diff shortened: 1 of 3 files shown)


src/modm/board/black_pill/module.lb
```diff
@@ -42,5 +49,5 @@ def build(env):
     env.copy('.')
 
     env.outbasepath = "modm/openocd/modm/board/"
-    env.copy(repopath("tools/openocd/modm/stm32f103_blue_pill.cfg"), "stm32f103_blue_pill.cfg")
-    env.collect(":build:openocd.source", "modm/board/stm32f103_blue_pill.cfg");
+    env.template(repopath("tools/openocd/modm/stm32f103_blue_pill.cfg"), "stm32f103_blue_pill.cfg")
```
salkinium, line 52:
```suggestion
    env.template(repopath("tools/openocd/modm/stm32f103_blue_pill.cfg.in"), "stm32f103_blue_pill.cfg")
```

salkinium:
This isn't the right way to do it, since the BSP shouldn't have to care about how the hardware is programmed, but I currently don't have an alternative, so… this is fine? It does solve your problem, and doesn't brake my use-case (I use the cheap STlink v2 clones to program this board).

The current upload functionality is quite limited, we would like to rework it in a more complete way: https://github.com/modm-io/modm/pull/168#issuecomment-464459385

salkinium:
Closed in favor of #240.

PR #237, opened 2019-07-05
[scons] Add source/runner options for unittest tool

salkinium:
This expands on an idea from @ASMfreaK in #220 to make the unittest runner accessible for outside projects too.

I think the unittest tool should be extracted into it's own python tool, and then wrapped by SCons like the XPCC tools. Then the unittests can also be integrated into any other build system or just used directly via python.

### DIFF
--- a/test/config/al-avreb-can.xml
+++ b/test/config/al-avreb-can.xml
@@ -2,6 +2,8 @@
 <library>
   <extends>modm:al-avreb-can</extends>
   <options>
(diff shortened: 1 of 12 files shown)


salkinium:
@ASMfreaK does this work for your use-case? Or does it break something for you?

asmfreak:
I'll try this on the following week. It looks good for my use-case

asmfreak (approved):

asmfreak:
Sorry, I just wanted to try out the reviews feature.

PR #240, opened 2019-07-10
[openocd] Do not hardcode user options in built-in configs

### LINKED ISSUE #234: Bluepill cannot be programmed with stlink v2.1

### LINKED ISSUE #238: Support for chinese STM32F103C8T6 clones
Hi!
We were unlucky to buy a bunch of BluePills with chinese clone of STM32 called CKS32F103C8T6. As far as I understand, it is a perfect replacement, but openocd fails to upload a firmware in it. If I understand correctly, it is not perfectly compatible - openocd recieves the same CPUTAPID as STM32F3x series MCUs. If I put 
```openocd
set CPUTAPID 0x2ba01477
```
just before include of mcu's config (so it cannot be fixed by using a custom script), it uploads just fine and works.
It requires futher testing.
I want to propose a change to add an option to bluepill bdf's to  support it.

salkinium:
The modm board configurations hard-coded the programmer, which is an issue when using a different one.

This replaces #235 and fixes #234 and fixes #238.

cc @ASMfreaK

### DIFF
--- a/src/modm/board/black_pill/module.lb
+++ b/src/modm/board/black_pill/module.lb
@@ -22,6 +22,36 @@ Cheap and bread-board-friendly board for STM32 F1 series.
 Sold for less than 2 USD on well known Internet shops from China.
 
 http://wiki.stm32duino.com/index.php?title=Black_Pill
+""" + descr_programming
+
+descr_programming = """
+## Programming
+
+Since the board doesn't have a programmer on-board, you need to use your
+own and *specify* which one you're using in a custom `openocd.cfg` file:
+
+```
+# Replace this with your custom programmer
+source [find interface/stlink-v2.cfg]
+
+# If you use the clone CKS32F103C8T6 chip you need to overwrite this ID
+#set CPUTAPID 0x2ba01477
+
+# To select a specific programmer you can specify its serial number
+#hla_serial "\x53\x3f\x6f\x06\x50\x77\x50\x57\x12\x17\x14\x3f"
+# You can discover the serial via `stlink --hla-serial` or `st-info --hla-serial`.
+```
+
+Then include this file in your build options like so:
+
+```xml
+<library>
+  <extends>modm:black-pill</extends>
+  <options>
+    <option name="modm:build:openocd.cfg">openocd.cfg</option>
+  </options>
+</library>
+```
 """
 
 def prepare(module, options):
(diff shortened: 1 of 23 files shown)


salkinium:
@ASMfreaK I think the PR is a fix for both your issues and should allow future customization for these situations. Can you check if this works for you?

tools/build_script_generator/module.lb
```diff
@@ -157,6 +152,10 @@ def post_build(env):
         "gitignore": gitignore,
         "elf_files": elf_files,
     }
+    # prepare custom path
+    openocd_cfg = env.get(":build:openocd.cfg", "")
```
asmfreak, line 156: I think we should warn user or even generate an error if this option is empty.

asmfreak (changes requested):

tools/build_script_generator/module.lb
salkinium: Good idea, I've implemented this as a warning displaying part of the description again:
```
 $ lbuild build
[WARNING] user.modm.board.blue-pill: You need to provide the programmer via a custom OpenOCD config!

## Programming

Since the board doesn't have a programmer on-board, you need to use your
own and *specify* which one you're using in a custom `openocd.cfg` file:

    # Replace this with your custom programmer
    source [find interface/stlink-v2.cfg]

    # If you use the clone CKS32F103C8T6 chip you need to overwrite this ID
    #set CPUTAPID 0x2ba01477

Then include this file in your build options like so:

    <library>
      <extends>modm:blue-pill</extends>
      <options>
        <option name="modm:build:openocd.cfg">openocd.cfg</option>
      </options>
    </library>
```

salkinium:
@ASMfreaK Ok with you to merge this?

src/modm/board/black_pill/module.lb
```diff
@@ -22,6 +22,32 @@ Cheap and bread-board-friendly board for STM32 F1 series.
...
+descr_programming = """
+## Programming
+
+Since the board doesn't have a programmer on-board, you need to use your
+own and *specify* which one you're using in a custom `openocd.cfg` file:
+
+```
+# Replace this with your custom programmer
+source [find interface/stlink-v2.cfg]
+
```
asmfreak, line 36: It would be nice to add the following:
```openocd
# If you want to select specific instance of programmer,
# uncomment the following line:
# hla_serial "..."
# You can get hla_serial by calling `stlink --hla-serial`
```

asmfreak (changes requested):
I like how this is going

src/modm/board/black_pill/module.lb
salkinium: Added this too with example serial

asmfreak (approved):
This is great! Thanks!

PR #241, opened 2019-07-10
Modularize CMSIS-DSP and add full examples

### LINKED ISSUE #239: CMSIS DSP duplicate targets for arm_bitreversal2.o
CMSIS DSP added a file `arm_bitreversal2.c` which conflicts with `arm_bitreversal2.S`.

salkinium:
This modularizes CMSIS-DSP to reduce the amount of source files that need to be compiled for your project and fixes #239.

This also wraps all CMSIS-DSP examples into modm examples to check for regressions in future.

cc @se-bi @chris-durand

### DIFF
--- a/examples/nucleo_f429zi/cmsis_dsp/linear_interp/main.cpp
+++ b/examples/nucleo_f429zi/cmsis_dsp/linear_interp/main.cpp
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2019, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <arm_math.h>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-compare"
+#define main arm_cmsis_dsp_example
+#include "../../../../ext/arm/cmsis/CMSIS/DSP/Examples/ARM/arm_linear_interp_example/math_helper.h"
+#include "../../../../ext/arm/cmsis/CMSIS/DSP/Examples/ARM/arm_linear_interp_example/math_helper.c"
+/* `float arm_linear_interep_table[188495]` is 736kB large and MUST be placed in Flash
+ * memory NOT SRAM otherwise it won't fit on the device. Since we don't want (=cannot)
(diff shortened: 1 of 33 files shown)


examples/nucleo_f429zi/cmsis_dsp/linear_interp/main.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-compare"
+#define main arm_cmsis_dsp_example
+#include "../../../../ext/arm/cmsis/CMSIS/DSP/Examples/ARM/arm_linear_interp_example/math_helper.h"
+#include "../../../../ext/arm/cmsis/CMSIS/DSP/Examples/ARM/arm_linear_interp_example/math_helper.c"
+/* `float arm_linear_interep_table[188495]` is 736kB large and MUST be placed in Flash
+ * memory NOT SRAM otherwise it won't fit on the device. Since we don't want (=cannot)
+ * to change the example, we need to take drastic measures... */
+#define float const float
```
salkinium, line 23: This is epic… ARM thought it would be a fantastic idea to declare `arm_linear_interep_table` not as const, so it gets placed into SRAM, which at 736kB does not fit (surprise, surprise).
I can fix that with this define, however, the `arm_linear_interp_instance_f32` structure requires a non-const `float*` 
pointer, so there is a cast missing, which normally causes a compile error. But with `-fpermissive` I can reduce this to a warning:

```
In file included from main.cpp:25:
../../../../ext/arm/cmsis/CMSIS/DSP/Examples/ARM/arm_linear_interp_example/arm_linear_interp_example_f32.c: In function 'int32_t arm_cmsis_dsp_example()':
../../../../ext/arm/cmsis/CMSIS/DSP/Examples/ARM/arm_linear_interp_example/arm_linear_interp_example_f32.c:151:80: warning: invalid conversion from 'const float*' to 'float32_t*' {aka 'float*'} [-fpermissive]
   arm_linear_interp_instance_f32 S = {188495, -3.141592653589793238, XSPACING, &arm_linear_interep_table[0]};
                                                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is the best I can do…

salkinium:
So to summarize… All these examples run in a `int32_t main()` function and write their success to a `arm_status status` variable and always end with `while(1);` and it's the only use of the `while` keyword in that file…

Screenshot 2019-07-11 at 01 54 44

Hmmmmmmm 🤔🤔🤔🤔  what if we just redefine a few things…

Screenshot 2019-07-11 at 01 55 21

😳 **`#define while return status; void`** 👀

Screenshot 2019-07-11 at 01 51 31

PR #244, opened 2019-07-16
Bugfix: STM32 acknowledge of interrupt clears all pending bits 

### LINKED ISSUE #243: STM32 acknowledge of interrupt clears all pending bits
I've used the STM32 feature, where several external interrupts are handled by one ISR (EXTI9_5). In my application I used GPIO5 to GPIO7. When several interrupt request are pending an I acknowlede just one of them, the other ones are cleared too.

The acknowledge routine looks like this:

`inline static void acknowledgeExternalInterruptFlag() { EXTI->PR |= mask; }`

It reads the content of the PR register, ORs it with the mask and writes the result back to the PR register.  But writing a 1 to the PR register will reset the according bit and therefore all bits are reset.

I've changed the routine to:

`inline static void acknowledgeExternalInterruptFlag() { EXTI->PR = mask; } `

and now it seems to work correctly. 

Maybe there is someone who can confirm this behaviour.

19joho66:
Fixes #243:

When using `inline static void acknowledgeExternalInterruptFlag() { EXTI->PR |= mask; }`,
EXTI->PR is read into a cpu-register which will then be ORed with the mask (so it does basicaly nothing) and then the register value is written back to PR which then clears all set interrupt pending bits (and not just the desired one).

### DIFF
--- a/src/modm/platform/gpio/stm32/pin.hpp.in
+++ b/src/modm/platform/gpio/stm32/pin.hpp.in
@@ -156,7 +156,7 @@ public:
 	{ EXTI->FPR1 |= mask; EXTI->RPR1 |= mask; }
 %% else
 	inline static bool getExternalInterruptFlag() { return (EXTI->PR{{pf}} & mask); }
-	inline static void acknowledgeExternalInterruptFlag() { EXTI->PR{{pf}} |= mask; }
+	inline static void acknowledgeExternalInterruptFlag() { EXTI->PR{{pf}} = mask; }
 %% endif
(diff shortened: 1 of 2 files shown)


asmfreak:
Is this same on STM32F1x?

19joho66:
According to STM32F100xx reference manual:

grafik

So yes, it seems to apply to STMF1x too.

asmfreak:
@19joho66 Thanks!

salkinium (approved):
Thanks for the fix!

PR #250, opened 2019-07-20
[cmake] Align capabilities with SCons generator

### LINKED ISSUE #125: Support separate compile flags per module/file in CMake

### LINKED ISSUE #245: CrashCatcher hard fault handler asm file not linked with cmake
When compiling using CMake, the CrashCatcher_armv7m.S doesn't get assembled and linked into the resulting binary.
When using Scons, everything is working as expected (but it doesn't integrate with CLion which I'm using as my IDE)

salkinium:
I'm trying to port as much of the `:build:scons` features to the CMake generator, to make it less of a second-class build system, since most people use it with their IDEs.

- [x] Generate one CMakeLists.txt per repo
- [x] Add linking with libraries
- [x] Add compile flags per file
- ~Better find application files~ (apparently not how CMake works)
- [x] Parse pkg-config
- [x] Compilation for avr-gcc
- [x] Compilation for hosted
- [x] Compile more edge-case examples with CMake (like CMSIS-DSP)
- [x] Fix use of avr-size (also for SCons)
- [x] Update documentation
- [x] Copy only required SCons tools
- [x] Remove SCons CMakeWrapper tool (it's now worse than using `:build:cmake`)

## Discussion

There are many SCons only features, since they require tools that are only written as SCons tools and are not part of this PR. These tools should be refactored into build-system agnostic Python scripts whenever possible and provided by `:build` and then only wrapped by the specific build systems.

It's difficult for me to understand how to make this truly modular, since Modern CMake seems to advise for completely separate compile flags for all your libraries (?), but modm requires a lot of global compile flags.

The SCons generator also adds all (non-file-specific) compile flags to the global set, so you application gets compiled with the same flags that modm gets compile with. Maybe we can split it up into required global flags (like ARCHFLAGS) that are required for everything, and just modm-specific flags (like warnings). However, it's unclear to me how to specify this difference in the compile flag Collectors.

This fixes #125 and also fixes #245.

cc @siy @ASMfreaK @XDjackieXD @fb39ca4 @dergraaf


salkinium:
Well… this PR escalated a little, but I'm done refactoring now.

salkinium:
I tested these changes successfully with the CLion IDE. It's nice to have code completion for once. Would be nice to also generate programming and debugging support files, but CLion doesn't support cross-compiling stuff (but there seems to be a [plug-in](https://plugins.jetbrains.com/plugin/10115-openocd--stm32cubemx-support-for-arm-embedded-development)).
There's also [PlatformIO IDE](https://platformio.org/platformio-ide), which is significantly more tailored towards embedded. 

I'm not sure how users use modm to be honest… I don't want to invest time maintaining some integration that isn't used a lot.

asmfreak:
We are using modm with compilation db and some makefiles to make everything work with CLion.

XDjackieXD:
Thank you very much! I will give this a throughout test as soon as I get back home (probably in about 4 days).
My primary usecase is with CLion and some automated Ci building

salkinium:
Supporting Compilation DB natively is important to me too, since a lot of C++ IDEs start to support it now (like Qt Creator) and all the required information is already there anyways.

salkinium:
Just a heads up, I something isn't right with the CMake generator I wrote, it works for the Linux CI, but not on macOS… #263.

PR #255, opened 2019-07-24
[docs] Use only Python 3 with macOS

strongly-typed:
We can discuss the `--user` option which does not work for me.

I tried this on a very new installation and cleaned up as much brew before.

### DIFF
--- a/docs/src/guide/installation.md
+++ b/docs/src/guide/installation.md
@@ -34,8 +34,7 @@ Please help us [keep these instructions up-to-date][contribute]!
 We will use [Homebrew](http://brew.sh/) to install the minimal build system:
 
 	brew install python3 scons git doxygen
-	pip3 install --user modm
-	pip install --user pathlib pyelftools
+	pip3 install modm
 
 Install the [AVR toolchain from source](https://github.com/osx-cross/homebrew-avr):
 
@@ -49,6 +48,12 @@ well:
 	brew install arm-gcc-bin
 	brew install openocd --HEAD
 
+scons now works with Python 3. Unfortunately, macOS still defaults to Python 2.
+For a less intrusive way to run all scons scripts with Python 3 add this to your
+~/.bashrc or ~/.zshrc
+
+	alias scons="/usr/bin/env python3 $(which scons)"
+
 To compile modm for macOS you need to install these libraries:
 
 	brew install boost gtkmm sdl zmqpp gcc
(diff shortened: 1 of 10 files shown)


strongly-typed:
I just found that `log_itm.py` and `program_openocd.py` still have `python2` in the header.

~~I am running into some problems with `env.Listify` from f69245a77e48459bccec0baf50808f3bb2765955 right now.~~ Argh, submodule.

docs/src/guide/installation.md
```diff
@@ -49,6 +48,12 @@ well:
 	brew install arm-gcc-bin
 	brew install openocd --HEAD
 
+scons now works with Python 3. Unfortunately, macOS still defaults to Python 2.
+The less intrusive way to run all scons scripts with Python 3 add this to your
+~/.bashrc or ~/.zshrc
+
+	alias scons="/usr/bin/env python3 $(which scons)"
```
salkinium, line 55: This is giving me an issue with `tools/scripts/examples_compile.py`:
```
Building: game_of_life with SCons
scons: Building targets ...
scons: building terminated because of errors.scons: *** [modm/src/modm/architecture/detect.hpp] ValueError : unsupported pickle protocol: 4
```
Can you confirm? Maybe this is just my local issue.

salkinium: Running `scons` via zsh alias works fine.

salkinium:
FYI: [Python2.7 is getting deprecated in macOS 10.15 and removed in macOS 10.16](https://www.macobserver.com/analysis/macos-catalina-deprecates-unix-scripting-languages/).

docs/src/guide/installation.md
strongly-typed: I cannot confirm this error. I can compile the game of life from the directory manually.

But the script does not inherit the alias, so it seems that scons again uses Python2.

strongly-typed: It might need
```python
commands.append("python3 `which scons` build --cache-show --random")
```

salkinium: Yep, that fixes it, great!

salkinium (dismissed):
Good riddance.

strongly-typed:
Sorry for pushing a second after your approval. Did not see it on time. We can discuss what goes into this PR, and what not.

salkinium:
I'd say this is only for Python3 and the ROS stuff goes into a new one.

strongly-typed:
Shall it be "/usr/bin/env python" or "/usr/bin/env python3"?

salkinium:
> Shall it be "/usr/bin/env python" or "/usr/bin/env python3"?

I think it should be python3, but SCons ignores this completely (it has to) and uses the Python version it was started with, so it can also probably just be removed entirely. It's not like we run the individual SCons tools manually from the command line.

strongly-typed:
I'll go for the `python3`

docs/src/guide/installation.md
```diff
@@ -61,7 +66,12 @@ For Ubuntu 18.04LTS, these commands install the basic build system:
 	sudo apt-get install python3 python3-pip scons git
 	sudo apt-get --no-install-recommends install doxygen
 	pip3 install --user modm
-	pip install --user pathlib pyelftools
+
+scons now works with Python 3. Unfortunately, Ubuntu still defaults to Python 2.
+The less intrusive way to run all scons scripts with Python 3 add this to your
```
salkinium: Grammar is wonky. For a less...

salkinium (approved):
Thanks!

PR #260, opened 2019-07-26
Various improvements, deprecations and updates

salkinium:
Adds several changes:

- Generating `.gitignore` files for any repo.
- Updates ignored AVR list.
- Removes boost dependency of hosted unittests by moving `modm::Terminal` to `modm/driver/io/`.
- Removes annoying example count.
- Removes deprecated TIPC XPCC backend.
- Removes SDL display and dependency on gtk and cairomm.


tools/scripts/examples_compile.py
```diff
@@ -35,6 +36,9 @@ def generate(project):
 	path = project.parent
 	output = ["=" * 90, "Generating: {}".format(path)]
 	options = " ".join("-D{}={}".format(k, v) for k,v in global_options.items())
+	# Compile Linux examples under macOS with hosted-darwin target
+	if "Darwin" in platform.platform() and "hosted-linux" in project.read_text():
+		options += " -D:target=hosted-darwin"
```
salkinium: @strongly-typed with these changes you should just be able to compile the linux and generic examples without new macOS specific examples.

salkinium: I've added the TravisCI configuration to this PR, lets see how that works.

salkinium:
This… escalated a little.

tools/scripts/examples_compile.py
salkinium: I'm having issues linking with `zmqpp` on both my local system and the CI for `linux/xpcc` and `zmq/*` examples. Can you reproduce this?

salkinium: Moved to a separate PR, since I cannot get all examples to compile.

PR #262, opened 2019-07-28
[docs] Improve content and generation

salkinium:
Significant changes:

- New generic(!) module docs generator for all(!) targets
- Author list update
- Module docs improvements
- Readme and Homepage docs improvements
- Added full MPLv2 License text


.circleci/config.yml
```diff
@@ -297,9 +297,15 @@ jobs:
...
-            (cd docs/src/reference && ln -s ../../../examples/stm32f469_discovery/blink/modm/docs/module)
+            python3 tools/scripts/generate_module_docs.py
             (cd docs && mkdocs build)
+      # FIXME: Allow localhost:port to work with Docker
+      # - run:
+      #     name: Check Homepage
+      #     command: |
+      #       apt install -y nodejs npm
+      #       npm install -g broken-link-checker-local
+      #       blcl modm/docs/modm.io/docs -ro --host-requests 100 --filter-level 1
```
salkinium, line 308: This would be super useful, but `blcl` opens up a random port on `localhost` and the Docker container doesn't allow that…

asmfreak: I might be asking stupid  things, but can't `blcl` open a unix socket and/or use something like [appswitch](http://appswitch.io/) to make networking easier?
appswitch, if understand it correctly, rewrites every `bind` and `connect` to tcp. If connection is local, appswitch even replaces the tcp socket with a unix socket, making it easier to configure and perform quicker (due to less networking overhead).

salkinium: My issue was that I don't know enough about Docker/Containers/CircleCI to fix this and a quick googling of the issue didn't yield an *simple* solutions.

The check is also only *somewhat* useful, since a few links are only valid with sign-in (the CircleCI links) and some links to servers are overwhelmed ([Blue Pill](http://wiki.stm32duino.com/forums.html?title=Blue_Pill)). So this could only be an optional step anyways (I mean, gosh the internet is fickle…)

asmfreak:
I'm curious - is there any way to run the whole circle-ci check locally? (Should I make this question into an issue?)

salkinium:
No, but you can launch the Docker container locally, and the execute the commands from the CI script.

salkinium:
Out of motivation to polish this more, as it's already significantly better now than before. I'll write more docs when I feel like it…

PR #267, opened 2019-08-07
[lbuild] Warn if submodules are not up-to-date

salkinium:
Adds a more comprehensive check for distiguishing between missing submodules (error) and out-of-date submodules (warning).

cc @se-bi

### DIFF
--- a/repo.lb
+++ b/repo.lb
@@ -22,18 +22,7 @@ from distutils.version import StrictVersion
 from git import Repo
 from os.path import normpath
 
-sys.path.append(repopath("ext/modm-devices/tools/device"))
-try:
-    import modm_devices.parser
-except Exception as e:
-    print(e, "\n")
-    print("You might be missing the git submodules in `ext/`.\n"
-          "Try:\n"
-          "    cd modm\n"
-          "    git submodule update --init\n"
-          "then build again.")
-    exit(1)
-
+# Check for miminum required lbuild version
 import lbuild
 min_lbuild_version = "1.12.1"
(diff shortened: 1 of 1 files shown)


se-bi:
Does it through an error, if it is out of sync? 

Does it distinguish between old state and something to be tested, if that occurs often with the submodules?

salkinium:
> Does it through an error, if it is out of sync?

No, since that would be quite annoying for me.

> Does it distinguish between old state and something to be tested, if that occurs often with the submodules?

No, it only checks if the submodule hexsha is the current checkout hexsha.

salkinium:
I could probably add a way to fail when calling lbuild verbosely with `lbuild -v discover`, would that help? (this would be quite fragile, how often do you call `lbuild -v`?)

se-bi:
Mhm, sorry. I had a comment  on a different Issue in mind.

Warning is fine, user must just find it. :stuck_out_tongue_closed_eyes:

se-bi:
> I could probably add a way to fail when calling lbuild verbosely with `lbuild -v discover`, would that help? (this would be quite fragile, how often do you call `lbuild -v`?)

The `lbuild discover` should not fail, but big red warning at the end would be an option :wink:

salkinium:
> The lbuild discover should not fail, but big red warning at the end would be an option 😉

Hm, I can only put it at the beginning for now (it's using barebone printf), since there isn't a lbuild API for checking the repo.lb validity. Maybe this is a good idea to add to the `init` function…

PR #274, opened 2019-09-03
Fix to isConversionFinished()

cajt:
It previously checked the EndOfSampling flag instead of EndOfConversion.
This typically causes sampled values from the ADC to be invalid or from
the previous conversion due to reading the value too early.


salkinium (approved):
Thanks!

Issue #44, opened 2018-07-29
Tags and Releases are missing.


strongly-typed:
[Some](https://github.com/modm-io/modm/commit/c442b5ce1fc4630a37f2b3bf7e126ead8091f816) commits are marked as releases, although no Github [Releases](https://github.com/modm-io/modm/releases) and not [tags](https://github.com/modm-io/modm/tags) are defined.

salkinium:
Those commits come from merging the xpcc history into modm. 

I only planning to do modm releases and changelog with 2018q4, when xpcc with be archived. It's unclear to me how to "merge" the changelogs/releases of xpcc into modm, so I just procrastinate this problem into the future. 😜

PR #281, opened 2019-09-09
Doxypress API documentation on https://docs.modm.io/

rleh:
Doxygen documentation is available here:
# [docs.modm.io](https://docs.modm.io/)

- [x] Script to batch-generate Doxygen documentation for all relevant targets. (Only the largest package and the largest memory for each series)
- [x] Automated updating: Daily.
- [x] Update target device list.
- [x] Hosting on https://doc.modm.io/; ~server provided by @roboterclubaachen, thanks!~ hosting on my server.
- [x] Optional/Later:
  - [x] Beautify frontpage with some CSS :wink:
  - [x] ~Fix Doxygen settings: SVG images, relative paths~
  - [x] ~Use of https://mcss.mosra.cz/documentation/doxygen/~ Does not work out of the box :unamused: 
- [x] Replace Doxygen with Doxypress
- [x] Automated building in a CI job (and deploy to server)

### DIFF
--- a/docs/src/reference/documentation.md
+++ b/docs/src/reference/documentation.md
@@ -54,31 +54,36 @@ available for {target}*". Additionally the view does not show peripheral
 instance submodule to reduce visual noise.
 
 
-## Doxygen
+## Doxypress / Doxygen
 
-The generated C/C++ API is documented using Doxygen and therefore only available
-*after* code generation.
+The generated C/C++ API is documented using Doxypress or Doxygen and therefore only
+available *after* code generation.
 
 To view this documentation, include the `modm:docs` module in your `project.xml`
 configuration and run `lbuild build`. A `modm/docs` folder is created containing
-the `doxyfile.cfg` for modm as well as all the `@defgroup`s mirroring the
-module structure.
+the `doxypress.json` and `doxyfile.cfg` for modm as well as all the `@defgroup`s
+mirroring the module structure.
 
 You must then call Doxygen manually, which can take several minutes, and
 compiles the target-specific documentation into `modm/docs/html`:
 
 ```sh
 lbuild build -m "modm:docs"
+
+# With Doxypress
+(cd modm/docs && doxypress doxypress.json)
+# Or Doxygen
 (cd modm/docs && doxygen doxyfile.cfg)
+
 # then open: modm/docs/html/index.html
 ```
 
-The generated Doxygen documentation contains the original module documentation
+The generated Doxypress/Doxygen documentation contains the original module documentation
 *including the option choices you made*. This makes it easier to map the
 lbuild options to changes in the generated source code.
 
 
-!!! bug "No online API docs available (yet)"
-	Due to the complexity of creating a superimposed API documentation for
-	hundreds of different targets, we currently do not have an online API browser
-	available. Please generate the Doxygen API documentation locally for now.
+### Online API documentation
+
+API documentation is available at https://docs.modm.io/ for a large and representative
+number of targets with all modules enabled.
--- a/tools/scripts/docs_modm_io_generator.py
+++ b/tools/scripts/docs_modm_io_generator.py
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018, Niklas Hauser
+# Copyright (c) 2018-2019, Raphael Lehmann
+#
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+import shutil
+import lbuild
+import zipfile
+import tempfile
+import argparse
+import datetime
+import multiprocessing
+import os, sys
+from pathlib import Path
+from jinja2 import Environment, FileSystemLoader
+from collections import defaultdict
+
+def repopath(path):
+    return Path(__file__).absolute().parents[2] / path
+def relpath(path):
+    return os.path.relpath(path, str(repopath(".")))
+
+sys.path.append(str(repopath("ext/modm-devices/tools/device")))
+from modm_devices.device_identifier import *
+
+
+def get_targets():
+    builder = lbuild.api.Builder()
+    builder._load_repositories(repopath("repo.lb"))
+    option = builder.parser.find_option(":target")
+    ignored = list(filter(lambda d: "#" not in d, repopath("test/all/ignored.txt").read_text().strip().splitlines()))
+    raw_targets = sorted(d for d in option.values if not any(d.startswith(i) for i in ignored))
+    minimal_targets = defaultdict(list)
+
+    for target in raw_targets:
+        option.value = target
+        target = option.value._identifier
+        short_id = target.copy()
+
+        if target.platform == "avr":
+            short_id.naming_schema = short_id.naming_schema \
+                .replace("{type}-{speed}{package}", "") \
+                .replace("-{speed}{package}", "")
+
+        elif target.platform == "stm32":
+            short_id.naming_schema = "{platform}{family}{name}"
+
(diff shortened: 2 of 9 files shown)


salkinium:
Perhaps "optional optional" is use of https://mcss.mosra.cz/documentation/doxygen/.

salkinium:
FYI: We once came up with this URL scheme for docs.modm.io, but this depends on me doing releases again (#44):

```
modm releases: /{release}
/develop, /2019q1

lbuild modules: /{release}/module/
Doxygen API: /{release}/api/
Target info: /{release}/target/

group target info: /{release}/target/{target-family}/
specific target info: /{release}/target/{target}/
specific or grouped lbuild module: /{release}/module/{module}/
specific lbuild module: /{release}/module/{module}/{target-family}/
specific doxygen api: /{release}/api/{target}/

target specialization order:
  https://docs.modm.io/{release}/module/
   https://docs.modm.io/{release}/module/{module}/
    https://docs.modm.io/{release}/module/{module}/{target-family}/
     -> https://docs.modm.io/{release}/targets/{target-family}/
      -> https://dosc.modm.io/{release}/api/{target}/
```

salkinium:
Can you perhaps add a small test for this script to the CI, so that we can catch issues when refactoring?

rleh:
What kind of test are you thinking of?
We could run the script in test mode, which generates documentation for `["hosted-linux", "at90can128", "stm32f103rbt", "stm32f429ziy"]`:
```
./docs_modm_io_generator.py -t -c
```

salkinium:
Yes, that sounds great! Add it to the build-docs job and let's see how much more time it needs.

rleh:
Any idea how to keep the target device list up-to-date?
I would like to generate a list of the useful targets by choosing the largest (pin-count and memory) STM32 device for every sub-family and then removing broken targets from a blacklist.

Is there a way to programmatically access the device list in the error message of `lbuild -r modm/repo.lb discover-options`?

salkinium:
Yes, you can use the lbuild Python API directly, see for example: https://github.com/modm-io/modm/blob/develop/tools/scripts/generate_module_docs.py#L34

salkinium:
Here's the public facing API: https://github.com/modm-io/lbuild/blob/develop/lbuild/api.py
I'll rewrite this script to programmatically choose the modules \ boards too, but next WE.

rleh:
*Test run of docs.modm.io-generator-script* in CircleCi has some issues:
```
ERROR: In 'modm:docs':

  File "/usr/lib/python3.6/textwrap.py", line 248, in _wrap_chunks
    raise ValueError("invalid width %r (must be > 0)" % self.width)
ValueError: invalid width -16 (must be > 0)

warning: tag INPUT: input source `dox/' does not exist
```

tools/scripts/docs_modm_io_generator.py
```diff
@@ -0,0 +1,403 @@
...
+<ul>
+{% for d in devices %}
+ <li id="{{ d }}"><a href="/develop/api/{{ d }}/">{{ d }}</a></li>
+{% endfor %}
+</ul>
+</noscript>
+</main>
+</body>
+</html>
+"""
```
salkinium: Please put this into a separate .html.in file, this will likely grow even longer in future.

rleh: I want to keep the script self-contained because it makes document generation on the server side more flexible.
The server can simply do the following:
```
wget -O - https://raw.githubusercontent.com/modm-io/modm/develop/tools/scripts/docs_modm_io_generator.py | python3
```

rleh:
grafik

tools/scripts/docs_modm_io_generator.py
salkinium: But you need to clone the repository anyways, I don't see how this is better than `git clone modm && python3 modm/tools/scripts/foo.py`.

salkinium:
Well hot dayum this is cool!

asmfreak:
There is [Copperspice's doxypress](https://github.com/copperspice/doxypress). It is, as I can see from documentation, a better alternative to Doxygen for modern C++ code, which is modm.

tools/scripts/docs_modm_io_generator.py
rleh: I don't like the idea of automatically running a script downloaded from the Internet and would not do that on my server or on the roboterclubaachens server.

On an isolated VM (like CI) this would be a different matter, but it is problematic to save/transmit the output there and to execute the re-generation regularly.

rleh:
>  I'll rewrite this script to programmatically choose the modules \ boards too, but next WE.

Any progress?

tools/scripts/docs_modm_io_generator.py
salkinium: And yet you're executing the modm lbuild modules which are just plain Python scripts downloaded from the internet…

salkinium:
> Any progress?

I'm currently busy paying the rent and not super motivated to context switch to work on modm, sorry!

tools/scripts/docs_modm_io_generator.py
rleh: Good argument.
I will change the script to use the repo instead of cloning it to temporary directory and put the template into a separate file.

I will add some proper sand-boxing (chroot, cgroups, ...) on the server and add a separate helper script.

salkinium:
Btw, here is some WIP to generate the target list via the lbuild Python API.
https://github.com/salkinium/modm/commit/b19f7d40b4449ef6e51f35114752c9d1472c7306

rleh:
Two issues, first:
```
ERROR: In 'modm:docs':

  File "/usr/lib/python3.6/textwrap.py", line 248, in _wrap_chunks
    raise ValueError("invalid width %r (must be > 0)" % self.width)
ValueError: invalid width -16 (must be > 0)
```
@salkinium Do you have any idea what that should mean?

---

Second issue:
Doxypress hangs when parsing the generated code. I can reproduce the same behavior on my local Fedora machine. I'm investigating...

rleh:
Steps to reproduce:
```bash
lbuild -r modm/repo.lb -D "modm:target=hosted-linux" -p /tmp/modm__hosted-linux/ build  -m"modm:**"

cd /tmp/modm__hosted-linux/modm/docs

# run doxypress
doxypress
# optional with debug flags:
# doxypress -d classes -d commentcnv -d commentscan -d extcmd -d filteroutput -d findmembers -d functions -d lex -d markdown -d preprocessor -d printtree -d validate -d variables

^C
# This will run forever

(Re)move the following files:
rm ../src/modm/processing/resumable/resumable.hpp
rm ../src/modm/processing/resumable/nested_resumable.hpp
rm ../src/modm/architecture/interface/i2c.hpp
rm ../src/modm/platform/gpio/base.hpp
rm ../src/modm/driver/radio/nrf24/nrf24_data.hpp
rm ../src/modm/driver/display/hd44780_base.hpp

# run doxypress
doxypress

# Success. WTF?
```

With debug flags enabled the last output is: `Finished lexical analyzer: ../src/generated/default_args.cpp` (>10 times)

rleh:
Bug in doxypress found: https://github.com/copperspice/doxypress/issues/34

docs/src/reference/documentation.md
```diff
@@ -54,31 +54,36 @@ available for {target}*". Additionally the view does not show peripheral
...
+The generated Doxypress/Doxygen documentation contains the original module documentation
 *including the option choices you made*. This makes it easier to map the
 lbuild options to changes in the generated source code.
 
 
-!!! bug "No online API docs available (yet)"
-	Due to the complexity of creating a superimposed API documentation for
-	hundreds of different targets, we currently do not have an online API browser
-	available. Please generate the Doxygen API documentation locally for now.
+### [Online API documentation](https://docs.modm.io/)
```
salkinium: Links in a Header??? Let's not.

salkinium (changes requested):
Excellent work all around!

docs/src/reference/documentation.md
rleh: Ok

salkinium (dismissed):

PR #288, opened 2019-09-10
[board] fix comment on L3GD20 MOSI pin

WasabiFan:
Super minor; I assume this is what the comment intended to say.

### DIFF
--- a/src/modm/board/disco_f072rb/board.hpp
+++ b/src/modm/board/disco_f072rb/board.hpp
@@ -94,7 +94,7 @@ using Int2 = GpioInputC2;	// MEMS_INT2 [L3GD20_DRDY/INT2]: GPXTI1
 
 using Cs   = GpioOutputC0;	// CS_I2C/SPI [L3GD20_CS_I2C/SPI]
 using Sck  = GpioOutputB13;	// SPI2_SCK [L3GD20_SCL/SPC]
-using Mosi = GpioOutputB15;	// SPI2_MISO [L3GD20_SDA/SDI/SDO]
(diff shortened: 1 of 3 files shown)


salkinium (changes requested):
Thanks, but please squash into one commit. (Trying to keep the commit noise low)

WasabiFan:
You can squash it when you merge from the GitHub GUI :wink: Just click "squash and merge" rather than "merge". That's why I only gave the first one a message.

salkinium:
> You can squash it when you merge from the GitHub GUI

GitHub insists on using merge commits, so you can squash and rebase all you want, they still put a merge commit in there. I don't know what they're smoking, but it must be good.

WasabiFan:
For reference, [this repo](https://github.com/ev3dev/ev3dev-lang-python/commits/ev3dev-stretch) is an example with a bunch of "squash and merge"es; aside from some older merge commits (either intentionally, by accident, or before GH added the feature because not all contributors were capable of squashing) the more recent pull requests were all merged by squashing from the web UI.

[This commit](https://github.com/ev3dev/ev3dev-lang-python/commit/4394507f59f448115a012f04ee212e68fe27df26), for example, was done that way. It has me as the committer and the original PR author as the author, and it shows there is only one parent, i.e. it's a regular commit.

salkinium:
I'm surpressing the urge to yell "get off my lawn" while sitting in a rocking chair 😉. I tested this sometime ago and wasn't happy, but I don't remember why. Maintaining this stuff is hard, humor me.

WasabiFan:
Point taken, you should do what's best for you :grin:. I'll manually squash before opening PRs.

PR #289, opened 2019-09-10
[scons] fix Windows path handling in "program"

WasabiFan:
The backslashes in a Windows-style path weren't interpreted correctly when passing the source path to openocd. This caused the following beautiful error:

image

The strange line rewriting in the error message was enough to convince me to go learn Tcl and start escaping stuff.

### DIFF
--- a/tools/build_script_generator/scons/resources/build_target.py.in
+++ b/tools/build_script_generator/scons/resources/build_target.py.in
@@ -35,10 +35,10 @@ def build_target(env, sources):
 	env.Alias("artifact", env.CacheArtifact(program))
 
 		%% if upload_with_artifact
-	env.Alias("program", [env.OpenOcd(program, commands=["modm_program $SOURCE"]), "artifact"])
+	env.Alias("program", [env.OpenOcd(program, commands=["modm_program {$SOURCE}"]), "artifact"])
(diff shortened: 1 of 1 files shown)


salkinium (dismissed):
Nice!

WasabiFan:
I know it probably wouldn't help with this particular bug, but are you aware there are free Windows CI providers for open-source projects (Azure DevOps, AppVeyor)? It probably wouldn't be too hard to run the same tests you run on Linux (or some subset of them)  on Windows.

salkinium:
> are you aware there are free Windows CI providers

Yes, it's on my TODO list. It's just very annoying to set up, particularly since I haven't used Windows in over a decade (and I'm definitely not dedicating 40GB of SSD space to a VM). Windows support is just best-effort from my part.

WasabiFan:
Understood; I'll try to fix these things as I find them.

For reference, "scons gdb" appears to be broken too, but running the openocd server manually worked fine so I am not immediately worried. There were some comments in the source file about it not working right anyway so maybe there isn't much point.

salkinium:
Open an issue, otherwise I can't track it. The gdb scons tool opens openoce in the background, that's probably broken on windows.

PR #291, opened 2019-09-10
Fixed condition in I2C ISR for continued writes

### LINKED ISSUE #290: I2C EEPROM driver get stalled when writing
Greetings,
I'm using the `i2cEeprom` module to read and write data to a FM24CL64 FRAM chip. The firmware is running on a STM32F042K6 chip.

Reading data from the FRAM works without any problems but writing stalls the whole chip (as I call the write method with `RF_CALL_BLOCKING`).

When sniffing on the I2C bus, the two address bytes got transmitted and afterwards the chips holds the clock down on ground and seems to be in an endless loop waiting for the transaction to finish.

I pinned  down the problem to the `modm::i2cEeprom::DataTransmissionAdapter::writing()` function in the i2c_eeprom.cpp file. The method only got called once for the address bytes but then never called again. Weirdly, when i change the `next` parameter to `OperationAfterWrite::Restart`, it will be called again, and the data will be put on the bus. 

Also if I don't use the driver but instead building custom transactions to communicate with the chip, the write operation works that way. So there is no hardware issue.

Does anyone have a clue what could cause the problem? I was digging into the code of the I2C Master but couldn't find anything obvious that could cause the problem.

Thanks in advance, Ben

nesos:
Fixed the condition to wait for continued writes.
When a write operation followed by another write operation occurs, the ISR will falsely wait for a STOP condition, which will not be triggered as more data from the following write operation has to be transmitted.
This fix will wait for a TC interrupt if the next operation is a restart **or a write operation** and wait for a STOP interrupt otherwise

Fixes #290.

### DIFF
--- a/src/modm/platform/i2c/stm32-extended/i2c_master.cpp.in
+++ b/src/modm/platform/i2c/stm32-extended/i2c_master.cpp.in
@@ -422,7 +422,8 @@ MODM_ISR(I2C{{ id }}_EV)
 
 		if (writing.length == 0)
 		{
-			if (nextOperation == modm::I2c::Operation::Restart)
+			if ((nextOperation == modm::I2c::Operation::Restart) or
+				(nextOperation == modm::I2c::Operation::Write))
 			{
 				DEBUG_STREAM("Wait for TC IRQ");
 				I2C{{ id }}->CR1 |= I2C_CR1_TCIE;


salkinium (dismissed):
Fantastic!

salkinium:
cc @strongly-typed

salkinium:
CI is complaining about one trailing whitespace 😉

salkinium:
(please squash)

PR #304, opened 2019-10-15
Fixes clocks of USART1 and I2C1 on Nucleo F303K8 board.

### LINKED ISSUE #298: STM32F303x8 USART1 clock defaults to PCLK1
The clock for the USART1 on the Nucleo-F303K8 board defaults to PCLK1, which is divided by 2. The line

` static constexpr uint32_t Usart1 = Apb2;`

must be changed to

` static constexpr uint32_t Usart1 = Apb1;`

The reference manual tells it's connected to APB2, but the clock configuration register CFGR3 sets it to PCLK1 per default. See pages 162ff.

Screen Shot 2019-10-02 at 8 57 05 AM

mikewolfram:
According to data sheet USART1 clock defaults to PCLK1 and I2C1 clock to HSI.

Fixes #298.

### DIFF
--- a/src/modm/board/nucleo_f303k8/board.hpp
+++ b/src/modm/board/nucleo_f303k8/board.hpp
@@ -41,11 +41,13 @@ struct SystemClock {
 
 	static constexpr uint32_t Spi1   = Apb2;
 
-	static constexpr uint32_t Usart1 = Apb2;
+	// USART1 is on APB2 bus, but default clock is PCLK1
+	static constexpr uint32_t Usart1 = Apb1;
 	static constexpr uint32_t Usart2 = Apb1;
 	static constexpr uint32_t Usart3 = Apb1;
 
-	static constexpr uint32_t I2c1   = Apb1;
+	// I2C1 clock source is HSI per default
+	static constexpr uint32_t I2c1   = 8_MHz;
 
 	static constexpr uint32_t Apb1Timer = Apb1 * 2;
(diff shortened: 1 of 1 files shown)


salkinium:
FYI: clock data rendered from CubeMX: https://salkinium.com/stm32/clock/
Is just I2C1 set to HSI8, or also other peripherals? The graph suggests there is at least a connection: https://salkinium.com/stm32/clock/F303.svg

mikewolfram:
Your graph seems to be generic for the F303 family. This fix is specific to the F303x8, which has less peripherals. This is the clock tree from the data sheet, only I2C1 is on HSI.

image

salkinium:
Ah, very interesting. Yes, the data is generic, which is why this whole clock stuff isn't generated already. I didn't find any better data though, hmpf.

salkinium (approved):
Thanks!

mikewolfram:
Hang on for a second. My project uses I2C1 as well to drive a small display and I had no issues with the original clock speed. Let me double check in the office tomorrow.

salkinium:
Then the I2C bus would use a higher prescaler than correct, but the display probably wouldn't complain about a 8x slower bus.

mikewolfram:
You are right, it would run at 50 instead of the desired 400 kHz.

salkinium:
Did you have time to check your setup?

mikewolfram:
Display works fine with 8 MHz clock.

PR #305, opened 2019-10-17
Added NACKIE flag to read and write operations

nesos:
The read and write operations are missing the NACK interrupt enable flag. This causes devices to be trapped in an endless loop when the addressed device is not responding.
In this case no interrupt will be raised, the transaction will not be detached and stays busy forever.

I enabling the NACK interrupt when starting the transaction is fixing this issue


salkinium:
Is there also a NACKIE that needs to be enabled for the non-extended STM32 driver?

nesos:
> Is there also a NACKIE that needs to be enabled for the non-extended STM32 driver?

I haven't checked that. I'm currently working only with the STM32F042K6. I'll check this later tonight on on a STM32F411 board

nesos:
The problem on getting the system stuck when trying to read from I2C nodes that don't respond does not appear on the non-extended master driver. So no need to change on that part.

salkinium:
Great!

Can you rebase your commit so I can fast forward develop?

nesos:
> Great!
> 
> Can you rebase your commit so I can fast forward develop?

Done!

salkinium:
Thanks!

PR #309, opened 2019-10-31
Added timestamps to CAN messages when receiving

nesos:
To implement a time-synchronization mechanism for a CAN network I added a timestamp feature to the CAN driver.
This feature will append the current timestamp to all incoming messages, such that one can read the time of arrival when processing the message.

I'm not sure if this featured is needed or desired in the modm mainline code, but I thought I'd offer it.

### DIFF
--- a/src/modm/architecture/interface/can_message.hpp
+++ b/src/modm/architecture/interface/can_message.hpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2014, 2016, Niklas Hauser
  * Copyright (c) 2015-2016, Sascha Schade
+ * Copyright (c) 2019, Benjamin Weps
  *
  * This file is part of the modm project.
  *
@@ -16,6 +17,7 @@
 #include <stdint.h>
 #include <modm/io/iostream.hpp>
 #include <modm/architecture/utils.hpp>
+#include <modm/processing/timer/timestamp.hpp>
 
 namespace modm
 {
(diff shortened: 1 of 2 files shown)


salkinium:
Is the timestamp a platform-independent CanMessage feature? Otherwise I'd add it like the filter_id:
https://github.com/modm-io/modm/pull/226#issuecomment-499602327

This'll also line up with the FDCAN API that @rleh is [preparing in the background](https://github.com/modm-io/modm/commit/8fbebbc60c23b0e4a19756e47473a99a219ab103#diff-aec730b7667d09c0eb46d73487ad5fb8R203).

nesos:
> Is the timestamp a platform-independent CanMessage feature?

It is platform independent. It doesn't use any features of the CAN driver itself. It only needs the clock module (and the interriupt of the driver)

strongly-typed:
> > Is the timestamp a platform-independent CanMessage feature?
> 
> It is platform independent. It doesn't use any features of the CAN driver itself.

Sadly, the Time Triggered CAN in STM32 is only marketing.

From the [STM32F42xx and STM32F43xx Errata sheet:](https://www.st.com/content/ccc/resource/technical/document/errata_sheet/38/e6/37/64/08/38/45/67/DM00068628.pdf/files/DM00068628.pdf/jcr:content/translations/en.DM00068628.pdf)
```
2.13.1 BxCAN time triggered communication mode not supported
```
So nothing lost here.

Great to see the feature implemented in this way!

But, is it really required that **all** CAN messages get the timestamp? I would opt in for an option of the module to enable time stamps, with disabled as default and add a ``CanStamped`` data type (with reference to ROS). 

IMHO, same is true for the ``filter_id`` which is in most my cases irrelevant and increases RAM usage for heavy CAN traffic significantly here.

PR #314, opened 2019-11-14
GraphicDisplay virtual methods

mikewolfram:
Make fillRectangle and fillCircle of GraphicDisplay virtual to allow overriding them in derived classes for improved performance.

### DIFF
--- a/src/modm/ui/display/graphic_display.hpp
+++ b/src/modm/ui/display/graphic_display.hpp
@@ -314,7 +314,7 @@ namespace modm
 		/**
 		 * Fill a rectangle.
 		 */
-		void
+		virtual void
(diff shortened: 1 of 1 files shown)


salkinium:
Ok, we can do that, however in general I would recommend to use a proper embedded GUI library, since ours is super old and not really supported anymore.
I've been looking at https://littlevgl.com/ to integrate into modm, but my TODO list is long and time is short.

salkinium (approved):

mikewolfram:
Sorry, didn't know that yours is old. Actually wanted to contribute an ILI9341 SPI driver once finished.
I've been looking into LittlevGL. It takes too many resources and might be a bit overkill for what we need/want to do.

salkinium:
You may of course add your driver and use the stack, it does work, however no new features/bugfixes will be added *by me*.

PR #317, opened 2019-11-20
Add filter to extract the ADC channel number only

### LINKED ISSUE #316: STM32 ADC channels >= 10 broken
Commit 086581565e088c4fda60e8d13a8a8b1fc9f0912b (adding support for L1) breaks ADC channels >= 10. A change in *src/modm/platform/gpio/stm32/pin.hpp.in* sets the ADC channel to always 1 when the Signal is `In10...In15`. Before supplying a fix I wanted to ask if this change has a purpose:

```diff
- Gpio{{ port ~ pin }}::AdcChannel<Peripheral::{{ signal.driver }}> = {{ signal.name.replace("In", "") }};
+ Gpio{{ port ~ pin }}::AdcChannel<Peripheral::{{ signal.driver }}> = {{ signal.name.replace("In", "")[0] }};
```

mikewolfram:
I'm not a Python expert but hope the fix is ok.

Fixes #316.

### DIFF
--- a/src/modm/platform/gpio/stm32/module.lb
+++ b/src/modm/platform/gpio/stm32/module.lb
@@ -260,7 +260,7 @@ def build(env):
         po, pi = gpio["port"], gpio["pin"]
         properties.update(bprops[po + pi])
         header_name = "gpio_{}{}.hpp".format(po.upper(), pi)
-        env.template(gpio_source, header_name)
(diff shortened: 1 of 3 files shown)


salkinium (dismissed):

salkinium:
Actually, can you also use the filter here for consistency? https://github.com/modm-io/modm/blob/develop/src/modm/platform/gpio/stm32/pin_f1.hpp.in#L225

PR #320, opened 2020-01-08
Correct enum SampleTime to match values in datasheets of G4 and L4 MCUs.

mikewolfram:


salkinium (commented):
Thanks! I wish we had some better data for these details, but until then, manual fixes it is.

Can you squash your commits into one commit?

mikewolfram:
Done.

mikewolfram:
Looking into the ADC code of STM32 I found more issues. Like the F7 series has different resolutions, but in code it is fixed to 12 bits while for others it is configurable.

The ADC template files are hard to read. I'd look into it but was wondering what would be best way to do? Actually it would be nice to have a "common" class across all F/G/L STM32 (as much as possible). Our project is made for L4 and F7, where I stumbled over the differences.

salkinium:
> The ADC template files are hard to read.
> I'd look into it but was wondering what would be best way to do? Actually it would be nice to have a "common" class across all F/G/L STM32 (as much as possible). 

Yes, I had my issues with this too when porting G0 ADC. I didn't have a quickfix though (short of rewriting all implementations). I too would like to see more splitting of the commonalities and differences per family on a lbuild level, ie. the lbuild modules can use (semi-)shared templates in other folders, like `env.template("../stm32.common/template.cpp.in")` (or really anywhere inside a repo). The difficult part of extracting the actual code commonalities is still the difficulty on this though.

The C++ interface is supposed to be [the `modm::Adc` class](https://github.com/modm-io/modm/blob/develop/src/modm/architecture/interface/adc.hpp#L51). However, it's still from AVR times and way too simple for the STM32 ADCs. The STM32 ADC implementations don't adhere to it anyways, there is quite some discrepancy between the functionality as you discovered.

You're welcome to start a completely clean API for the L4 and F7, I can port and test the remaining families. (Tip: return `False` from the `prepare()` stage of the "old" modules, this way you can keep them around for reference).

salkinium:
Another idea: Here the cycles is just a map of decimal integers to binary integers, the formatting is always the same. Instead of encoding this "data" so explicitly in the templates, we could move them to the module Python code without much effort. That would make it a little easier to understand the difference in *data* and once we see some commonality and patterns we can move it to modm-devices and maintain it there.
At least that's how I did it for the RCC module and the "max sysclk -> flash latency" mapping.

mikewolfram:
With that mapping, what about specifying the conversion time in milli/micro/nano seconds and then calculate the correct bitfield based on ADC clock?

salkinium:
Yes, absolutely! That would make a lot more sense than just the cycles. It would work analog to the Baudrate calculator, for which there is a generic API now, explicitly for the purpose of being used for other things than just baudrates, like this.

- [API and tests](https://github.com/modm-io/modm/pull/318/commits/2c3026a2e3bf3372e479ca096b0884acf91d5beb#diff-0a313d1254fc4be2fda4b10a6eb955dbR29)
- [Usage and refactoring](https://github.com/modm-io/modm/pull/318/commits/e0ff422dc2432c7915f8f2611dd08a0b0369a0ff)

PR #322, opened 2020-01-20
[docs] Add hint for dependencies

strongly-typed:
numpy, PyX and LaTeX are required for the system_layout.py example

### DIFF
--- a/docs/src/guide/installation.md
+++ b/docs/src/guide/installation.md
@@ -57,7 +57,12 @@ For a less intrusive way to run all scons scripts with Python 3 add this to your
 To compile modm *for macOS* (and not the embedded target) you need to install
 some of these libraries as well, depending on what modm modules you use:
 
-	brew install boost zmqpp gcc
+	brew install boost zmqpp gcc cmake
+
+For all example code that is also run by the CI, you need additional dependencies:
+
+	pip3 install numpy pyx
+	MacTeX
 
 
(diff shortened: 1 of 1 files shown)


salkinium:
XPCC is dying, it's only a matter of time until it is removed ([UAVCAN](https://new.uavcan.org) is the *maintained* replacement).

I don't think the installation guide is the right place for this, perhaps the `:communication:xpcc` module description. I would even say that the instructions for compiling modm for hosted should be there, since its such a niché use-case (only unittests use it, and they should really run on the devices not on hosted!).

salkinium:
I don't understand the use-case here, are you building your own CI? Then shouldn't there be a separate "CI installation guide"? Why push this into the user installation guide, it's already confusing enough.

strongly-typed:
> I don't understand the use-case here, are you building your own CI?

I was just checking if my install on a new machine was complete and working. So I just manually run the commands from the `.circleci/config.yml`.

> XPCC is dying, it's only a matter of time until it is removed

Did not know that, but it was expected. Will you migrate all the existing projects to UAVCAN?

So I close this in lack of (future) relevance.

salkinium:
> Will you migrate all the existing projects to UAVCAN?

Ideally yes, however, I'm still waiting a little for them to become more stable and for me to have some time, motivation and a use-case.

PR #323, opened 2020-01-30
Add installation of openocd for Linux environment

dhebbeker:
Why not add `openocd` for the Linux installation procedure as well?


salkinium (approved):
That's a good point!

PR #331, opened 2020-02-27
Move the check for I2C NACKF to the event handler

### LINKED ISSUE #328: STM32L4 I2C NACK should be handled by I2C_EV
Current implementation is handling the NACK in the error interrupt, but in reality it is an event and should therefor be handled there:

image

I made a quick a dirty fix and copied the part from the I2C_ERR handler to I2C_EV handler right after checking for stop. Works now.
Since I'm not an I2C expert maybe you could tell me, if that's the right position and especially if that needs to be handled in the template file to change it for L4 only? With some advice I'll provide a fix.

mikewolfram:
As explained in the reference manual, NACK is handled by the I2C_EV handler and not the
I2C_ERR handler.
Fixes #328.

### DIFF
--- a/src/modm/platform/i2c/stm32-extended/i2c_master.cpp.in
+++ b/src/modm/platform/i2c/stm32-extended/i2c_master.cpp.in
@@ -280,13 +280,6 @@ namespace
 			I2C{{ id }}->ICR |= I2C_ICR_BERRCF;
 			error = modm::I2cMaster::Error::BusCondition;
 		}
-		else if (sr1 & I2C_ISR_NACKF)
-		{	// acknowledge fail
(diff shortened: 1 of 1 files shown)


salkinium:
Cross-Check: A quick `ack -l "NACKF "` through the cmsis-header-stm32 repo, finds this flag is available on the L0, L4, F0, F3, F4, F7, G0, G4 and H7 devices, which [according to this list](https://github.com/modm-io/modm-devices/blob/develop/tools/generator/dfg/stm32/stm_peripherals.py#L327-L391) maps to the `stm32-extended` driver.
The F4 has this flag too, however in the form of `FMPI2C_ISR_NACKF`, which relates to the fast-mode plus peripheral which isn't really supported anyways.

salkinium (dismissed):

Issue #275, opened 2019-09-06
Passing refrences to Gpio pin port deifinitons to classes 


W-M-D:
Hello, I have an led array that has Rgb leds and i was wondering if I could pass a gpio to a function. It seems that gpios are set to a specific definition at compile time but there's no generic gpio class i can reference in my function definition?  I'm not sure if this is the same as #201... Sorry...

salkinium:
You can pass the GPIOs function, for example `GpioB0::set(bool level)` to your function at runtime. Alternatively you can make your function a template function and pass the entire Gpio into it, but that just works for one function. Unfortunately a fully dynamic Gpio interface with base functions is too expensive, but maybe a computed interface might be interesting for modm.

W-M-D:
That makes me really sad :/ why is it expensive if you pass by reference?  
I was thinking of a class like this and then just passing a pointer of the port to it and having the class store the pointer. This would make it easy to reuse code. (this actually compiles lol) 

class RGB_LED
{
public:

    RGB_LED(Gpio * red_l,Gpio * green_l,Gpio * blue_l );
    
    int8_t set_red();
    int8_t set_green();
    int8_t set_blue();
    uint8_t led_handler();

private:
    
    Gpio * red_led;
    Gpio * green_led;
    Gpio * green_led;

};

salkinium:
Then the Gpio class would need virtual functions, that each specific pin would implement. Those vtables add up to a lot of space, and are especially slow on AVRs. It also prevents efficient implementation of `modm::platform::SoftwareGpioPort` and other template/constexpr helpers. Consider the GPIO classes a universal building block for more dynamic things.

You can make RGB_LED a template class and pass all the GPIOs as template arguments, or create your own GPIO interface with virtual functions and wrap the modm GPIO classes into that.

salkinium:
For reference, this is the fancy stuff you can build with our GPIOs: https://github.com/modm-io/modm/pull/19#issuecomment-390767555

W-M-D:
Ah i didn't consider the AVR case... Thank you again for all that you have done with this project. It's very nice :D

salkinium:
[...]

PR #337, opened 2020-03-03
Add parameters for more control over general purpose timer CR1.

dhebbeker:
For compatibility reasons with legacy code I need more control over the `TIMx->CR1` register. Thus I added more parameters. Default values render the previous behavior.

I am not sure, if this is the best / correct solution?

### DIFF
--- a/src/modm/platform/timer/stm32/general_purpose.cpp.in
+++ b/src/modm/platform/timer/stm32/general_purpose.cpp.in
@@ -44,7 +44,8 @@ modm::platform::Timer{{ id }}::disable()
 void
 modm::platform::Timer{{ id }}::setMode(Mode mode, SlaveMode slaveMode,
 		SlaveModeTrigger slaveModeTrigger, MasterMode masterMode,
-		bool enableOnePulseMode)
+		bool enableOnePulseMode, bool bufferAutoReloadRegister,
+		bool limitUpdateEventRequestSource)
 {
 	// disable timer
 	TIM{{ id }}->CR1 = 0;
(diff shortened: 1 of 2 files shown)


salkinium:
> I am not sure, if this is the best / correct solution?

I… guess? 🤷‍♂
The timers have so many use-cases, it's difficult to make *one* API that serves them all.
Backwards compatibility is great, but adding all those bools isn't ideal in C++, due to a lack of named arguments.

I think @rleh and @se-bi have more stake in the Timer APIs than me, what do you think?

rleh:
At @roboterclubaachen it is common to write the registers directly for complex timer configurations.

rleh:
See this https://github.com/modm-io/modm/issues/275#issuecomment-531222414:
> Implementing a generic API for timers is probably not possible and in my opinion the wrong attempt of abstraction. On STM32 every timer and every channel sometimes has different features/options and behaves slightly different.
> If you look at more manufacturers, it gets even worse in terms of compatibility/interoperability.
>
> I could imagine a platform independent API which provides a PWM signal with adjustable frequency and duty cycle at one pin.

Depending on how much abstraction we consider suitable, the user should not care about the hardware.
For example, he wants to generate PWM with a certain frequency and duty cycle. Then I imagine the interface like this:
```c++
modm::Pwm<Gpio> pwm{100_kHz /* frequency */};
pwm.setDutyCycle(50_pct);
```
Or the user wants to run a DC or BLDC/PMSM motor:
```c++
librobots2::motor::Bldc<Timer, Pins, ...> ...
```
Or measure a pulse width. And so on...

For more complex timer configurations I would suggest to directly access the registers.

Since it is easily possible and well documented (in STs datasheets) to access the registers I don't think we need an pass-through wrapper.

dhebbeker:
I agree that the interface should be clear and unambiguous.

The [code documentation](https://github.com/modm-io/modm/blob/086581565e088c4fda60e8d13a8a8b1fc9f0912b/src/modm/platform/timer/stm32/general_purpose.hpp.in#L40) has this warning:

 > The Timer has much more possibilities than presented by this interface (e.g. Input Capture, Trigger for other Timers, DMA). It might be expanded in the future.

I am not persuaded that directly accessing registers is the way I want to use it¹ as it kind of breaks the programming paradigm. 

¹: The usage, and possibly alternative solution, being outside of modm.

se-bi:
> At @roboterclubaachen it is common to write the registers directly for complex timer configurations.

>For more complex timer configurations I would suggest to directly access the registers.

True, we did that for a case with AdvancedTimer and the I-want-it-to-work-now situation...

(But i would suggest to bring that stuff also into modm, I remember it as quite error prune to get what we wanted out of the documentation into the registers, better for the future to use the ready-to-go solution...)

>I am not persuaded that directly accessing registers is the way I want to use it¹ as it kind of breaks the programming paradigm.

I would agree here, since also the General Purpose Timer (I hope the "basic" one is meant \*) is the same or very similar for stm32 devices or at least some families\*?

\* Correct if i am wrong...

Regarding:


> >I am not sure, if this is the best / correct solution?

> I… guess? :man_shrugging:
> The timers have so many use-cases, it's difficult to make one API that serves them all.
> Backwards compatibility is great, but adding all those bools isn't ideal in C++, due to a lack of named arguments.

Bools..:man_shrugging:
Backwards compatibility :+1: what against that :wink:

rleh (dismissed):
Don't get me wrong: I don't mind merging this pull request at all.

rleh:
> I am not persuaded that directly accessing registers is the way I want to use it¹ as it kind of breaks the programming paradigm.

That's right.
Maybe we should rename the current API to 'TimerHal' (as in Uart) when implementing the application-driven APIs.

salkinium:
> Maybe we should rename the current API to 'TimerHal' (as in Uart) when implementing the application-driven APIs.

Yes!

> Don't get me wrong: I don't mind merging this pull request at all.

Ok, I will rebase and merge these changes.

dhebbeker:
> Ok, I will rebase and merge these changes.

@salkinium: Is it a general policy to merge ff-only for this repository, or why the rebase? I could attempt to comply with that in future.

salkinium:
Yes, otherwise the history is full of merge commits which creates a lot of noise.

PR #342, opened 2020-03-10
Add enableComparePreload parameter to Timer{{ id }}::configureOutputChannel()

dhebbeker:
For compatibility reasons with legacy code I need more control over the `TIMx->CCMRx` register. Thus I added an additional parameter. Default value renders the previous behaviour.

The boolean decides if the `TIM_CCMR1_OC1PE` bits are enabled on the register when configuring output compare channels. The API has no reasonable alternative - except for writing to the registers directly. But that is ugly as channel arithmetic has to be done.


rleh (approved):
:+1:

rleh (approved):

PR #343, opened 2020-03-11
[c++] Move ABI implementation to own module, add option to use exceptions and RTTI

salkinium:
This moves the `__cxa*` function implementations from the `:platform:core` modules to the `:stdc++` module. These functions were always out-of-place in `:platform:core` modules as implementing C++ ABI isn't that dependent on the underlying hardware, and much more on the compiler and library supports.

For Cortex-M targets options to enable C++ exceptions and RTTI is added, should this be desired. Some minor changes to the linkerscript and the compile flags was necessary, however nothing particularly interesting here, except watching the code size explode, yay.

cc @se-bi @chris-durand


se-bi:
Work's for me, work's for my project. :relaxed:
I'm happy with that, thanks!

:+1:

se-bi (approved):

salkinium:
I've added moving the target and language specific compile flags out of the `:build` module and into the modules that are responsible for these flags, ie. `:stdc++` for CXXFLAGS and `:platform:core` for ARCHFLAGS etc. This ties the code implementations closer to the compiler flags (like using `-Wl,-wrap,_malloc_r` to implement malloc) as well as removes some duplicate checks in `:build`.

However, this makes it harder to port to another compiler that isn't GCC compatible, however, since this involved likely a lot more than just adapting compile flags, it's probably not a relevant disadvantage.

Issue #345, opened 2020-03-12
Thread-safe static initialization for interrupts and FreeRTOS


salkinium:
During #343, I discovered that we compile with `-fno-threadsafe-statics`.
Background information (these are some good reads):
- https://iamroman.org/blog/2017/04/cpp11-static-init/
- https://manishearth.github.io/blog/2015/06/26/adventures-in-systems-programming-c-plus-plus-local-statics/

When compiling without this flag, the default implementation of stdlibc++ is used, which compiles to this:
```asm
0x8003158 <__cxa_guard_acquire>         ldr    r3, [r0, #0]
0x800315a <__cxa_guard_acquire+2>       lsls   r3, r3, #31
0x800315c <__cxa_guard_acquire+4>       bmi.n  0x800316c <__cxa_guard_acquire+20>
0x800315e <__cxa_guard_acquire+6>       ldrb   r3, [r0, #1]
0x8003160 <__cxa_guard_acquire+8>       cbz    r3, 0x8003164 <__cxa_guard_acquire+12>
0x8003162 <__cxa_guard_acquire+10>      udf    #255    ; 0xff
0x8003164 <__cxa_guard_acquire+12>      movs   r3, #1
0x8003166 <__cxa_guard_acquire+14>      strb   r3, [r0, #1]
0x8003168 <__cxa_guard_acquire+16>      mov    r0, r3
0x800316a <__cxa_guard_acquire+18>      bx     lr
0x800316c <__cxa_guard_acquire+20>      movs   r0, #0
0x800316e <__cxa_guard_acquire+22>      bx     lr
0x8003170 <__cxa_guard_release>         movs   r3, #1
0x8003172 <__cxa_guard_release+2>       str    r3, [r0, #0]
0x8003174 <__cxa_guard_release+4>       bx     lr
```
which if you [compare this to the source code](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/libsupc%2B%2B/guard.cc#L242-L346) and mentally resolve all necessary #defines, is basically this simple implementation:
```c
  // The ARM EABI uses the least significant bit of a 32-bit
  // guard variable.  */
#define _GLIBCXX_GUARD_TEST(x) ((*(x) & 1) != 0)
#define _GLIBCXX_GUARD_SET(x) *(x) = 1
#define _GLIBCXX_GUARD_BIT 1
#define _GLIBCXX_GUARD_PENDING_BIT __guard_test_bit (1, 1)
[...]

PR #346, opened 2020-03-13
[c++] Add static initialization guards for Cortex-M

salkinium:
This adds a hopefully atomic statics initialization guard implementation to implement #345.
In case of recursive initialization, a debug assertion `cxa.guard.recursion` is raised.
For AVRs, the threadsafe statics are not implemented at all to save code space.

Should the guard implementation be optional for ARM Cortex-M? The guard itself takes 4 bytes per function static, and the guards a little bit of code. I don't think this is too much size, but for very small targets it may be useful to have this as an option. I'm also not aware of this not being atomic being an issue in the past.

Not sure how to overwrite this for FreeRTOS use, I'll fix it when it becomes an issue.

cc @chris-durand @rleh @dergraaf

### DIFF
--- a/examples/stm32f469_discovery/threadsafe_statics/main.cpp
+++ b/examples/stm32f469_discovery/threadsafe_statics/main.cpp
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2020, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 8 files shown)


salkinium:
I also had to implement the GCC atomic builtins for ARMv6-M, since apparently the toolchain delegates the implementation to the user for these things. They are now done using atomic locks, which is fairly brute force, but should work just fine.
https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html

PR #347, opened 2020-03-13
Clear ADC interrupt flags when starting a conversion

### LINKED ISSUE #330: STM32L4 ADC returns wrong values when gcc optimization not -O0
Turning on optimization in gcc (e.g. -Og) results in a weird behavior of the ADC. The very first conversion is fine, all of the following return useless data. Example is reading the temperature sensor, right after start it reads about 34C, the next reading is about 800C.

The only difference I found is the ISR register of the ADC. After reset it is 0, after the conversion it is ~~0x07~~ 0x0c (EOS, EOC and EOSMP). Starting the next conversion clears on EOSMP and OVR, EOC is implicitly cleared.

Modifying startConversion() method like this
```cpp
void
modm::platform::Adc1::startConversion(void)
{
    // TODO: maybe add more interrupt flags
    acknowledgeInterruptFlag(InterruptFlag::EndOfRegularConversion | 
        InterruptFlag::EndOfSampling | InterruptFlag::Overrun);
    // starts single conversion for the regular group
    ADC1->CR |= ADC_CR_ADSTART;
}
```
resolves the issue.

Question is now, if it is ok to simply add the flag to the stm32f3 ADC class for all types? Maybe somebody could check it? Otherwise I'd add a 'if' to the template file to change it for L4 only.

I have absolutely no idea, why without optimization everything is fine. Searching Google brings up few results, that the problem of the ADC with optimization turned on is known, but the solutions provided there didn't work for me. For instance the DIFSEL register is always 0, also a different tool chain didn't resolve the problem.

Using ST gcc 7-2018-q2-update.

mikewolfram:
Clears the interrupt flag EndOfRegularConversion as well when a conversion is started. Thix fixes issues with wrong values read from the ADC when compiling
with optimization turned on.
Fixes #330.

### DIFF
--- a/src/modm/platform/adc/stm32f3/adc_impl.hpp.in
+++ b/src/modm/platform/adc/stm32f3/adc_impl.hpp.in
@@ -185,7 +185,8 @@ void
 modm::platform::Adc{{ id }}::startConversion(void)
 {
 	// TODO: maybe add more interrupt flags
-	acknowledgeInterruptFlag(InterruptFlag::EndOfSampling | InterruptFlag::Overrun);
+	acknowledgeInterruptFlag(InterruptFlag::EndOfRegularConversion |
+			InterruptFlag::EndOfSampling | InterruptFlag::Overrun);
 	// starts single conversion for the regular group
 	ADC{{ id }}->CR |= ADC_CR_ADSTART;
 }


salkinium (approved):
Thanks!

PR #365, opened 2020-03-21
[board] Use pllFactors instead of deprecated method

strongly-typed:


strongly-typed:
argh, just saw that this is already included in #363

salkinium:
Now fixed in #363. 😇

PR #367, opened 2020-03-22
[board] Fix Nucleo F042 I2C clock

chris-durand:
The default clock for I2C1 in F0x2 controllers is the HSI clock, not the APB clock as defined in the board header. From the reference manual:

image


rleh:
Is the I²C bus speed somehow limited with a clock source of only 8 MHz?

chris-durand:
8 MHz is enough for 400 kHz "Fast mode". ST has an example calculation for 500 kHz Fast mode+ and 8 MHz clock in the reference manual. I'm not sure about 1 MHz Fm+ but I have never encountered a peripheral which supports that.

salkinium (approved):

Issue #355, opened 2020-03-17
More flexible firmware upload support for multiple targets and programmers


salkinium:
`scons program` should be renamed to `scons upload` and support a more generic way to upload firmware to the device (preferrably only wrapped by SCons so it can be reused by CMake).

openocd != ST-Link, but so far this is the only combination we use. openocd offers many other [debug adapters](http://openocd.org/doc/html/Debug-Adapter-Configuration.html#Debug-Adapter-Configuration).

[Original comment](https://github.com/modm-io/modm/pull/168#issuecomment-464459385):

> arm-none-eabi-gdb + Black Magic Probe is the other combination I am using.
> 
> Also to be considered:
> * Deal with serial and CAN bootloaders 
>   * at least the ones which are preprogrammed to STM devices
>   * optionally, all 'official' bootloader protocols like app note AVR109 (which is commonly used by Arduino I think)
> 
> > I would like to see a concept for retargeting the existing semantics of the CLI onto the user's specific targets:
> > * `scons program` transparently uses either OpenOCD or BMP or DFU or STLink
> > * `scons gdb` transparently runs both GDB and programmer
> 
> I am in favour of
>  * `scons upload` (or even `flash` if target is flash.)
>  * `scons debug`
> 
> And `debug` can also be something else, not tied to the `gdb` tui forever. Saw some nice web-based tool.

salkinium:
Additionally when having multiple programmers connected, it should be possible to detect *which* targets are connected and automagically choose the right one for the current context.
So if I have a bunch of Nucleos connected, the script needs to probe all targets IDs (once with cache?) and choose the Nucleo that matches the current `:target`. Alternatively a mechanism must be provided to select a specific programmer or specific device (using UID, USB port or serial number?).

It would also be nice to automatically select the right serial port for bootloaders, if that is possible.

rleh:
`scons upload` or `scons download`?
Arguments could be found for both terms, so I am strongly in favour of `scons program`!

salkinium:
Well, 🤓🤓akschtually🤓🤓 OpenOCD is doing both an upload and a download or the other way around to verify the firmware writing was successful. But joking aside, of course we would alias `scons program` to a new command if any.

salkinium:
[...]

PR #370, opened 2020-03-24
[build] Add SCons tools into pure Python library for embedded tooling

salkinium:
We have many cool tools that are basically just Python scripts written in SCons.
This makes it annoying to port to other build systems or to use them without any at all.

I'm moving these tools into a small stand-alone python library called `modm_tools`. These also have integrated command liner parsers to make it easy to use without any build system.
For SCons and (C)Make I'm just wrapping these tools to provide the same level of feature.

Summary of changes:

1. Added independent Python library for programming and debugging targets, unittesting, size reporting, finding files, build/git info, bitmap conversion, build id and logging.
1. Wrapped GDB to generically combine two frontends (tui and gdbgui) with three backend (openocd, bmp and crashdebug).
1. gdbgui has been added as a GDB frontend. \o/
1. Wrapped all commands in SCons and Make, which makes it easier to port features.
1. Unittest generator is now completely self-standing and simply generates a static `int unittest::Controller::run(unittest::Reporter reporter)` function, which frees the runner to be generated by lbuild instead of SCons. This makes it all much less magic and the `modm:build:scons:unittest.runner` option becomes unnecessary.
1. Use gitpython for accessing git information in info tool, instead of parsing git output manually.
1. Provided guessing of serial ports for AvrDude and BMP tools. Works fine most of the time.
1. Generic ITM logging tool via OpenOCD backend.
1. Generic size tool for CMake.
1. Lots of documentation.

Several commands are renamed for consistency and/or added:

- `scons/make program [port={serial}]` programs the target via OpenOCD or AvrDude.
- `scons/make program-bmp [port={serial}]` programs the target via Black Magic Probe.
- `scons program-remote [host={ip of hostname}]` programs the target via remote target.

- `scons/make size` for size report.

- `scons/make debug profile={debug|release} ui={tui|web}` launches GDB with the OpenOCD backend.
- `scons/make debug-bmp profile={debug|release} ui={tui|web}` launches GDB with the Black Magic Probe backend.
- `scons/make debug-coredump profile={debug|release} ui={tui|web}` launches GDB with the CrashDebug backend.
- `scons debug-remote [host={ip of hostname}]` debug remote target.

- `scons reset{-bmp|-remote}` resetting via OpenOCD, BMP or Remotely

Related to #355.

TODO:
- [x] GBD TUI & GUI (tested)
- [x] Local OpenOCD (tested)
- [x] Remote OpenOCD (tested)
- [x] AvrDude (tested)
- [x] Coredump (tested)
- [x] Artifact (tested)
- [x] FindFiles (tested)
- [x] LogITM (tested)
- [x] BMP (tested)
- [x] git info (tested)
- [x] size (tested)
- [x] unittest (tested)
- [x] bitmap (tested)
- [x] command line interfaces where it makes sense
- [x] Super-duper documentation

Things to do for Future Niklas™:
- Relative paths for SCons (no clue)
- font (unused, I think)
- Stlink (unused)
- LPC-Link (unused)
- Correct Dependencies for SCons (couldn't find anything more weird than normal)

cc @rleh @se-bi @siy @dergraaf @strongly-typed


salkinium:
```
scons debug-coredump profile=debug ui=gui firmware=E79E16CEDAE4E476DEEE84B86691C62807475513
```
Now results in this, so all backends of GDB are truly generic. `ui=tui` then is the usual TUI interface.

salkinium:
I'm ready to merge this if nobody complaints ;-P

se-bi:
> I'm ready to merge this if nobody complaints ;-P

Mhm, how much of breaking changes / behavior is that? :wink:

salkinium:
> Mhm, how much of breaking changes / behavior is that? 😉

The breaking changes are:

- Renaming `scons/make gdb` -> `scons/make debug` for #355.
- Renaming `make upload` -> `make program` for consistency.
- `MODM_GIT_UPDATED_NOT_MERGED` was removed, since I couldn't get that information from gitpython, it seems to be computed otherwise.
- The unittest runner is now generated via Jinja2 template interally, generating a .cpp file that only contains the `unittest::Controller::run()` function, that can be called whenever you like. This means the `main.cpp` that sets up the environment for the unittest (in modm/test/runner/*) can  be generated by the `modm-test:test` like a normal template/copy.

The rest should be under the hood changes, and a lot of tools were resurrected and made available in SCons *and* Make.

se-bi:
I was testing the `example/stm32f469_discovery/blink`:

first i got an `EnvironmentError: No module named modm_tools:`

After i added an empty __init__.py within the generated `modm/modm_tools/` (`touch modm/modm_tools/__init__.py`),
i get the following, also for e.g. `examples/linux/logger`:

```
$ scons
info tool: Not a Git repository!
Invalid change type: C
```

salkinium:
> I added an empty __init__.py within the generated `modm/modm_tools/`

Yes, I haz the dumb and just didn't copy the file. Not sure why this worked on my machine *and* the CI. Better not think too much about that or I'll go crazy.

> Invalid change type: C

The info tool is possessed by the Git Devil!1!!

se-bi:
btw. since the tools are refactored here,

executing _scons_ explicitly with _python3_:
```
user@host:~/modm/examples/linux/logger (feature/common_modm_tools)                                                                  [60/662]
$ /usr/bin/env python3 $(which scons) -j1                                                                                                                     
scons: Building targets ...                                                                                                                                   
scons: *** [modm/src/info_git.c] TypeError : sequence item 1: expected str instance, bytes found                                                              
  File "/usr/lib/scons/SCons/Action.py", line 689, in __call__
    cmd = self.strfunction(target, source, env, executor)
TypeError: CacheRetrieveString() takes 3 positional arguments but 4 were given

During handling of the above exception, another exception occurred:

  File "/usr/lib/scons/SCons/Taskmaster.py", line 241, in execute
    if not t.retrieve_from_cache():
  File "/usr/lib/scons/SCons/Node/FS.py", line 2925, in retrieve_from_cache
    return self.get_build_env().get_CacheDir().retrieve(self)
  File "/usr/lib/scons/SCons/CacheDir.py", line 263, in retrieve
    if CacheRetrieve(node, [], env, execute=1) == 0:
  File "/usr/lib/scons/SCons/Action.py", line 691, in __call__
    cmd = self.strfunction(target, source, env)
  File "/usr/lib/scons/SCons/CacheDir.py", line 66, in CacheRetrieveString
    cachedir, cachefile = cd.cachepath(t)
  File "/usr/lib/scons/SCons/CacheDir.py", line 224, in cachepath
    sig = node.get_cachedir_bsig()
  File "/usr/lib/scons/SCons/Node/FS.py", line 3391, in get_cachedir_bsig
    result = self.cachesig = SCons.Util.MD5collect(sigs)
  File "/usr/lib/scons/SCons/Util.py", line 1550, in MD5collect
    return MD5signature(', '.join(signatures))
TypeError: sequence item 1: expected str instance, bytes found
```

se-bi:
Getting closer :stuck_out_tongue_winking_eye: 

the previous issue with ` /usr/bin/env python3 $(which scons)` (on Debian10) remains,
only in the combination of enabled _SconsCache_ and enabled option _info_git_

other rather funny thing:
```
[..]modm/examples/linux/git (feature/common_modm_tools) *                                                                         
$ lbuild build && scons

[..]

File Status:
[..]
Renamed:   0
Copied:    -1
```

salkinium:
I'm disabling SCons Cache for the examples, it is only used for (significantly) speeding up the CI. It gets set by `modm/tools/scripts/examples_compile.py` anyways, and that is only used in the CI.

The `-1` is intentional, this happens when that number could not be determined for whatever reason. I'm not sure how well gitpython works on other platforms, maybe I should've kept the old code, but maybe the git status information is a bit useless too…

salkinium:
>  with /usr/bin/env python3 $(which scons) (on Debian10) remains

I got a similar issue, and solved it with some `.encode("utf-8")`. I'll deal with git info at some later time, I've run out of motivation for this tool ;-P

salkinium:
Merged manually.

PR #372, opened 2020-03-29
Add simple driver for HX711 weight scale / bridge amplifier and integrated ADC

strongly-typed:
@daniel-k also used this device, but as far as I remember without modm ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,44 +193,45 @@ can easily configure them for you specific needs.
 <td align="center">HD44780</td>
 <td align="center">HMC58x</td>
 <td align="center">HMC6343</td>
+<td align="center">HX711</td>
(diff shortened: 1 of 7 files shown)


salkinium (approved):
Very nice!

rleh:
Is there any source for this chip? Can't find it on Mouser/Digikey/Avnet/Arrow/... not even LCSC has stock.

strongly-typed:
E.g. [Sparkfun](https://www.sparkfun.com/products/13879) or your favorite, trustworthy Chinese warehouse ... It comes with the cheapo weight sensors and load cells.

BTW: After the innovative WS18B20, another shift in technology leadership?

PR #383, opened 2020-04-27
[ext] Update submodules

salkinium:
In preparation for SAM device support, we need to update the submodules which leads to changes unrelated to SAM devices. Hence this PR.


ext/arm/core.lb
```diff
@@ -27,7 +27,10 @@ def build(env):
     env.copy("cmsis/CMSIS/Core/Include/cmsis_version.h", "cmsis_version.h")
     env.copy("cmsis/CMSIS/Core/Include/cmsis_compiler.h", "cmsis_compiler.h")
     env.copy("cmsis/CMSIS/Core/Include/cmsis_gcc.h", "cmsis_gcc.h")
-    if "m0" not in core:
+    if core != "0": # 0+ has MPU support though!
         env.copy("cmsis/CMSIS/Core/Include/mpu_armv7.h", "mpu_armv7.h")
+    print(core)
```
mikewolfram, line 32: Do we need to print the core version?

salkinium: I'm getting old, I need to be reminded what I'm working with every build!1!!

salkinium: After merging the DMA PR, I'll push a commit to remove this.

PR #391, opened 2020-05-01
[ci] Build status in README

strongly-typed:


strongly-typed:
Ahh, it might be autogenerated? Let me check.

strongly-typed:
> Ahh, it might be autogenerated? Let me check.

No, it isn't. It is edited in place by a script.

salkinium:
We removed the build status on purpose:

1. The CircleCI status is not accessible for outside developers, not even for signed-in GitHub users, only for owners and contributers. We cannot link to the latest job either.
2. Only the PR runs the tests, the develop branch only runs the docs generation and upload jobs.
3. The build status is already shown by GitHub itself with the green checkmark at the top.
4. The develop branch jobs may fail, due to network issues when uploading the job.

salkinium:
Additionally I believe this is a superiour UI, since it shows *all* sub-tests in one place, instead of linking to two different CI providers, that require sign-in.
Screenshot 2020-05-01 at 21 53 38

But I am going to disable TravisCI for the develop branch, it's only supposed to run for PRs.

salkinium:
So… I have no intention to merge this, due to above reasons. Is that ok with you? 😵

strongly-typed:
Totally agree. Wasn't really aware of the green checkmark. Thanks for the in-depth reasoning.

PR #394, opened 2020-05-05
Explicitly set the stack pointer in the reset handler.

### LINKED ISSUE #393: Help needed: table_zero crash during start up
Hi,

I'd need some help figuring out what's going wrong on my L431. I have class which has some larger data members. During start there is a crash and the PC goes to 0xfffffff9. What could be the reason, how to trouble shoot it? Compiling debug with -O0, code size is 84kB, data size 12kB.

Actually it is a protocol handler for SECS messages, here are the failing data members:

```c++
struct Message {
    enum class
    FormatCode : uint8_t {
        List = 0,
	Binary = 0x20,
	Boolean = 0x24,
	Ascii = 0x40,
	Jis8 = 0x44,
	Unicode16 = 0x48,
	Signed8Byte = 0x60,
	Signed1Byte = 0x64,
	Signed2Byte = 0x68,
	Signed4Byte = 0x70,
	Float8Byte = 0x80,
	Float4Byte = 0x90,
	Unsigned8Byte = 0xa0,
	Unsigned1Byte = 0xa4,
	Unsigned2Byte = 0xa8,
	Unsigned4Byte = 0xb0,
    };
    static constexpr uint8_t i(FormatCode format) { return uint8_t(format); }

    struct Header {
	uint16_t deviceId;
	uint8_t streamId;
	uint8_t functionId;
	uint16_t blockNummber;
	uint32_t systemBytes;
    } __attribute__((packed));

    // some methods...
    uint8_t length;
    using Data = std::array<uint8_t, 256>;
    Data data;
    uint16_t crc;
};

class ProtocolHandler : public mode:pt:Protothread
{
public:
    // ...
private:
    // ...
    modm::atomic::Queue<Message, 10> rxQueue;
    modm::atomic::Queue<Message, 10> txQueue;
    // ...
};
```

When I reduce the size of the queue to 1 it runs without issues.

mikewolfram:
It turned out that running from debugger the stack pointer might be wrong
(initialized to the value of the system flash).

### DIFF
--- a/src/modm/platform/core/cortex/reset_handler.sx
+++ b/src/modm/platform/core/cortex/reset_handler.sx
@@ -16,6 +16,8 @@
 	.type	Reset_Handler, %function
 	.func   Reset_Handler
 Reset_Handler:
+	ldr r0,=__main_stack_top
+	mov sp,r0
 	bl __modm_initialize_platform
 	b __modm_startup
 	.endfunc


salkinium:
image

I think, you first need to load it into r0, then mov it to sp. BECAUSE REASONS.

salkinium:
This should work for both ARMv6-M and ARMv7-M:
```asm
ldr r0,=__main_stack_top
mov sp,r0
```

salkinium:
> This should work

Tested this on Cortex-M0 and Cortex-M4, this works.

mikewolfram:
> > This should work
> 
> Tested this on Cortex-M0 and Cortex-M4, this works.

Done

salkinium (approved):

PR #395, opened 2020-05-06
[communication] Revive Asynchronous Multi-Node Bus (AMNB) protocol

salkinium:
A flexible multi-node bus using p-persistent CSMA/CD over a shared wired medium.
This can work with using UART with differential signalling (typically CAN transceivers for automatic direction control) or without for short distances with TX in Open-Drain mode with an external pull-up, for some STM32 even just using a single pin via the built-in half-duplex mode for UART.

Consider this a low-cost alternative to I2C, with the goal to be more stable, more flexible and faster.


salkinium:
Design goals:
- important information (address, command, type) in the header first, so I can discard the message if it's not for me.
- very little overhead for small messages, here CRC8 can protect (max 256 bits => 32 bytes incl. header).
- support large payloads via CRC16 (max 2^16 bits => 8kB).
- small payload header is compatible with large payload header, requiring only the first 4 Bytes to decide if its a small or large payload, and max 32 Bytes to decide whether the heade is valid (because small payload is part of the header CRC8).
- to have the payload in a continous memory to be able to `reinterpret_cast` the payload on-device to whatever the type is, to avoid explicit serialization just like in XPCC. This requires 4-bytes alignment of both payloads.
- Simply "filling" the values on the wire into the packed message struct is enough. This involves a choosing a layout that avoids unaligned access for Cortex-M0, requiring 2-bytes alignment for uint16_t and 4-bytes alignment for uint32_t pointers.
- Option to convert payload into shared_ptr, which requires only copying small payloads and because the large payloads are already allocated on the heap anyways.
- Delegating memory management to the application or at least the node, so the protocol stack doesn't require any buffers and only operates on messages owned somewhere else.

The C++ api is still WIP and currently untested.

@dergraaf You're a protocol wizzard, would you mind giving some feedback? 
@rleh @chris-durand @se-bi I'm using this for the ELVA semaphores, but perhaps this interesting for RCA use-cases too?

salkinium:
Things that really bug me: There seems to be no way in the UART hardware to indicate whether a symbol is *currently* being received. All the RX flags are only set *after* the full symbol has been received. This feature could turn the "Collision Detection" into a "Collision Avoidance", which would be a nice stretch goal.

src/modm/communication/amnb/handlers.hpp
```diff
@@ -0,0 +1,363 @@
...
+	Callback function;
+	uint8_t command;
+	uint8_t address;
+};
+
+}
+
+/// @cond
+#define	MODM_AMNB_ACTION(command, length, function)		\
+	{	reinterpret_cast<modm::amnb::Action::Callback>(&function),
```
chris-durand: I would suggest to replace this `reinterpret_cast` with something more type safe. The present solution will compile even when the user puts in a function with wrong argument count or something not callable at all. As far as I understand it, the purpose of this cast is to allow the user to provide a callback of type `void (*)(Response& response, const T *payload)` but to store it internally as `void (*)(Response& response, const void *payload)`. Calling the function through the cast pointer is undefined behavior anyway.

I might have a working solution which checks the correct function signature at compile time and is not undefined (if the payload actually is a `T*`). You could use a lambda which forwards the call to the user function and has the type `Action::Callback`. If it has no capture it will convert to a plain function pointer and can be stored within your Action class. This way you only cast the data from `void*` to `T*` in the lambda but not the function pointer itself. Thus, the compiler can check the function signature. You could also make it magically work with member function pointers. [Here](https://gcc.godbolt.org/z/bSEbU6) is a quick proof-of-concept in Compiler Explorer.

chris-durand (changes requested):

src/modm/communication/amnb/handlers.hpp
salkinium: Excellent, exactly the feedback I'm looking for!

This whole callback stuff was adapted from SAB without thinking, so I'll redo the whole thing entirely from scratch with your ideas.

src/modm/communication/amnb/message.hpp
```diff
@@ -0,0 +1,157 @@
...
+	inline bool allocateData(uint16_t max_size=0)
+	{
+		if (length() > SMALL_LENGTH) {
+			const size_t size = max_size ? std::min(length(), max_size) : length();
+			large.data.reset(new uint8_t[size]);
+		}
+	}
+
+protected:
+	// NOTE: This arrangement allows this data structure to be 4-byte aligned
```
chris-durand: Do you think an [`alignas()` specifier](https://en.cppreference.com/w/cpp/language/alignas) would make sense here to explicitly enforce alignment?

src/modm/communication/amnb/message.hpp
```diff
@@ -0,0 +1,157 @@
...
+	//       so that any Cortex-M0 device has no issues with unaligned access
+	//       of neither the header values nor the data cast to another struct!
```
chris-durand: Sadly, there is no way to make `char buffer[...] = {...}; reinterpret_cast<some_struct*>(buffer)->value` legal C++ yet. There is [p0593](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0593r6.html) but the library part (`std::start_lifetime_as<T>()`) won't be in C++20 and will probably come in C++23. We can just hope no compiler optimizer will actually mess with the code until then...

src/modm/communication/amnb/message.hpp
salkinium: Yes!

salkinium: I think ARM EABI requires either 4B or 8B alignment of objects, so I think this isn't useful. On AVRs I'm not sure what the alignment requirements are, could be 2B for pointers?
So not that useful.

src/modm/communication/amnb/message.hpp
salkinium: Hm, that's bad because that's exactly what I need to do here. I've added a basic implementation to create a union of `uint8_t[]` and a `std::shared_ptr<uint8_t[]>` that I'm reinterpret casting to variable types…

I'm starting to suspect that none of this is legal in C++, and it probably doesn't even do the correct thing with calling constructors/destructor etc. Is there any way to do this right? IIRC doing *anything* with the underlying memory of non-POD objects is dangerous, so "serializing" this data by simply copying the underlying memory is wrong?

chris-durand: > I'm starting to suspect that none of this is legal in C++, and it probably doesn't even do the correct thing with calling constructors/destructor etc. Is there any way to do this right? IIRC doing anything with the underlying memory of non-POD objects is dangerous, so "serializing" this data by simply copying the underlying memory is wrong?

If your type is neither trivially default constructible or trivially copy constructible the whole approach is not sane. Either copying your object must be equivalent to copying bytes (memcpy) or any data must be a valid representation of your object.

A source of undefined behavior here is C++'s object model. You can't use an object that has not been created.

In C++ the following things create objects (from the standard draft)
> An object is created by
>   - a definition,
>   - by a new-expression,
>   -  when implicitly changing the active member of a union,
>   -  when a temporary object is created. 

and with the reinterpret_cast we do none of these.

For this reason, calling `malloc()`, casting the result to a POD struct and using it is strictly speaking undefined behaviour in C++ before 20. In C++20 this is solved with [p0593](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0593r6.html) where they add that some operations implicitly create objects.

[placement new](http://www.cs.technion.ac.il/users/yechiel/c++-faq/placement-new.html) is probably the solution. But to do the placement new with the right type your code needs to know the type. This would imply some template things somewhere.

salkinium: Ok, thanks for the input. I've removed shared_ptr in favor of a simple manual memory manegement. I'm prematurely optimizing things here.

> copying your object must be equivalent to copying bytes (memcpy)

Yes, and it makes no sense to have it any other way for sending an obj state over wire. So fancy constructors are not possible anyways, even POD can be dangerous if they contain pointers. But this is something the user-facing API needs to check, perhaps even with Concepts (in the future).

salkinium:
I'll add documentation and AVR examples at a later stage.

PR #396, opened 2020-05-06
Add nucleo_f746zg board

FelixPetriconi:
Improve slightly the docs

PR #401, opened 2020-05-08
How to create a SystemClock struct

rleh:
From https://github.com/modm-io/modm/pull/396#discussion_r422372031


salkinium (approved):
Thanks!

PR #403, opened 2020-05-12
Fixes __wrap__free_r() to ignore freeing null pointers.

mikewolfram:
Moving the heap into its own module also changed the default allocator from newlib to tlsf, which fails to "delete" null pointers.

This fix adds the missing check for null pointer as it can be found in ext/tlsf.


salkinium (approved):
Thanks!

PR #404, opened 2020-05-13
[scons] add command-line arguments to manipulate CCFLAGS,CFLAGS,etc

rleh:

```
```

---

(Same as #124, which I'm unable to reopen)


salkinium (approved):
Gogogo

PR #407, opened 2020-05-22
[spi] Fix writing SPI data size into correct register

### LINKED ISSUE #406: SPI data size written to wrong register on L431
Hi,

figured out that the function to set the SPI data size on a L431 is wrong. The bits are written to CR1, but must be written to CR2. SpiXHal::initialize() does it right, but SpiXHal::setDataSize() does it wrong.

Could it be that the check for data size in features needs to be negated?
```c++
void inline
modm::platform::SpiHal{{ id }}::setDataSize(DataSize dataSize)
{
%% if "data-size" in features
	// TODO: implement as set/reset bit
	SPI{{ id }}->CR1 = (SPI{{ id }}->CR1 & ~static_cast<uint32_t>(DataSize::All))
										 | static_cast<uint32_t>(dataSize);
%% else
	SPI{{ id }}->CR2 = (SPI{{ id }}->CR2 & ~static_cast<uint32_t>(DataSize::All))
										 | static_cast<uint32_t>(dataSize);
%% endif
}
```
When you look into initialize() then the check is
```
%% if "data-size" not in features
```

Thanks.

mikewolfram:
Also adds the missing method to the SPI master class.
Fixes #406.

### DIFF
--- a/src/modm/platform/spi/stm32/spi_hal_impl.hpp.in
+++ b/src/modm/platform/spi/stm32/spi_hal_impl.hpp.in
@@ -80,8 +80,8 @@ modm::platform::SpiHal{{ id }}::setDataOrder(DataOrder dataOrder)
 void inline
 modm::platform::SpiHal{{ id }}::setDataSize(DataSize dataSize)
 {
-%% if "data-size" in features
 	// TODO: implement as set/reset bit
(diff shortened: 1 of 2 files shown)


salkinium (approved):
Thanks!

PR #413, opened 2020-06-07
Different Clock Config for Nucleo F042K6 to fix UART timings

nesos:
This is a rather small fix to setup the clock on the STM32F042 differently. The result is the same but for some weird reason, with a prediv of 1 and a multiplier of 6, the UART seems to have timing issues and only prints out garbage. With a prediv of 2 and a multiplier of 12 it works just fine.

I tried the blink demo with two different Nucleo-F042K6 and the results were the same. I'm not exactly sure why this is an issue after all. According to the datasheet the originial settings should work.

### DIFF
--- a/src/modm/board/nucleo_f042k6/board.hpp
+++ b/src/modm/board/nucleo_f042k6/board.hpp
@@ -55,10 +55,10 @@ struct SystemClock {
 	enable()
 	{
 		Rcc::enableInternalClock();	// 8MHz
-		// (internal clock / 1) * 6 = 48MHz
+		// (internal clock / 2) * 12 = 48MHz
 		const Rcc::PllFactors pllFactors{
-			.pllMul = 6,
-			.pllPrediv = 1
+			.pllMul = 12,
+			.pllPrediv = 2
 		};
 		Rcc::enablePll(Rcc::PllSource::InternalClock, pllFactors);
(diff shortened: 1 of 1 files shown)


rleh:
Which Uart? Usart1, Usart2 or both?
Usart1 has a clock mux, but that should not change anything here...

rleh (approved):
image

salkinium (approved):

nesos:
> Which Uart? Usart1, Usart2 or both?
> Usart1 has a clock mux, but that should not change anything here...

It's USART2, the one connected to the debug port.
I also checked with the CubeMX tool and it doesn't give me any errors, when I put in the prediv 1, multiplier 6 config.

rleh:
I discovered the actual issue: `Rcc::enablePll(Rcc::PllSource::InternalClock,...)` selects the Hsi clock divided by 2 (and not divided by prediv) as pll source.

```c++
enum class PllSource : uint32_t {
  /// High speed internal clock (8 MHz)
  Hsi = RCC_CFGR_PLLSRC_HSI_DIV2,
  InternalClock = Hsi,
  // [...]
}
```
```c++
#define RCC_CFGR_PLLSRC_HSI_DIV2 (0x00000000U) /*!< HSI clock divided by 2 selected as PLL entry clock source */
```

grafik

PR #416, opened 2020-06-10
Add modm:platform:cortex-m:linkerscript.override option

### LINKED ISSUE #411: Support for custom linker script
Hi, 

Is there any support for -- or intention for -- proving a custom linker script? I'd like to be able to carve out a couple sectors of flash for non-volatile app storage. For example: put the ISR vectors in the first sector, then use two of the small 16K sectors at the top of flash for app config, and then place the application code in the remaining flash sectors. 

I'm not seeing any way to disable the `-Tlinkerscript.ld` option in modm, which is all I need really to provide my own in my CMakeLists.txt. I guess I will simply disable the generation of this flag for now; would there be interest in a PR to make this an option that can be set in project.xml?

mcbridejc:
Allow a project to provide its own linkerscript, by setting the `modm:platform:cortex-m:linkerscript.override` option to a path relative to the top-level project directory. 

Closes #411

### DIFF
--- a/src/modm/platform/core/cortex/module.lb
+++ b/src/modm/platform/core/cortex/module.lb
@@ -190,6 +190,15 @@ def prepare(module, options):
             maximum=hex(flash_size),
             default=0))
 
+    module.add_option(
+        PathOption(
+            name="linkerscript.override",
+            description="Path to project provided linkerscript",
+            default="",
+            empty_ok=True,
+            absolute=False,
+        ))
+
     module.add_collector(
         StringCollector(
             name="linkerscript.memory",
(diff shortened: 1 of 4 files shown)


src/modm/platform/core/cortex/module.lb
```diff
@@ -315,9 +324,13 @@ def build(env):
     if env.has_module(":architecture:unaligned"):
         env.template("unaligned_impl.hpp.in")
 
+    env.collect(":build:linkflags", "-Wl,--no-wchar-size-warning", "-L{project_source_dir}")
     # Linkerscript
-    env.collect(":build:linkflags", "-L{linkdir}", "-Tlinkerscript.ld",
-                                    "-Wl,--no-wchar-size-warning")
+    linkerscript_override_option = env["modm:platform:cortex-m:linkerscript.override"]
```
salkinium:
```suggestion
    linkerscript_override_option = env["linkerscript.override"]
```
(The name resolver is context aware, makes this more maintainable)

salkinium (changes requested):
Very nice!

salkinium (approved):
Thanks!

salkinium:
Squashed and adapted commit message, will be merged after CI passes.

PR #422, opened 2020-06-20
Added checks for tx and rx calls

nesos:
Improved the checks for receiving and transmitting message for the SX1276 driver.
This prevents calling transmit and receive in the wrong modes and fixed an error in the flag handling of the IRQ registers

### DIFF
--- a/src/modm/driver/radio/sx1276_impl.hpp
+++ b/src/modm/driver/radio/sx1276_impl.hpp
@@ -152,34 +152,43 @@ Sx1276<SpiMaster, Cs>::setSyncWord(uint8_t syncWord)
 // -----------------------------------------------------------------------------
 
 template <typename SpiMaster, typename Cs>
-modm::ResumableResult<void>
+modm::ResumableResult<bool>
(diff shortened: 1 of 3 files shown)


salkinium (approved):

src/modm/driver/radio/sx1276_impl.hpp
```diff
@@ -152,34 +152,43 @@ Sx1276<SpiMaster, Cs>::setSyncWord(uint8_t syncWord)
...
+		{
+			Cs::set();
+		}
+
+		RF_CALL(changeMode(ModemMode::TX));
+	}
+	else
 	{
-		Cs::set();
+		MODM_LOG_ERROR<<"SX1276: Transmission failed as the modem is busy"<<modm::endl;
```
rleh, line 188: I thought logging in a driver to MODM_LOG_ERROR was not desirable? @salkinium 

Ok, the driver depends on `modm:debug` :+1:

salkinium: It's not, but there isn't an alternative. It would be nice to completely remove/configure logging at compile-time if `modm:debug` isn't available, but we don't have that yet.

PR #426, opened 2020-07-02
Driver for the LSM6DS33 Acceleration and Gyro Sensor

nesos:
This is a driver for the LSM6DS33 Acceleration and Gyro Sensor by ST.
It can read the raw and scaled values from the sensor and all data rates and scales can be configured.
I added an example using the STM32F042 Nucleo and the Pololu IMU-10 v5


salkinium (approved):
Very nice!

PR #431, opened 2020-07-08
[ci] Fix Travis CI Python version

salkinium:
Does some `find` hacking to make the fix more robust and independent of the Python version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ before_install:
   - brew update
   - brew tap osx-cross/avr
   - brew tap osx-cross/arm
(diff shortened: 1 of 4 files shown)


rleh:
Isn't there a symlink `/Users/travis/Library/Python/3/` that links to `/Users/travis/Library/Python/3.7/`?

salkinium:
> Isn't there a symlink /Users/travis/Library/Python/3/ that links to /Users/travis/Library/Python/3.7/?

Nope. But the CI is now also not finding GCC-10, so it tries to compile the hosted tests with LLVM… not sure what that is about.

Issue #432, opened 2020-07-11
Wait for interrupt in resumeables


henrikssn:
Currently, modm isn't exactly sleep friendly. I want to improve that.

The resumeables are built around two states; running (wait for condition) and stop. I propose to add a third state "wait for interrupt" which indicates that the resumeable is not going to make progress until an interrupt occurs.

What the main scheduler then can do is to run the resumeables until they all are blocked on wait for interrupt (or stopped), then put CPU to sleep and let the interrupt wake it back up (and repeat). The interrupt could be timer/rtc (for delays), external or anything else the platform supports while idling/sleeping.

Most likely the main challenge is to make sure we don't accidentally put the interrupt source to sleep, however, maybe that could be handled separately by some sleep/power controller.

Happy to hear your thoughts on this before I get my hands dirty and prototype something.

salkinium:
> I propose to add a third state "wait for interrupt" which indicates that the resumeable is not going to make progress until an interrupt occurs.

Hm, difficult. There is no main schedulers, it's all implicitly scheduled by the control flow of the program, but you could of course access shared memory.

Perhaps something like `RF_WAIT_WHILE/UNTIL` would increment a shared counter that signifies how many resumable are currently *polling* for a condition (and decrement afterwards), and perhaps `RF_WAIT_FOR_INTERRUPT(condition);` would *not* increment/decrement this counter, so that the main loop simply goes to sleep if the counter is zero.
Note that you cannot poll the NVIC interrupt flag directly (ie. `UART_IRQn`) because the interrupt has to reset this flag before exiting the interrupts, otherwise the NVIC will trigger the interrupt again (causing a deadlock). So you need to poll a volatile variable that's set in the interrupt (and reset by who?).

In case the counter is not zero, you still have to poll for the interrupt as normal, but you don't need to build up the entire call stack, you could poll for the interrupt at a "higher" call point (however that would be implemented).

But it won't be as useful as you think. For example all `modm::Timeout`s and `modm::PeriodicTimer`s are polling, especially the µs resolution ones. And the SysTick only fires at 1ms intervals on ARMv6-M and 250ms intervals on ARMv7-M (see [this PR as for why](#217)), so just having *one* resumable that waits on even just a millisecond timer will break sleeping.

Oh, and then there's also the possiblity of just manually polling `if (timer.execute())` in the main loop, and you're allowed to mix this with resumables/protothreads, so you're not going to have fun prototyping this 😭.

[...]

PR #434, opened 2020-07-13
Add WaitForInterrupt to resumeables

henrikssn:
More context in #432

### DIFF
--- a/src/modm/processing/resumable/macros.hpp
+++ b/src/modm/processing/resumable/macros.hpp
@@ -106,13 +106,36 @@
 #define RF_WAIT_UNTIL(condition) \
 	RF_WAIT_WHILE(!(condition))
 
+/// Cause resumable function to wait **while** given `condition` is true, while
+/// signaling that it is expecting an interrupt to happen before. This is
+/// useful to let the CPU idle/sleep until an interrupt happens.
+/// @hideinitializer
+#define RF_WAIT_FOR_INTERRUPT_WHILE(condition) \
+		do { \
+			RF_INTERNAL_SET_CASE(__COUNTER__); \
+			if (condition) { \
+				this->popRf(); \
+				return {modm::rf::WaitForInterrupt}; \
+			} \
+		} while(0)
+
+/// Cause resumable function to wait **until** given `condition` is true, while
+/// signaling that it is expecting an interrupt to happen before. This is
+/// useful to let the CPU idle/sleep until an interrupt happens.
+/// @hideinitializer
+#define RF_WAIT_FOR_INTERRUPT_UNTIL(condition) \
+	RF_WAIT_FOR_INTERRUPT_WHILE(!(condition))
+
 /// Calls a resumable function and returns its result.
 /// @hideinitializer
 #define RF_CALL(...) \
 	({ \
 			RF_INTERNAL_SET_CASE(__COUNTER__); \
 			auto rfResult = (__VA_ARGS__); \
-			if (rfResult.getState() > modm::rf::NestingError) { \
+			if (rfResult.getState() == modm::rf::WaitForInterrupt) { \
+				this->popRf(); \
+				return {modm::rf::WaitForInterrupt}; \
+			} else if (rfResult.getState() > modm::rf::NestingError) { \
 				this->popRf(); \
 				return {modm::rf::Running}; \
 			} \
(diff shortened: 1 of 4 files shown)


salkinium:
Is this more of a Draft, ie. do you want to add some examples/scheduler code or do you want this merged as is?

henrikssn:
Sorry, definitely needs examples/docs, I was just looking for some initial feedback to learn how this matched our discussion earlier.

One thing that came to my mind is that there does not seem to be a pattern for running resumeables concurrently at the moment (which doesn't make sense to me, then we could just do things blocking). What am I missing?

salkinium:
`RF_CALL_BLOCKING` is just a convenience/debug macro, without forward progress guarantee (it doesn't update anything else after all). Resumables are only scoped around (nested) functions, the application should consists out of several concurrent Protothreads that call resumables via `PT_CALL`:

```c++
bool MyThread::update()
{
    PT_BEGIN();
    while(true)
    {
        PT_CALL(resumable());
        PT_YIELD(); // or wait until timer, or on some condition
    }
    PT_END();
}
```

You would then run multiple protothread together by calling their `update()` function in the main loop. That's the bespoke way to run resumables concurrently, and it mirrors the use of pre-emptive threading, where you cannot call two "normal" functions at the same time either.

This is mostly a semantical difference, the protothread implementation is slightly different and I had to keep it for backwards compatibility. You can implement the same with Resumables just slightly more annoying. With C++20 there would probably not be a difference between Protothreads and Resumables anymore, this is currently just historical baggage.

If you mean, running two resumables in parallel, via some "join" operation this is possible with non-nested resumables, which have a `obj.join({1,2,3,4})` function, but it's not very user-friendly. The RF_CALL macro magic limits the possiblities here, this would be better with C++ coroutines objects, with their proper call API. In general I'm not happy exposing such a low-level API, I've been bitten a lot by raw threads and really enjoy using abstractions like worker thread pools or event loops, etc…, so I didn't advertise such operations on resumables in the docs/examples.

salkinium:
So in this context, you would implement the call to `WFI()` in `PT_CALL` macro, right? And make the leaf optimization part of the Protothread class implementation? The Resumables are just communicating their state up, the protothread has to interpret it.

So the implicit scheduler consists of:
 
1. `main() { while(1) {} }` loop calls `Protothread.update()`
2. Protothread calls resumable via `PT_CALL(resumable)` (it can also spawn other protothreads via `PT_SPAWN()`, these were the precursor to resumbales, so their never used anymore.).
3. Resumable calls (nested) resumable via `RF_CALL(resumable)`.

henrikssn:
Hmm, I still don't see how protothreads run multiple things at once (they are just resumables with different name, right?)

If I understand correctly, `PT_SPAWN` is waiting on completion of the child thread which does not allow for concurrency.

Take this example:

```c++
bool ChildOne::update()
{
    PT_BEGIN();
    while(true)
    {
        PT_WAIT_UNTIL(timer.execute());
        LedOne::toggle();
    }
    PT_END();
private:
  PeriodicTimer timer(100ms);
}

bool ChildTwo::update()
{
    PT_BEGIN();
    while(true)
    {
        PT_WAIT_UNTIL(timer.execute());
        LedTwo::toggle();
    }
    PT_END();
private:
  PeriodicTimer timer(200ms);
}
```

and the main loop implemented as `main() { while(1) { mainThread.update() } }`.

How do you implement the mainThread so that LedOne toggles every 100ms and LedTwo every 200ms?

salkinium:
There is no mainThread, you have to call every thread's update method manually. modm does not schedule anything for you at all. *You* are the scheduler!

You can group your threads manually into another thread, and "centralize" the call to update() of your "child" threads, but you still have to call their update() functions *somewhere*.

```c++
Thread1 thread1;
Thread2 thread2;
modm::ShortPeriodicTimer tmr(1s);

bool
MoreThreads::update()
{
    // Call before PT_BEGIN()
    thread3.update();
    thread4.update();

    PT_BEGIN();
    while(1)
    {
        PT_YIELD();
    }
    PT_END();
}

MoreThreads moreThreads;

int main()
{
    Board::initialize();

    while(1)
    {
        thread1.update();
        thread2.update();
        moreThreads.update();

        // "inline" Protothread
        if (tmr.execute()) LedThree::toggle();
    }
    return 0;
}
```

henrikssn:
Thanks, that makes it clear.

What is missing then is some way to check the result type of all the "top level threads" and either loop around or issue WFI instruction. Sure, you can already do something like

```
    while(1)
    {
        sleep = true;
        if(resumeable1.resume().getState() == modm::rf::WaitForInterrupt) sleep &= true;
        if(resumeable2.resume().getState() == modm::rf::WaitForInterrupt) sleep &= true;
        if(resumeable3.resume().getState() == modm::rf::WaitForInterrupt) sleep &= true;
        // ...
        if (sleep) WFI();
    }
```

But maybe we want to prettify this? The boring part is that it would need to be implemented twice if we also want it for protothreads...

salkinium:
> But maybe we want to prettify this?

Yes, that's what I meant with adding a separate module `:processing:sleep`, that centralizes the configuration of this.

salkinium:
I'm closing this in favor of fibers and also in the long term the removal of protothread/resumable functions (in favor of fibers).

PR #442, opened 2020-08-10
Fix the method isConversionFinished() to wait on EndOfConversion instead of EndOfSampling.

mikewolfram:
For some reason the ADC of the F3 implementation was waiting on EndOfSampling instead of EndOfConversion. This causes invalid data as the ADC DR doesn't contain the converted data yet, but still holds the data of the previous conversion.

### DIFF
--- a/src/modm/platform/adc/stm32f3/adc_impl.hpp.in
+++ b/src/modm/platform/adc/stm32f3/adc_impl.hpp.in
@@ -194,7 +194,7 @@ modm::platform::Adc{{ id }}::startConversion(void)
 bool
 modm::platform::Adc{{ id }}::isConversionFinished(void)
 {
-	return static_cast<bool>(getInterruptFlags() & InterruptFlag::EndOfSampling);
+	return static_cast<bool>(getInterruptFlags() & InterruptFlag::EndOfRegularConversion);
 }
 
 // ----------------------------------------------------------------------------


salkinium:
Didn't you fix a similar issue before?

mikewolfram:
Out of curiosity, the other implementations call the flag _EndOfConversion_, but on F3 it is called _EndOf**Regular**Conversion_? Is there a specific reason for that?

mikewolfram:
> Didn't you fix a similar issue before?

No, it was about clearing the flags before starting the next conversion. And looking back, yes I should have seen that at the same time. But the previous fix solved my problem somehow, now that we are running continuous tests I figured out, that once in a while I get wrong results, especially when you also read the Temperature and Battery values. After reading the battery the following conversion of an input always showed the same value that the battery had before.

salkinium:
The ADC on STM32 has always given me nightmares, since it's such an irregular implementation across the families. So it didn't get as much attention as it should have. Maybe @rleh can review?

mikewolfram:
> The ADC on STM32 has always given me nightmares, since it's such an irregular implementation across the families. So it didn't get as much attention as it should have. Maybe @rleh can review?

If I had more time I'd like to rework the implementation, kind of a simple to convert single channels and more advanced to use DMA and allow scanning multiple channels (continuously). My brain is already working on it, but the body is busy doing the other work. :-D

chris-durand (approved):

salkinium (approved):

rleh:
> Out of curiosity, the other implementations call the flag EndOfConversion, but on F3 it is called EndOfRegularConversion? Is there a specific reason for that?

The F3 ADC IP has both _regular_ and _injected_ conversions.

rleh (approved):

PR #443, opened 2020-08-12
fixed issue with escaping of arguments for size script

gueldenstone:
I found that the escaping of the spaces in the argument of the custom command for the "size.py" script did not work correctly. The "VERBATIM"-Option makes sure, that it works on all platforms. I came across this issue on mac OS Catalina.

for reference from [documentation](https://cmake.org/cmake/help/latest/command/add_custom_command.html "documentation")
>VERBATIM
>All arguments to the commands will be escaped properly for the build tool so that the invoked command receives each argument unchanged. Note that one level of escapes is still used by the CMake language processor before add_custom_command even sees the arguments. Use of VERBATIM is recommended as it enables correct behavior. When VERBATIM is not given the behavior is platform specific because there is no protection of tool-specific special characters.


salkinium (approved):

salkinium:
How did you encounter this bug though, since the post build method is commented out?

gueldenstone:
I wanted to use the feature, so I uncommented it. I interpreted the "fix me" message above as a hint to manually configure it. But there should be a way to configure the correct path to the tools directory so that the line can be uncommented by default since modm configures this cmake script. I will try to figure it out and push another pull-rq if I find a solution. Or is there any other reason why the size tool should not be used in the CMake context? I'm quite new to modm so I don't have any knowledge about possible side effects.

salkinium:
> Or is there any other reason why the size tool should not be used in the CMake context?

No, I don't even remember the original problem with getting the build path, I think it was something about relative paths, but I'm not sure.

PR #454, opened 2020-08-31
Stm32 tim gp en out

lukh:
Hi !
Proposing a fix for the PR: #405 , where it looks like, that the TIM14 actually has a BDTR reg.
Providing the fix where the template doesn't filter out the tim14


salkinium:
Please cherry-pick your fix on top of develop (or I can do it if you don’t know how).

salkinium (approved):

lukh:
Thank you !

PR #456, opened 2020-09-01
Update and simplify installation instructions for linux

rleh:


docs/src/guide/installation.md
```diff
@@ -30,6 +30,40 @@ additional build system documentation.
...
+
+!!! warning "arm-none-eabi-gcc on Ubuntu ≤ 19.10"
+	Ubuntu ≤ 19.10 does not provide a recent arm-none-eabi-gcc toolchain.
+	Install the ARM toochain by downloading [the pre-built version][gcc-arm-toolchain]
+	for 64-bit Linux and adding its `/bin` directory to your path.
+
+!!! bug "avr-gcc on Ubuntu"
+	Ubuntu does not provide an up-to-date version of avr-gcc that supports C++17.
+	For our CI we've created a [precompiled version of avr-gcc for Ubuntu][avr-gcc-latest].
+	Unfortunately its path is hardcoded to `/work/avr-gcc`.
```
salkinium:
```suggestion
	Unfortunately its path is hardcoded to `/work/avr-gcc`, [please help us fix it](https://github.com/modm-ext/docker-avr-gcc/issues/1).
```
too much begging??? 😳

docs/src/guide/installation.md
```diff
@@ -30,6 +30,40 @@ additional build system documentation.
...
+
+		alias scons="/usr/bin/env python3 $(which scons)"
+
+	If you get errors about missing `pyelftools`, check that you're really
+	running SCons with Python 3!
+
+!!! warning "arm-none-eabi-gcc on Ubuntu ≤ 19.10"
+	Ubuntu ≤ 19.10 does not provide a recent arm-none-eabi-gcc toolchain.
+	Install the ARM toochain by downloading [the pre-built version][gcc-arm-toolchain]
+	for 64-bit Linux and adding its `/bin` directory to your path.
```
salkinium: Wouldn't this conflict with the above install instructions? The user first does `sudo apt-get install gcc-arm-none-eabi`, then manually installs this toolchain and now has both toolchains in the path in probably undefined order?

docs/src/guide/installation.md
```diff
@@ -30,6 +30,40 @@ additional build system documentation.
...
+For Ubuntu 20.04LTS, these commands install the basic build system:
+
+	sudo apt-get install python3 python3-pip scons git gcc-arm-none-eabi openocd \
+	  gcc build-essential libboost-all-dev
+	pip3 install modm gdbgui
+
+!!! warning "OpenOCD for recent targets"
+	The latest OpenOCD release v0.10.0 (as of May 2020) is too old for some targets
+	(STM32G0, STM32G4, STM32F7). To program these targets you need to compile the
+	[HEAD version of OpenOCD from source][openocd-source].
```
salkinium: Same issue here, installing first the old openocd, then adding the new openocd to the path.

docs/src/guide/installation.md
```diff
@@ -30,6 +30,40 @@ additional build system documentation.
 Please help us [keep these instructions up-to-date][contribute]!
 
 
+## Linux
+
+For Ubuntu 20.04LTS, these commands install the basic build system:
+
+	sudo apt-get install python3 python3-pip scons git gcc-arm-none-eabi openocd \
+	  gcc build-essential libboost-all-dev
+	pip3 install modm gdbgui
```
salkinium, line 39: What happened to avrdude etc?

salkinium (commented):
The old instructions were structured according to the target, since the user does not need to install avr-gcc for ARM or arm-gcc for avr etc. Your changes now squash everything together, is this a good idea?

salkinium:
Also while we're here, lets change the order of the guide and place *Installation* first: https://github.com/modm-io/modm/blob/develop/docs/mkdocs.yml#L72-L73

salkinium:
> Your changes now squash everything together

Maybe we can add a TL;DR: installation two-liner right at the beginning for the most likely use-case, which is ARM Cortex-M development on Ubuntu 20.04 LTS? Then after that a more detailed view?

docs/src/guide/installation.md
rleh: I would assume the user does not install the outdated GCC version when reading this warning.


> probably undefined order

Depends the path order in his `$PATH`.

salkinium: > Depends the path order in his $PATH.

I meant from our perspective undefined when they open an issue about it.

> I would assume the user does not install the outdated GCC version when reading this warning.

I would assume the user immediately executes first string that contains `sudo apt-get install`, mostly because that's what I would do ;-P

docs/src/guide/installation.md
rleh: AVR Toolchain on Ubuntu is too old (GCC 5, modm requires >7 (?)).

You could leave the installation command for `avrdude`, but again I think someone will be able to do that if they can compile a modern AVR-GCC themselves...

docs/src/guide/installation.md
rleh: If the user gets to this warning by reading it, I think he is competent to remove the OpenOCD package from the apt command.

docs/src/guide/installation.md
rleh: `... and adding its /bin directory to the beginning of your $PATH.` should fix the issue and works in case the user already has installed any arm-none-eabi-gcc.

docs/src/guide/installation.md
rleh: _... compile [OpenOCD], install it and add it to the beginning of your `$PATH`_.
In case the user installed openocd with the command above or openocd was previously installed the up-to-date OpenOCD will be used.

rleh:
> Maybe we can add a TL;DR

Exactly this was my motivation to bring install all packages with a single apt (& one pip) command.

> Then after that a more detailed view?

Done in the warnings.

salkinium (approved):
Ok, let's go with this!

PR #460, opened 2020-09-11
Fix SysTick scaling issue on SAM

### LINKED ISSUE #459: SysTick on SAM has wrong scaling
Just for tracking: Seems like timers are 8x off on SAM due to the CPU not implementing reference clock (see errata).

We therefore need to turn of the prescaler always on SAM.

henrikssn:
Resolves #459

### DIFF
--- a/src/modm/platform/clock/systick/module.lb
+++ b/src/modm/platform/clock/systick/module.lb
@@ -36,9 +36,12 @@ def validate(env):
 def build(env):
     core = env[":target"].get_driver("core")["type"]
     default = 1000 if "m0" in core else 4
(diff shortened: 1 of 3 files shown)


salkinium:
I think this is a better and generic solution, since the reference clock is implementation defined and thus not necessarily AHB/8.
I think the max clock speed is now limited to ~67MHz instead of ~536MHz at least on ARMv7-M architectures, on ARMv6-M it should be fine I think (not entirely sure).

Could you test this on your device? Then I can squash the commits and merge.

henrikssn:
Thanks! LGTM, tested both SAM and STM32.

salkinium (approved):

PR #437, opened 2020-07-19
[board] Add basic support for STM32_F32VE board and [driver] ILI9341 via #464

asmfreak:
This PR add support for [this board](https://stm32-base.org/boards/STM32F407VET6-STM32-F4VE-V2.0.html).

* [X] `board.hpp` actually works (ported from `disco_f407vg` board)
* [X] Examples ported
    * [X] Blink ex
    * [X] Uart works
    * [x] LCD Display works (requires new functionality for ILI9341 in `ParallelDisplay` module).
    * [x] Touch screen works
    * [x] On-board memory unit works

PR #464, opened 2020-09-13
Ili9341 support (SPI only).

mikewolfram:
First version of ILI9341 support. Currently SPI only. For comparison with version of @ASMfreaK.

### DIFF
--- a/src/modm/driver/display/ili9341_spi.hpp
+++ b/src/modm/driver/display/ili9341_spi.hpp
@@ -1,0 +1,320 @@
+/*
+ * Copyright (c) 2019, Mike Wolfram
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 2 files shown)


asmfreak:
I'd like to work on this implementation to add parallel bus interface, but I'm not sure about the copyright notice on the files.

mikewolfram:
> I'd like to work on this implementation to add parallel bus interface, but I'm not sure about the copyright notice on the files.

Sorry, happens to me all the time. Fixed the copyright.

mikewolfram:
BTW, I have also code for Ili9486, which is basically the same like the Ili9341 (few registers different only). That code uses 16 bits parallel interface. If you want I can open another PR with that code.

For the parallel interface I was thinking to implement DMA support as well.

salkinium:
This PR has been merged into #437 which also adds a parallel interface to the driver. Thanks!

PR #466, opened 2020-09-14
First implementation of Ethernet for STM32F7 (and later STM32F4).

mikewolfram:
Et voila, here we go. This is the code I currently use plus an example for F767ZI Nucleo board. The example is a quick hack of code I already use.

Quite a few things to do, e.g. the module.lb file is missing. Also the network interface file for FreeRTOS might need a better location. And in it there is hardcoded LAN8720 PHY. I was thinking to make that an option in the XML or required the use to create an include file with the proper declaration.

Let's start from here, any help is appreciated. I'm a bit short of time until the end of the year, but together we should be able to get it done.

### DIFF
--- a/examples/nucleo_f767zi/ethernet/main.cpp
+++ b/examples/nucleo_f767zi/ethernet/main.cpp
@@ -1,0 +1,256 @@
+/*
+ * Copyright (c) 2020, Mike Wolfram
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/driver/ethernet/lan8720a.hpp>
+#include <modm/processing/rtos.hpp>
+#include <FreeRTOS_IP.h>
+#include <FreeRTOS_Sockets.h>
+
+using namespace Board;
+
+namespace Ethernet
+{
+	using RMII_Ref_Clk = GpioInputA1;
+	using RMII_Mdio = GpioA2;
+	using RMII_Crs_Dv = GpioInputA7;
+	using RMII_Tx_En = GpioOutputG11;
+	using RMII_Tx_D0 = GpioOutputG13;
+	using RMII_Tx_D1 = GpioOutputB13;
+	using RMII_Mdc = GpioOutputC1;
+	using RMII_Rx_D0 = GpioInputC4;
+	using RMII_Rx_D1 = GpioInputC5;
+	using Port = Eth<modm::Lan8720a>;
+}
+
+UBaseType_t ulNextRand;
+
+void vApplicationIPNetworkEventHook(eIPCallbackEvent_t eNetworkEvent);
+
+class NetworkInitTask : modm::rtos::Thread
+{
+public:
+	NetworkInitTask()
+	: Thread(5, 2048, "network_init")
+	{}
+
+	void
+	run()
+	{
+		uint8_t ipAddress[4] { 192, 168, 1, 1 };
+		uint8_t netmask[4] { 255, 255, 255, 0 };
+		uint8_t gatewayAddress[4] { 0, 0, 0, 0 };
+		uint8_t dnsAddress[4] { 0, 0, 0, 0 };
+
+		// local MAC address
+		uint8_t macAddress[] { 0x02, 0x00, 0x00, 0x00, 0x00, 0x00 };
+
+		// A real MAC address can be retrieved from the Microchip 24AA02E48
+		// I2C EEPROM, which is locaed at address 0xFA.
+
+		// initialize random numbers
+	    time_t now;
+	    time(&now);
+	    ulNextRand = uint32_t(now);
+
+	    FreeRTOS_IPInit(ipAddress,
+	    		netmask,
+				gatewayAddress,
+				dnsAddress,
+				&macAddress[0]);
+
+	    vTaskDelete(0);
+	}
+};
+
+class HttpConnection
+{
+	static constexpr TickType_t shutdownTimeout { pdMS_TO_TICKS(5000) };
+	static constexpr TickType_t receiveTimeout { pdMS_TO_TICKS(5000) };
+	static constexpr TickType_t sendTimeout { pdMS_TO_TICKS(5000) };
+
+public:
+	static constexpr char name[] { "HTTPConnection" };
+	static constexpr uint8_t httpText[] = {
+			"HTTP/1.1 200 OK \r\n"
+			"Content-Type: text/html\r\n"
+			"Connection: keep-alive\r\n"
+			"\r\n"
+			"<html><body><h1>Hello from your STM32!</h1></body></html>"
+	};
+
+	enum class
+	ResponseStatus : uint16_t {
+		Ok = 200,
+		BadRequest = 400,
+		NotFound = 404,
+	};
+
+	static void
+	run(void *parameter)
+	{
+		Socket_t connectedSocket = reinterpret_cast<Socket_t>(parameter);
+		uint8_t *buffer = reinterpret_cast<uint8_t *>(pvPortMalloc(ipconfigTCP_MSS));
+
+		if (buffer) {
+			FreeRTOS_setsockopt(connectedSocket, 0, FREERTOS_SO_RCVTIMEO, &receiveTimeout,
+					sizeof(receiveTimeout));
+			FreeRTOS_setsockopt(connectedSocket, 0, FREERTOS_SO_SNDTIMEO, &sendTimeout,
+					sizeof(sendTimeout));
+
+			while (true) {
+				std::memset(buffer, 0, ipconfigTCP_MSS);
+				int32_t bytes = FreeRTOS_recv(connectedSocket, buffer, ipconfigTCP_MSS, 0);
+				if (bytes <= 0)
+					break;
+				if (FreeRTOS_send(connectedSocket, httpText, sizeof(httpText) - 1, 0) < 0)
+					break;
+			}
+		}
+
+		FreeRTOS_shutdown(connectedSocket, FREERTOS_SHUT_RDWR);
+		TickType_t shutdownTime { xTaskGetTickCount() };
+		do {
+			if (FreeRTOS_recv(connectedSocket, buffer, ipconfigTCP_MSS, 0) < 0)
+				break;
+		} while ((xTaskGetTickCount() - shutdownTime) < shutdownTimeout);
+
+		vPortFree(buffer);
+		FreeRTOS_closesocket(connectedSocket);
+		vTaskDelete(0);
+
+	}
+};
+
+class HttpServerListener
+{
+	static constexpr TickType_t receiveTimeout { portMAX_DELAY };
+	static constexpr BaseType_t backlog { 20 };
+
+public:
+	static constexpr char name[] { "HTTPListener" };
+
+	static void
+	run(void *)
+	{
+		Socket_t listeningSocket;
+		Socket_t connectedSocket;
+
+		listeningSocket = FreeRTOS_socket(FREERTOS_AF_INET,
+				FREERTOS_SOCK_STREAM,
+				FREERTOS_IPPROTO_TCP);
+		FreeRTOS_setsockopt(listeningSocket, 0, FREERTOS_SO_RCVTIMEO, &receiveTimeout,
+				sizeof(receiveTimeout));
+
+#if ipconfigUSE_TCP_WIN == 1
+		WinProperties_t winProps {
+			.lTxBufSize = ipconfigTCP_TX_BUFFER_LENGTH,
+			.lTxWinSize = 2,
+			.lRxBufSize = ipconfigTCP_RX_BUFFER_LENGTH,
+			.lRxWinSize = 2,
+		};
+		FreeRTOS_setsockopt(listeningSocket, 0, FREERTOS_SO_WIN_PROPERTIES,
+				reinterpret_cast<void *>(&winProps), sizeof(winProps));
+#endif
+
+		struct freertos_sockaddr bindAddress {
+			.sin_port = FreeRTOS_htons(80),
+		};
+		FreeRTOS_bind(listeningSocket, &bindAddress, sizeof(bindAddress));
+		FreeRTOS_listen(listeningSocket, backlog);
+
+		struct freertos_sockaddr clientAddress;
+
+		while (true) {
+			connectedSocket = FreeRTOS_accept(listeningSocket, &clientAddress, 0);
+			char buffer[16];
+			FreeRTOS_inet_ntoa(clientAddress.sin_addr, buffer);
+			xTaskCreate(HttpConnection::run, HttpConnection::name, configMINIMAL_STACK_SIZE * 5,
+					reinterpret_cast<void *>(connectedSocket), configMAX_PRIORITIES, 0);
+		}
+	}
+};
+
+NetworkInitTask networkInit;
+
+int
+main()
+{
+	Board::initialize();
+	Leds::setOutput();
+	MODM_LOG_INFO << "\n\nReboot: Ethernet Example" << modm::endl;
+
+	Ethernet::Port::connect<Ethernet::RMII_Ref_Clk::Refclk,
+		Ethernet::RMII_Mdc::Mdc,
+		Ethernet::RMII_Mdio::Mdio,
+		Ethernet::RMII_Crs_Dv::Rcccrsdv,
+		Ethernet::RMII_Tx_En::Txen,
+		Ethernet::RMII_Tx_D0::Txd0,
+		Ethernet::RMII_Tx_D1::Txd1,
+		Ethernet::RMII_Rx_D0::Rxd0,
+		Ethernet::RMII_Rx_D1::Rxd1>();
+
+    modm::rtos::Scheduler::schedule();
+
+	// we should never get here
+	return 0;
+}
+
+void vApplicationIPNetworkEventHook(eIPCallbackEvent_t eNetworkEvent)
+{
+	static bool taskCreated = false;
+
+	if (eNetworkEvent != eNetworkUp)
+		return;
+
+	if (not taskCreated) {
+		xTaskCreate(HttpServerListener::run, HttpServerListener::name, configMINIMAL_STACK_SIZE * 2, 0, configMAX_PRIORITIES + 1, 0);
+		taskCreated = true;
+	}
+
+	uint32_t ipAddress;
+	uint32_t netmask;
+	uint32_t gateway;
+	uint32_t dns;
+	char buffer[16];
+
+	FreeRTOS_GetAddressConfiguration(&ipAddress, &netmask, &gateway, &dns);
+	FreeRTOS_inet_ntoa(ipAddress, buffer);
+	MODM_LOG_DEBUG << "IP address: " << buffer << modm::endl;
+	FreeRTOS_inet_ntoa(netmask, buffer);
+	MODM_LOG_DEBUG << "Netmask   : " << buffer << modm::endl;
+	FreeRTOS_inet_ntoa(gateway, buffer);
+	MODM_LOG_DEBUG << "Gateway   : " << buffer << modm::endl;
+	FreeRTOS_inet_ntoa(dns, buffer);
+	MODM_LOG_DEBUG << "DNS       : " << buffer << modm::endl;
+}
+
+UBaseType_t uxRand( void )
+{
+	static constexpr uint32_t ulMultiplier = 0x015a4e35UL;
+	static constexpr uint32_t ulIncrement = 1UL;
+
+	/* Utility function to generate a pseudo random number. */
+
+	ulNextRand = ( ulMultiplier * ulNextRand ) + ulIncrement;
+	return( ( int ) ( ulNextRand >> 16UL ) & 0x7fffUL );
+}
+
+BaseType_t xApplicationGetRandomNumber(uint32_t* pulNumber)
+{
+	*(pulNumber) = uxRand();
+	return pdTRUE;
+}
+
+uint32_t ulApplicationGetNextSequenceNumber(uint32_t, uint16_t, uint32_t, uint16_t)
+{
+	return uxRand();
+}
--- a/ext/aws/module.lb
+++ b/ext/aws/module.lb
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2016-2017, Niklas Hauser
+# Copyright (c) 2016-2017, 2021, Niklas Hauser
 # Copyright (c) 2017, Fabian Greif
 #
 # This file is part of the modm project.
@@ -10,6 +10,23 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # -----------------------------------------------------------------------------
+
+class FreeRTOS_TCP_LAN8720A(Module):
+    def init(self, module):
+        module.name = "lan8720a"
+        module.description = """
+# a:FreeRTOS-Plus-TCP Ethernet via LAN8720A
+
+This module implements TCP over Ethernet via the LAN8720A transceiver.
+"""
+    def prepare(self, module, options):
+        module.depends(":platform:eth", ":driver:lan8720a")
+        return options[":target"].has_driver("eth:stm32*")
+
+    def build(self, env):
+        env.outbasepath = "modm/ext/freertos_plus_tcp"
+        env.copy("modm_lan8720a.cpp")
+
 
 class FreeRTOS_TCP(Module):
     def init(self, module):
@@ -29,6 +46,8 @@ the config macros, then redefine them with your options, for example:
 """
 
     def prepare(self, module, options):
+        module.add_submodule(FreeRTOS_TCP_LAN8720A())
+        module.depends(":platform:heap")
         return True
 
     def build(self, env):
--- a/src/modm/board/nucleo_f767zi/board.hpp
+++ b/src/modm/board/nucleo_f767zi/board.hpp
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) 2020, Mike Wolfram
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_STM32_NUCLEO_F767ZI_HPP
+#define MODM_STM32_NUCLEO_F767ZI_HPP
+
+#include <modm/platform.hpp>
+#include <modm/architecture/interface/clock.hpp>
+#include <modm/debug/logger.hpp>
+/// @ingroup modm_board_nucleo_f767zi
+#define MODM_BOARD_HAS_LOGGER
+
+using namespace modm::platform;
+
+/// @ingroup modm_board_nucleo_f767zi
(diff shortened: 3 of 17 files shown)


mikewolfram:
Wanted to make it a draft PR, any way to change it? Or delete and recreate?

strongly-typed:
```
ERROR: Cannot resolve name 'modm:platform:eth' of type Module!
```

Could you please also make your modm module available? Thank you!

mikewolfram:
> Could you please also make your modm module available? Thank you!

As of now I don't have a module file. Before opening the PR I had those files in my sources, I copied them to that location for the PR. So even with a module file I expect them to not compile as the includes etc. won't work.

Sorry for that. Right now I'm hacking together (is this real English? :-) ) a proof of concept while throwing away whatever I learned about software development, coding style etc.

Working 11-12 hours a day, 11 days in a row and living in a hotel isn't so much fun. :-( Let's see what I could for you...

mikewolfram:
And without a separate project I cannot event run "lbuild" as it would overwrite changes I made to some files. :-( I know, I completely messed it up.

salkinium:
Please don't burn yourself out!

(Sascha is also capable of adding a suitable `module.lb` himself \*wags finger like a mom\*)

mikewolfram:
> Please don't burn yourself out!

I'm already beyond a burnout... :-) 
> (Sascha is also capable of adding a suitable `module.lb` himself *wags finger like a mom*)

Sure he can, but he's afraid to destroy my genius work. :-)
(Sorry, couldn't resist).

strongly-typed:
Take your time! I am just curious in bringing Ethernet support to modm. I have no deadlines nor customers involved with this topic. I have a Nucleo F429ZI (which claims to have Ethernet) for further testing and maybe I can help with your current memory problems.

mikewolfram:
I have a few days off and won't be touching any keyboard...

The F4 series might be a bit different to the F7 in regards to ethernet. I used the FreeRTOS network interface for STM32 HAL as example and stripped off things. There are a few places where they differ between F4 and F7. They also provide a modified version of the HAL_ETH files. Maybe you want to have a look to get an idea what is the actual difference.

More to come next week.

mikewolfram:
I have updated with the latest from develop, don't know why all the commits are showing up here now. Did I screw something up?

Still investigating the memory issue. I have a feeling, that the stack of the IPTask might get corrupted. In my driver I send an event with the network descriptor pointer, which is always in the range 0x20013000 to 0x20015000. Had a hard fault, where the received event had a pointer of 0x2002ba24. Increased the stack from 640 to 1024 words, let's see if that changes anything.

The other tasks in theory should not be able to modify the IPTask stack, but you never know. I have a very simple task that only toggles a LED every second, even that one needs more than the minimal 128 words for the stack. Set it to 256 words and it is now running.

mikewolfram:
And another question. I saw that you moved a few files in FreeRTOS_TCP. When I added it I had a rather simple directory layout, but maybe it would make sense to move the headers into a freertos_tcp folder like freertos?

salkinium:
Delegating to @strongly-typed.

mikewolfram:
Short question, one example is failing, because it is complaining about a symbol defined in stm32f765xx.h. I thought this would be included before? Or did I miss something?

src/modm/platform/eth/stm32/module.lb
```diff
@@ -0,0 +1,41 @@
...
+
+def init(module):
+    module.name = ":platform:eth"
+    module.description = "Ethernet of STM32F4/F7"
+
+def prepare(module, options):
+    device = options[":target"]
+
+    # FIXME the driver is for L4 only
+    if device.identifier["family"] not in ["f7"]:
```
salkinium, line 24: Do all F7's have Ethernet? You could additionally check the device files with `device.has_driver("eth:stm32*")` or similiar.

salkinium: You can run the test locally with `(cd test/all; python3 run_all.py stm32f7)`. That's faster than waiting for the CI, since it remembers the failed ones and retries only those on the next call.

strongly-typed:
> And another question. I saw that you moved a few files in FreeRTOS_TCP. When I added it I had a rather simple directory layout, but maybe it would make sense to move the headers into a freertos_tcp folder like freertos?

I had duplicated includes so I only copied the needed files. Yes, a subdirectory would make perfect sense.

(I do not know if there are any dependencies at FreeRTOS that require the files to be in one directory.)

mikewolfram:
Looks like the code is running fine now. Maybe the fix to drop invalid frames was the issue? That would explain why it was working fine on an isolated network, but crashed on a busy network in the office?

mikewolfram:
I wasn't aware this is still a draft and was wondering why we didn't get any further. I turned the status to "Ready for review", so let's discuss what is needed to finally merge it (e.g. lack of documentation).

We have the code running for several months now, with no major problems (still don't know if this is a bug in the implementation or the result of some crazy network setup in the company).

rleh:
I don't know what happened with the MacOS CI, but it is now fixed (by re-running the job).
I will do a review later today!

src/modm/platform/eth/stm32/module.lb
mikewolfram: Added check for driver.

src/modm/platform/eth/stm32/eth.hpp
```diff
@@ -0,0 +1,705 @@
...
+
+//		Ethernet::RMII_Ref_Clk::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Mdc::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Mdio::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Crs_Dv::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Tx_En::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Tx_D0::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Tx_D1::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Rx_D0::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
+//		Ethernet::RMII_Rx_D1::configure(modm::platform::Gpio::OutputType::PushPull, modm::platform::Gpio::OutputSpeed::VeryHigh);
```
rleh: Why is this code commented out?
Are the GPIOs initialized somewhere else?

mikewolfram: Yes, this is done by the configurator and connector (see line 382). Need to remove the lines.

rleh (approved):
I made a few minimal changes in 10a4cff0f (remove commented-out code and inline some static variables), otherwise looks very good!

rleh:
@mikewolfram Could you squash the commits into ~two or three~ four commits?

mikewolfram:
> @mikewolfram Could you squash the commits into two or three commits?

How about 4? :-)

src/modm/board/nucleo_f767zi/board.hpp
```diff
@@ -0,0 +1,150 @@
...
+		Rcc::enableExternalClock(); // 8 MHz
+		const Rcc::PllFactors pllFactors{
+			.pllM = 4,		// 8MHz / M=4   -> 2MHz
+			.pllN = 216,	// 2MHz * N=216 -> 432MHz
+			.pllP = 2		// 432MHz / P=2 -> 216MHz = F_cpu
+			//.pllQ = 9		// 432MHz / P=2 -> 48MHz (USB, etc.)
+		};
+		Rcc::enablePll(Rcc::PllSource::ExternalClock, pllFactors);
+		PWR->CR1 |= PWR_CR1_ODEN; // Enable overdrive mode
+		while (not (PWR->CSR1 & PWR_CSR1_ODRDY)) ;
```
salkinium:
```suggestion
		Rcc::enableOverdrive();
```

examples/nucleo_f767zi/ethernet/main.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+	Board::initialize();
+	Leds::setOutput();
+
+	// Use the logging streams to print some messages.
+	// Change MODM_LOG_LEVEL above to enable or disable these messages
+	MODM_LOG_DEBUG   << "debug"   << modm::endl;
+	MODM_LOG_INFO    << "info"    << modm::endl;
+	MODM_LOG_WARNING << "warning" << modm::endl;
+	MODM_LOG_ERROR   << "error"   << modm::endl;
```
salkinium: Add something like "Reboot" instead of this.

examples/nucleo_f767zi/ethernet/main.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+	FreeRTOS_GetAddressConfiguration(&ipAddress, &netmask, &gateway, &dns);
+	FreeRTOS_inet_ntoa(ipAddress, buffer);
+	modm::log::debug << "IP address: " << buffer << modm::endl;
+	FreeRTOS_inet_ntoa(netmask, buffer);
+	modm::log::debug << "Netmask   : " << buffer << modm::endl;
+	FreeRTOS_inet_ntoa(gateway, buffer);
+	modm::log::debug << "Gateway   : " << buffer << modm::endl;
+	FreeRTOS_inet_ntoa(dns, buffer);
+	modm::log::debug << "DNS       : " << buffer << modm::endl;
```
salkinium: Use the macros for consistency.

salkinium (changes requested):
There are issues with the dependencies, in particular the `:platform:eth` module should not implement or depend on anything from  FreeRTOS. Instead the `:freertos:tcp` should depend on `:platform:eth` and also contain the STM32 port. This is also how TinyUSB is integrated with STM32.
This way other communication protocols (for example using raw ethernet frames) can be implemented without pulling in FreeRTOS etc.

Since this is a rather esoterical change, I'll push some fixup commits onto this branch myself ;-P

src/modm/board/nucleo_f767zi/board.hpp
mikewolfram: Done.

examples/nucleo_f767zi/ethernet/main.cpp
mikewolfram: Done.

examples/nucleo_f767zi/ethernet/main.cpp
mikewolfram: ?

mikewolfram:
> There are issues with the dependencies, in particular the `:platform:eth` module should not implement or depend on anything from FreeRTOS. Instead the `:freertos:tcp` should depend on `:platform:eth` and also contain the STM32 port. This is also how TinyUSB is integrated with STM32.
> This way other communication protocols (for example using raw ethernet frames) can be implemented without pulling in FreeRTOS etc.

You are right. So the `freertos_networkinterface.cpp` should be be moved somewhere else and the `freertos:tcp` would pull it in if we are building for STM32? In the `:freertos:tcp` I would check for the build type and add the dependency on `:platform:eth`?
 
> Since this is a rather esoterical change, I'll push some fixup commits onto this branch myself ;-P

Go ahead, any improvement is welcome.

Would also be great if somebody had a Nuclo F767 board and could do a test. I'm on a custom hardware and have no access to the Nucleo board at the moment.

salkinium:
I've added my changes. The last commit just the diff between your and [my branch here](https://github.com/salkinium/modm/commits/feature/ethernet-stm32), so that I don't overwrite any work. If you're happy with the results I can force push my branch onto yours. (I still need to fix git authors)

src/modm/platform/eth/stm32/eth.hpp
```diff
@@ -0,0 +1,695 @@
...
+			(void) readPhyRegister(PHY::Register::BSR, phy_register);
+			if ((phy_register & PHY::AutoNegotiationComplete) == PHY::AutoNegotiationComplete)
+				break;
+			modm::delay_ms(1);
+		} while (timeout-- > 0);
+		if (timeout <= 0)
+			return false;
+
+		// read auto-negotiation result
+		if (not readPhyRegister(PHY::Register::SR, phy_register))
```
chris-durand, line 599: This `SR` register is not part of the MII register specification, this is vendor specific. Thus, the STM32 ethernet driver depends on some implementation detail of the Microchip LAN8720A phy. We should try to come up with an abstraction that allows us to encapsulate phy specific details.
As far as I can see, reading the negotiated speed and duplex mode is the only vendor specific phy functionality used in this driver. It might be sufficient to define in the `PHY` class in which vendor specific register the driver can find that information. Then the driver passes the content to a function in the `PHY` class to evaluate the contents.

salkinium:
> So the `freertos_networkinterface.cpp` should be be moved somewhere else

It's now a submodule `:freertos:tcp:lan8720a`, which depends on `:platform:eth` and `:driver:lan8720a`. Not sure if the latter deserves it's own driver module, but at least it's consistent this way. In future other network interfaces can be added as submodules of `:freertos:tcp` perhaps even multiple at the same time if their resources don't conflict.

ext/aws/module.lb
```diff
@@ -11,6 +11,23 @@
...
+
+This module implements TCP over Ethernet via the LAN8720A transceiver.
+"""
+    def prepare(self, module, options):
+        module.depends(":platform:eth", ":driver:lan8720a")
+        return options[":target"].identifier.platform in ["stm32"]
+
+    def build(self, env):
+        env.outbasepath = "modm/ext"
+        env.copy("modm_lan8720a.cpp", "freertos/FreeRTOS-Plus-TCP/modm_lan8720a.cpp")
```
salkinium:
```suggestion
        env.copy("modm_lan8720a.cpp", "freertos_plus_tcp/modm_lan8720a.cpp")
```

src/modm/driver/ethernet/lan8720a.hpp
```diff
@@ -0,0 +1,54 @@
...
+	static constexpr int WriteTimeout = 0xffff;
+
+	static constexpr uint16_t Reset = 0x8000;
+	static constexpr uint16_t LoopBack = 0x4000;
+	static constexpr uint16_t FullDuplex100M = 0x2100;
+	static constexpr uint16_t HalfDuplex100M = 0x2000;
+	static constexpr uint16_t FullDuplex10M = 0x0100;
+	static constexpr uint16_t HalfDuplex10M = 0x0000;
+	static constexpr uint16_t AutoNegotiation = 0x1000;
+	static constexpr uint16_t RestartAutoNegotiation = 0x0200;
```
chris-durand, line 43: This should be scoped somehow. It is not immediately obvious these constants refer to the `BCR` register. There are bits with the same names in other registers with different values, e.g. `FullDuplex100M` in `AN`. That's probably why the driver code above uses hard-coded magic values.

Couldn't we use modm register abstractions here?

src/modm/driver/ethernet/lan8720a.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+	static constexpr uint16_t LinkedStatus = 0x0004;
+	static constexpr uint16_t DuplexStatus = 0x0010;
+	static constexpr uint16_t SpeedStatus = 0x0004;
+	static constexpr uint16_t AutoNegotiationComplete = 0x0020;
+	static constexpr uint16_t JabberDetection = 0x0002;
```
chris-durand, line 49: Same as above.

src/modm/platform/eth/stm32/eth.hpp
```diff
@@ -0,0 +1,694 @@
...
+				break;
+			modm::delay_ms(1);
+		} while (timeout-- > 0);
+		if (timeout <= 0) {
+			linkStatus = LinkStatus::Down;
+			return false;
+		}
+
+		// configure auto negotiation
+		phy_register = 0x100 // 100 FD
```
chris-durand, line 453: These values should be defined somewhere and not hard-coded.

src/modm/platform/eth/stm32/eth_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+Eth<PHY>::configureMacAddresses()
+{
+	static constexpr uint32_t ETH_MAC_ADDR_HBASE { ETH_MAC_BASE + 0x40 };  /* Ethernet MAC address high offset */
+	static constexpr uint32_t ETH_MAC_ADDR_LBASE { ETH_MAC_BASE + 0x44 };  /* Ethernet MAC address low offset */
+
+	static constexpr MacAddress zeroMac { 0 };
+
+	for (std::size_t i = 0; i < macAddresses.size(); ++i) {
+		auto const &macAddress = macAddresses[i];
+		if (std::memcmp(zeroMac.data(), macAddress.data(), macAddress.size()) == 0)
```
chris-durand, line 80: This could be `==` instead of `memcmp` since we have a `std::array`. Code-gen will also be `memcmp` in that case.

salkinium:
Before you commit on top of this branch, please let me force push my commits structure, so that I don't need to back-commit all the next changes.

Alternatively **you** can commit on top of my branch and force push your branch yourself.

mikewolfram:
> Before you commit on top of this branch, please let me force push my commits structure, so that I don't need to back-commit all the next changes.
> 
> Alternatively **you** can commit on top of my branch and force push your branch yourself.

Just do it, I won't be able to continue tonight. Thanks for help.

rleh:
All Nucleo-144 boards actually have a LAN8742A PHY chip, looks compatible to LAN8720 at first glance.
@mikewolfram Do you actually use the LAN8720 PHY or can you say something about exchangebility of the two PHYs?

I just ordered a Nucleo-F767 board, should arrive on ~tuesday~ thursday; then I will be able to test.

rleh:
I approved the PR and would have merged it as it was to get basic Ethernet support into modm, despite of some hardcoded non-generic values and all code in one lbuild module.
Would have been easy to improve this later in another PR, maybe together with adding support for STM32F4...

I have the impression that sometimes it would help not to complicate pull requests if possible but to merge them faster instead.

chris-durand:
Just to clarify, I am not against merging it soon. It is better to merge it than having it go stale. But that does not mean we shouldn't document findings from review and potential TODOs for future improvement. If we transfer remaining open points to an issue, it could potentially help whoever wants to continue and port it to different hardware. That's why I added comments and not requests for change.

salkinium (approved):

salkinium:
I'm both in favor of faster merging and thorough review :-P

Looks like I broke docs though… I'll fix it tomorrow.

mikewolfram:
> All Nucleo-144 boards actually have a LAN8742A PHY chip, looks compatible to LAN8720 at first glance.
> @mikewolfram Do you actually use the LAN8720 PHY or can you say something about exchangebility of the two PHYs?
> 
> I just ordered a Nucleo-F767 board, should arrive on tuesday; then I will be able to test.

The LAN8742A is can be used instead of LAN8720A, same pins and registers. LAN8742A has few more features like power saving. For our project we designed to use this one, but in the BOM I made a mistake and put in the order number of the LAN8720A. Since we don't care about power saving etc. this wasn't a problem.

rleh:
I can confirm, that
- [x] the `blink/`-example works on hardware
- [ ] the `ethernet/`-example does not work. 😒 

The logger just outputs `Reboot: Ethernet Example` and then nothing happens. The LEDs on the ethernet jack light up once I plug in a cable.

mikewolfram:
Thanks for testing. Let's find out what is different to my custom board and what might be wrong.

Did you check with the debugger what is causing the reboot? Maybe the stack needs to be increased?

rleh:
For some reason GDB does not work on my computer, i'm working on it...

Compiling in debug mode (`scons profile=debug`) tells me the following:
```
Reboot: Ethernet Example
Assertion 'eth.init' @ 0x00000001 (1) failed!
  ETH::initialize() timed out on software reset!
Abandoning...
```

src/modm/platform/eth/stm32/eth.hpp
```diff
@@ -0,0 +1,696 @@
...
+		int timeout = 1'000; // max 1ms
+		while ((DmaBusMode(ETH->DMABMR) & DmaBusMode_t(DmaBusMode::SoftwareReset)) and (timeout-- > 0)) {
+			// Wait until the PHY has reset.
+			modm::delay_us(1);
+
+			// NOTE: If the program hangs here check MII/RMII value.
+		}
+		if (not modm_assert_continue_fail_debug(timeout > 0, "eth.init",
+				"ETH::initialize() timed out on software reset!", 1))
+			return false;
```
rleh, line 407: Related to this code...

Are there any jumpers on the Nucleo-F767ZI board that have to be changed for Ethernet to work?

mikewolfram: Let me check the manual...

rleh: All jumper and solder bridges are set like described in the [Nucleo-144 manual](https://www.st.com/resource/en/user_manual/dm00244518-stm32-nucleo144-boards-mb1137-stmicroelectronics.pdf):

grafik

mikewolfram: I can't find the configuration of the pins. In my code I have this in board.hpp:
```c++
	namespace Ethernet
	{
		using RMII_Ref_Clk = modm::platform::GpioInputA1;
		using RMII_Mdio = modm::platform::GpioA2;
		using RMII_Crs_Dv = modm::platform::GpioInputA7;
		using RMII_Tx_En = modm::platform::GpioOutputB11;
		using RMII_Tx_D0 = modm::platform::GpioOutputB12;
		using RMII_Tx_D1 = modm::platform::GpioOutputB13;
		using RMII_Mdc = modm::platform::GpioOutputC1;
		using RMII_Rx_D0 = modm::platform::GpioInputC4;
		using RMII_Rx_D1 = modm::platform::GpioInputC5;
		using Port = modm::platform::Eth<modm::Lan8720a>;
	}
```
and this call in Board::initialize():
```c++
		Ethernet::Port::connect<Ethernet::RMII_Ref_Clk::Refclk, Ethernet::RMII_Mdc::Mdc,
				Ethernet::RMII_Mdio::Mdio, Ethernet::RMII_Crs_Dv::Rcccrsdv,
				Ethernet::RMII_Tx_En::Txen,
				Ethernet::RMII_Tx_D0::Txd0, Ethernet::RMII_Tx_D1::Txd1,
				Ethernet::RMII_Rx_D0::Rxd0, Ethernet::RMII_Rx_D1::Rxd1>();
```

rleh: SB13 is soldered on, but this should not cause the issue?!

mikewolfram: At least it starts and prints the IP:
```
Reboot: Ethernet Example
IP address: 192.168.1.1
Netmask   : 255.255.255.0
Gateway   : 0.0.0.0
DNS       : 0.0.0.0
```

rleh:
```
Reboot: Ethernet Example
IP address: 192.168.1.1
Netmask   : 255.255.255.0
Gateway   : 0.0.0.0
DNS       : 0.0.0.0
```

... and now it works!

mikewolfram: Ping and www are good?

rleh (approved):
Successfully tested on hardware.

Thanks!!

rleh:
> Ping and www are good?

No. 😒 
I was a bit hasty and accidentally pinged another device on the network.

rleh:
Code runs until
```c++
while (true) {
  connectedSocket = FreeRTOS_accept(listeningSocket, &clientAddress, 0);
//...
```
but never seems to receive a connection.

I'll look at this further later.

kikass13:
ping and http work for me on a nucleo F767ZI, after correcting two pins for the phy.

 * change https://github.com/modm-io/modm/pull/466/files#diff-06b49187e6be0cc40e1894bf66bfb208da0a12b739e4d973f72e4a26bdf74ccfR25-R26
```
using RMII_Tx_En = GpioOutputB11;
using RMII_Tx_D0 = GpioOutputB12;
```
 * to 
 ```
using RMII_Tx_En = GpioOutputG11;
using RMII_Tx_D0 = GpioOutputG13;
```

**Comment by rleh** [comment]
I applied @kikass13 fix, syncronized the docs and rebased onto current develop.
Thanks @kikass13 for finding the error!

@salkinium Are you ok with merging (after CI has finished)?

**Review by rleh — APPROVED** [review]
Tested: Ping and HTTP work ✔️

**Comment by salkinium** [comment]
Yes!

**Review by salkinium — APPROVED** [review]
🚢 it

**Comment by salkinium** [comment]
Something in this PR broke the docs generator. \*wags finger\*
```
Generating documentation for stm32f207zgt7 ...
Error generating documentation for device stm32f207zgt7: 'NoneType' object is not subscriptable
```

**Comment by kikass13** [comment]
> Something in this PR broke the docs generator. *wags finger*
> 
> ```
> Generating documentation for stm32f207zgt7 ...
> Error generating documentation for device stm32f207zgt7: 'NoneType' object is not subscriptable
> ```

someone was using a list iterator on an object without checking for None ... SHAME! :D

**Comment by salkinium** [comment]
I think… \*takes a long bong hit\*… we need a CI for our CI test coverage…

**Comment by rleh** [comment]
> Something in this PR broke the docs generator. *wags finger*
> `Generating documentation for stm32f207zgt7 ...`

... but 15 times, spread random across multiple STM32 sub-families.

And the test run of the [docs.modm.io](https://docs.modm.io) script [was successful](https://circleci.com/gh/modm-io/modm/26554?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). Maybe related to #553?

**Comment by salkinium** [comment]
It's probably something stupid, I'll check it out later.

**Comment by rleh** [comment]
> It's probably something stupid, I'll check it out later.

Did you find out anything?
I just restarted the last successful build on CircleCI (without ethernet merged) to fix the 404 errors on https://docs.modm.io/...

**Comment by salkinium** [comment]
I forgot, but it was something very stupid, so I found it very quickly: #565

File: examples/nucleo_f767zi/ethernet/main.cpp [hunk 13]
```diff
@@ -0,0 +1,256 @@
...
+		// local MAC address
+		uint8_t macAddress[] { 0x02, 0x00, 0x00, 0x00, 0x00, 0x00 };
+
+		// A real MAC address can be retrieved from the Microchip 24AA02E48
+		// I2C EEPROM, which is locaed at address 0xFA.
+
+		// initialize random numbers
+	    time_t now;
+	    time(&now);
+	    ulNextRand = uint32_t(now);
```
rleh, line 62: We could have used real random numbers here...

```suggestion
		// initialize random numbers
	    modm::platform::RandomNumberGenerator::enable();
	    while(!modm::platform::RandomNumberGenerator::isReady()) {}
	    ulNextRand = modm::platform::RandomNumberGenerator::getValue();
```

salkinium: Yes, and as a fallback you can seed srand with the device's UID.

salkinium: Also `time()` is not implemented. We could perhaps wire it to `modm::Clock`, but it's not done yet.

rleh: > Yes, and as a fallback you can seed srand with the device's UID.

I am not sure if this is a good idea or a serious security flaw...

salkinium: Well, you need to collect entropy somehow. There are a bunch more hardware calibration registers that provide entropy within some bounds. It's a somewhat good idea since the values will be different for every device (assuming a good hash function), however, only if the attacker cannot read them out remotely. However, security is definitely beyond the scope of modm and too much work to implement yourself. We'd have to integrate some embedded TLS library for encryption.

PR #470, opened 2020-09-16
[ext] Update modm-devices and stm32 submodules

### LINKED ISSUE #469: Remapped pins on STM32F0/G0 not supported
I have a board with an STM32F042F6 in a 20 Pin TSSOP package and would like to be able to use the remapped pins PA11/PA12. Some F0/G0 contain an internal mux which allows to select whether PA9/PA10 or PA11/PA12 are connected to the package pins. The F0 remaps both in a group whereas they can be mapped individually with the G0. The GPIO files for the remapped pins are currently not generated and the information is [currently not available](https://github.com/modm-io/modm-devices/issues/14) in modm-devices. The develop branch has the package pins, but it doesn't work with the current modm develop.

@salkinium What would it take to make this information available in modm? I would like to add support for remapped pins.

salkinium:
A lot of new STM32 devices have been added, let's see what the damage is.

- [x] Fix duplicate remap pin generation
- [x] Fix memory section renaming

Also fixes #469 

cc @chris-durand

### DIFF
--- a/repo.lb
+++ b/repo.lb
@@ -48,7 +48,7 @@ def check_submodules():
               "    git submodule update --init\n")
 
 # Import modm-device tools
-sys.path.append(repopath("ext/modm-devices/tools/device"))
+sys.path.append(repopath("ext/modm-devices"))
 try:
     import modm_devices.parser
 except ModuleNotFoundError:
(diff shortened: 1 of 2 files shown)


salkinium:
I'm closing this in favor of working on #471.

PR #471, opened 2020-09-16
Support for STM32 F0/G0 remap pins

### LINKED ISSUE #469: Remapped pins on STM32F0/G0 not supported

chris-durand:
This adds support for remapped gpio pins on small G0/F0 controllers. 

@salkinium https://github.com/modm-io/modm-devices/pull/47/files breaks it. So we need to find a solution for that.

### DIFF
--- a/src/modm/platform/gpio/stm32/module.lb
+++ b/src/modm/platform/gpio/stm32/module.lb
@@ -3,6 +3,7 @@
 #
 # Copyright (c) 2016-2018, Niklas Hauser
 # Copyright (c) 2017, Fabian Greif
+# Copyright (c) 2020, Christopher Durand
 #
 # This file is part of the modm project.
 #
@@ -123,6 +124,22 @@ def validate_alternate_functions(driver, env):
     if not success:
         env.log.debug("Gpio signal validation failed!")
 
+def get_remap_command(family, key):
+    reg = 'SYSCFG->CFGR1'
+    mask = {
+        ('f0', 'a9') : 'SYSCFG_CFGR1_PA11_PA12_RMP',
+        ('f0', 'a10'): 'SYSCFG_CFGR1_PA11_PA12_RMP',
+        ('f0', 'a11'): 'SYSCFG_CFGR1_PA11_PA12_RMP',
+        ('f0', 'a12'): 'SYSCFG_CFGR1_PA11_PA12_RMP',
+        ('g0', 'a9') : 'SYSCFG_CFGR1_PA11_RMP',
+        ('g0', 'a10'): 'SYSCFG_CFGR1_PA12_RMP',
+        ('g0', 'a11'): 'SYSCFG_CFGR1_PA11_RMP',
+        ('g0', 'a12'): 'SYSCFG_CFGR1_PA12_RMP',
+    }.get((family, key))
+    if mask is None:
+        raise ValidateException("Unknown Remap for GPIO: 'P{}'".format(key.upper()))
+    return (reg, mask)
+
 bprops = {}
 
 # -----------------------------------------------------------------------------
@@ -199,6 +216,16 @@ def validate(env):
                 raise ValidateException("Pin '{}' already in EXTI map!".format(str(num)))
             extimap[str(num)] = vec
 
+    # Compute the set of remapped pins
+    remapped_gpios = {}
+    for p in driver["package"][0]["pin"]:
+        variant = p.get("variant", "")
+        if "remap" in variant:  # also matches "remap-default"
+            name = p["name"][1:4].strip().lower()
+            if len(name) > 2 and not name[2].isdigit():
+                name = name[:2]
+            remapped_gpios[name] = (variant == "remap") # "remap-default" -> False
+
     all_signals = {}
     for gpio in driver["gpio"]:
         key = gpio["port"] + gpio["pin"]
@@ -215,11 +242,20 @@ def validate(env):
         all_signals.update(extracted_signals)
         signal_names = sorted(list(set(s["name"] for s in extracted_signals.values())))
         extracted_signals = OrderedDict([(name, sorted([s for s in extracted_signals.values() if s["name"] == name], key=lambda si:si["name"])) for name in signal_names])
+        has_remap = key in remapped_gpios
         bprops[key] = {
             "gpio": gpio,
             "exti_irqn": extimap[gpio["pin"]],
-            "signals": extracted_signals
+            "signals": extracted_signals,
+            "has_remap": has_remap,
         }
+        if has_remap:
+            reg, mask = get_remap_command(device.identifier.family, key)
+            bprops[key].update({
+                "remap_reg": reg,
+                "remap_mask": mask,
+                "remap_value": remapped_gpios[key]
+            })
 
     all_peripherals = [s["driver"] for s in all_signals.values()]
     # Signals are not enough, since there are peripherals that don't have signals.
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ git clone --recurse-submodules https://github.com/modm-io/modm.git
 ## Targets
 
 modm can generate code for <!--avrcount-->506<!--/avrcount--> AVR,
-<!--samcount-->163<!--/samcount--> SAM and <!--stmcount-->1959<!--/stmcount-->
+<!--samcount-->163<!--/samcount--> SAM and <!--stmcount-->2010<!--/stmcount-->
 STM32 devices, however, there are different levels of support and testing.
 
 <center>
--- a/ext/st/module.lb
+++ b/ext/st/module.lb
@@ -31,8 +31,18 @@ def getDefineForDevice(device_id, familyDefines):
     if len(deviceDefines) == 1:
         return deviceDefines[0]
 
-    # now we match for the size-id.
+    # sort with respecting variants
+    minlen = min(len(d) for d in deviceDefines)
+    deviceDefines.sort(key=lambda d: (d[:minlen], d[minlen:]))
+
+    # now we match for the size-id (and variant-id if applicable).
     devNameMatch = devName + "x{}".format(device_id.size.upper())
+    if device_id.family == "l1":
+        # Map STM32L1xxQC and STM32L1xxZC -> STM32L162QCxA variants
+        if device_id.pin in ["q", "z"] and device_id.size == "c":
+            devNameMatch += "A"
+        else:
+            devNameMatch += device_id.variant.upper()
     for define in deviceDefines:
         if devNameMatch <= define:
             return define
@@ -70,7 +80,7 @@ def common_rcc_map(env):
     define = None
 
     content = Path(localpath(folder, family_header)).read_text(encoding="utf-8", errors="replace")
-    match = re.findall(r"if defined\((?P<define>STM32[F|L|G].....)\)", content)
+    match = re.findall(r"if defined\((?P<define>STM32[F|L|G][\w\d]+)\)", content)
     define = getDefineForDevice(device.identifier, match)
     if define is None or match is None:
         raise ValidateException("No device define found for '{}'!".format(device.partname))
--- a/repo.lb
+++ b/repo.lb
@@ -48,7 +48,7 @@ def check_submodules():
               "    git submodule update --init\n")
 
 # Import modm-device tools
-sys.path.append(repopath("ext/modm-devices/tools/device"))
+sys.path.append(repopath("ext/modm-devices"))
 try:
     import modm_devices.parser
 except ModuleNotFoundError:
--- a/src/modm/platform/core/stm32/module.lb
+++ b/src/modm/platform/core/stm32/module.lb
@@ -38,8 +38,6 @@ def post_build(env):
     target = env[":target"].identifier
     properties["with_crashcatcher"] = env.has_module(":crashcatcher")
 
-    if target.family in ["l4"]: # FIXME!
-        properties["regions"].remove("sram2")
     if "backup" in properties["regions"]:
         properties["regions"].remove("backup")
 
--- a/src/modm/platform/gpio/stm32/pin.hpp.in
+++ b/src/modm/platform/gpio/stm32/pin.hpp.in
@@ -74,6 +74,17 @@ public:
 	inline static void setAnalogInput() { PinSet::setAnalogInput(); }
 	/// @endcond
 
+%% if has_remap
+	/// Remap this GPIO onto physical pins with selectable GPIO mappings
+	inline static void remap() {
+	%% if remap_value
+		{{remap_reg}} |= {{remap_mask}};
+	%% else
+		{{remap_reg}} &= ~{{remap_mask}};
+	%% endif
+	};
+%% endif
+
 public:
 	// GpioOutput
 	// start documentation inherited
--- a/tools/scripts/docs_modm_io_generator.py
+++ b/tools/scripts/docs_modm_io_generator.py
@@ -29,7 +29,7 @@ def repopath(path):
 def relpath(path):
     return os.path.relpath(path, str(repopath(".")))
 
-sys.path.append(str(repopath("ext/modm-devices/tools/device")))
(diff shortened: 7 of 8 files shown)


chris-durand:
There is the following in the g071 device file:

```
      <gpio port="a" pin="9">
        <signal device-pin="c|k" device-package="u" driver="ucpd" instance="1" name="dbcc1"/>
        <signal device-pin="c|k|r" device-package="t" driver="ucpd" instance="1" name="dbcc1"/>
        <signal af="0" driver="rcc" name="mco"/>
        <signal af="1" driver="usart" instance="1" name="tx"/>
        <signal af="2" driver="tim" instance="1" name="ch2"/>
        <signal af="4" driver="spi" instance="2" name="miso"/>
        <signal af="5" driver="tim" instance="15" name="bk"/>
        <signal af="6" driver="i2c" instance="1" name="scl"/>
      </gpio>
      <gpio device-pin="c|k|r" port="a" pin="9">
        <signal device-package="i" driver="ucpd" instance="1" name="dbcc1"/>
        <signal af="0" driver="rcc" name="mco"/>
        <signal af="1" driver="usart" instance="1" name="tx"/>
        <signal af="2" driver="tim" instance="1" name="ch2"/>
        <signal af="4" driver="spi" instance="2" name="miso"/>
        <signal af="5" driver="tim" instance="15" name="bk"/>
        <signal af="6" driver="i2c" instance="1" name="scl"/>
      </gpio>
```

Therefore, lbuild tries to generate the A9 gpio header twice for the g071rb and fails.

salkinium:
I see this is more complicated than I initially thought. I still think the device file has an issue and there should not be two A9 pins, so I'm checking this.

chris-durand:
Keep in mind that this PR still has the modm-devices version from before https://github.com/modm-io/modm-devices/pull/47 was merged.

salkinium:
I don't think this "automatic" API works for this package, since you cannot tell which Pin GpioA9 is called for, and thus if you wanted to use A9 **and** A11 at the same time, your API wouldn't let you:



I think we should only add the `GpioA9::remap()` and `GpioA11::remap()` functions (analog for A10/A12) and let the user call it when necessary.

chris-durand:
Yes, for these weird G0 it doesn't work. Do you know how I would enable the `UCPD1_DBCC*` functions on A9 and A10? I'm trying to figure out how it would be used when the normal A9/A10 pin are set to `UCPD` and A9/A10 are remapped onto the A11/A12 pins.

salkinium:
> Do you know how I would enable the UCPD1_DBCC* functions on A9 and A10?

Skimming though the UCPD section of the reference manual, I have a suspicion that they should be configured as Analog Inputs, and the analog part is connected *before* the digital multiplexor, thus it would make sense that they cannot be remapped in hardware.

But I couldn't find any specific information about this in the datasheet or reference manual, except this (RM0444 Rev 2):


(Interestingly the STM32G0 only has max. 8 AFs, not 16 AFs as before)

chris-durand:
I played around with CubeMX and the G071RB. The `UCPD1_CC` pins (not remap pins) are configured as analog mode. It seems the `UCPD1_DBCC` functionality gets activated on the physical A9/A10 pins just by setting flags in SYSCFG and the A9/A10 mode value controls the remapped pins on the physical A11/A12 pins. When I had `UCPD1_DBCC` enabled on A9/A10, A11/A12 remap active and UART1 RX/TX (A9/A10 AF) on the normal A11/A12 pins, the A9/A10 registers were configured to AF mode and UART AF, not analog.

salkinium:
> the A9/A10 registers were configured to AF mode and UART AF, not analog.

Of course, that makes sense, otherwise you'd need to have analog + AF mode, which isn't how this works.

I suggest the following: Let's add a public `remap()` function to the remappable pins, which set or clear the bit as you already coded. I find it much more confusing to automagically set this function in `setOutput()`/`setInput()`/`configure()`, especially with the weird G0 package problem.

chris-durand:
I agree, with the G0 an automatic solution is very confusing and will not work. A public method called by the user is the only sensible way to go. Where would you add it? In the default pin class, the non-default or both? If it were in both and had a bool parameter, it is very confusing.

salkinium:
> Where would you add it? In the default pin class, the non-default or both?

I'd add it to only the pins occupying the same position (ie. A9/A11, A10/A12) and simply calling `remap()` sets this Gpio as remapped (ie. setting or clearing the right bit).

That way you'll get a "free" compiler error if you try it on a pin that cannot be remapped (same for AFs that don't exist on that pin).

chris-durand:
OK, I'll do that.

chris-durand:
@salkinium I think the L1 failing in CI are not caused by my changes.

salkinium:
Looks like a CMSIS header issue, considering only the `stm32l162zct6` is failing, but not it's sibling devices. I'll push some fixed on top of your branch without rebasing.

chris-durand:
There are still some issues with some G0, for example `stm32g071rbi`

salkinium:
Could you stop rebasing for a second here, I'm trying to fix something here…

chris-durand:
Sorry, I was fighting against ST's inconsistent pin names across controller families, like for example: `PB1/BOOT`, `PB1_BOOT`, `PB1-BOOT`, `PB1 BOOT`, `PB1 [BOOT]`. I won't touch anything until you are done.

salkinium:
I know, I've been there, the trick is to not fight against ST's pin naming ;-P

src/modm/platform/gpio/stm32/module.lb
```diff
@@ -199,6 +216,17 @@ def validate(env):
...
 
+    # Compute the set of remapped pins
+    positions = defaultdict(list)
+    for p in driver["package"][0]["pin"]:
+        if p.get("type", "io") == "io":
+            name = p["name"][1:4].strip().lower()
+            positions[p["position"]].append( (name, p.get("variant", "") == "remap") )
+    # Only choose the GPIOs that are on the same package position
+    remapped_gpios = {pin[0]: pin[1]
+                          for pins in positions.values() if len(pins) > 1
```
salkinium: We only care about pin names that occupy the same position, which only have the `PA9 [PA11]` format. Hence `[1:4].strip()` is enough.

salkinium:
I think this code is equivalent to yours, but please double check.

src/modm/platform/gpio/stm32/module.lb
salkinium: Btw, the code used in modm-devices is this one:
```py
def pin_name(name):
    name = name[:4]
    if len(name) > 3 and not name[3].isdigit():
        name = name[:3]
    return (name[1:2].lower(), name[2:].lower())
```

salkinium: But I guess you're right, it would be more robust to use that code, incase of a future `PA9-BLUB [PA11]`.

chris-durand: Your solution looks robust enough. It will deal well with `PB3__OSC32_IN`, `PB3OSC32IN` or similar naming atrocities … But never underestimate STs creativity. I am waiting for a French `PÉ1` next to `PE1`.

salkinium:
`stm32l162zct6` is indeed failing because GPIOF and GPIOG are not defined in the CMSIS Header. This is because `stm32l162zct6a` is the variant that hase these pins.
So modm-devices has it wrong, it should not generate Port F/G pins on `stm32l162xc` devices only on `stm32l162xc-a` devices. Whyyyy, ST?

chris-durand (commented):
@salkinium Looks good

salkinium:
Oh, there's another issue with modm-devices. Some STM32G0 devices have multiple GPIOs bonded to the same pin without any multiplexing:

```xml
      <package device-pin="y" name="WLCSP18">
        <pin position="A1" name="PA13"/>
        <pin position="A3" name="PA14-BOOT0"/>
        <pin position="A3" name="PA15"/>
        <pin position="A5" name="PB7"/>
        <pin position="A5" name="PB8"/>
        <pin position="A7" name="PC15-OSC32_OUT (PC15)"/>
        <pin position="B2" name="PA12 [PA10]"/>
        <pin position="B2" name="PA10 [PA12]" variant="remap"/>
        <pin position="B4" name="PB3"/>
        <pin position="B4" name="PB4"/>
        <pin position="B4" name="PB5"/>
        <pin position="B4" name="PB6"/>
        <pin position="B6" name="PC14-OSC32_IN (PC14)"/>
        <pin position="B6" name="PB9"/>
        <pin position="C1" name="PA11 [PA9]"/>
        <pin position="C1" name="PA9 [PA11]" variant="remap"/>
        <pin position="C3" name="PA5"/>
        <pin position="C5" name="PA1"/>
        <pin position="C7" name="VDD/VDDA" type="power"/>
        <pin position="D2" name="PA7"/>
        <pin position="D4" name="PA3"/>
        <pin position="D4" name="PA4"/>
        <pin position="D6" name="VSS/VSSA" type="power"/>
        <pin position="E1" name="PB0"/>
        <pin position="E1" name="PB1"/>
        <pin position="E1" name="PB2"/>
        <pin position="E1" name="PA8"/>
        <pin position="E3" name="PA6"/>
        <pin position="E5" name="PA2"/>
        <pin position="E7" name="PF2 - NRST"/>
        <pin position="E7" name="PA0"/>
      </package>
```

chris-durand:
Oh no, ST, why?

src/modm/platform/gpio/stm32/module.lb
```diff
@@ -199,6 +216,19 @@ def validate(env):
...
+    for p in driver["package"][0]["pin"]:
+        if p.get("type", "io") == "io":
+            name = p["name"][1:4].strip().lower()
+            if len(name) > 2 and not name[2].isdigit():
+                name = name[:2]
+            positions[p["position"]].append( (name, p.get("variant", "") == "remap") )
+    # Only choose the GPIOs that are on the same package position and have remap name format
+    remapped_gpios = {pin[0]: pin[1]
+                          for pins in positions.values() if len(pins) > 1
+                              for pin in pins  if ("[" in pin[0] and "]" in pin[0])}
```
salkinium: This is a bit inelegant, I'll check if I can instead improve the modm-devices format here.

chris-durand: F042 has no `[...]` in the remapped pin names:
`<pin position="17" name="PA11" variant="remap"/>`

salkinium: Well, now I have to fix it in modm-devices 😛

salkinium: Fixed it by explicitly marking the remap pin pair. Also makes it much easier to process in modm and works with the F042 format.

salkinium:
The failure of the STM32L162QC and STML162ZC devices has its root problem in the selection of the appropriate define and thus CMSIS header file for the device.
They both do really have port F and port G, but the selected define and header file is wrong.

This is because modm tries to find the closest *regular* match, but the mapping is not regular.
Here the [relevant part from the `stm32l1xx.h` header](https://github.com/modm-io/cmsis-header-stm32/blob/master/stm32l1xx/Include/stm32l1xx.h#L65-L86):
```c
  /* #define STM32L162xC  */   /*!< STM32L162RC and STM32L162VC */
  /* #define STM32L162xCA */   /*!< STM32L162RC-A, STM32L162VC-A, STM32L162QC and STM32L162ZC */
```

```c
  /* #define STM32L152xC  */   /*!< STM32L152CC, STM32L152UC, STM32L152RC and STM32L152VC */
  /* #define STM32L152xCA */   /*!< STM32L152RC-A, STM32L152VC-A, STM32L152QC and STM32L152ZC */
```

```c
  /* #define STM32L151xC  */   /*!< STM32L151CC, STM32L151UC, STM32L151RC and STM32L151VC */
  /* #define STM32L151xCA */   /*!< STM32L151RC-A, STM32L151VC-A, STM32L151QC and STM32L151ZC */
```

You can see that ST assigns the QC and ZC devices the A-variant header. 🙄 
I've thus fixed the define/header search to respect variants and special fix this for these devices.

chris-durand:
The `remap()` method is still missing for F042F6 A9,A10 IOs, even when I update to the latest modm-devices. This means, you can set the the pins to the remapped setting but not back. For the G071RB they're all generated as expected.

salkinium:
Ah, I didn't think about disabling this. The issues is in the raw data, which doesn't include the ` [PA9]` part. Ugh… I was doing too many things at the same time yesterday.

```
 $ ack "PINREMAP" STM32F0*
STM32F042F4Px.xml
153:	<Pin Name="PA11" Position="17" Type="I/O" Variant="PINREMAP">
169:	<Pin Name="PA12" Position="18" Type="I/O" Variant="PINREMAP">

STM32F042F6Px.xml
153:	<Pin Name="PA11" Position="17" Type="I/O" Variant="PINREMAP">
169:	<Pin Name="PA12" Position="18" Type="I/O" Variant="PINREMAP">

STM32F042G(4-6)Ux.xml
163:	<Pin Name="PA11" Position="19" Type="I/O" Variant="PINREMAP">
180:	<Pin Name="PA12" Position="20" Type="I/O" Variant="PINREMAP">

STM32F048G6Ux.xml
152:	<Pin Name="PA11" Position="19" Type="I/O" Variant="PINREMAP">
168:	<Pin Name="PA12" Position="20" Type="I/O" Variant="PINREMAP">

STM32F070F6Px.xml
118:	<Pin Name="PA11" Position="17" Type="I/O" Variant="PINREMAP">
132:	<Pin Name="PA12" Position="18" Type="I/O" Variant="PINREMAP">
```

salkinium:
Try this modm-devices fix, which adds the remap-default back in for STM32F0.

salkinium:
Nonono, none of this works right.

salkinium:
> Nonono, none of this works right.

Now modm-devices using the right position information to find the remap pin pairs.

salkinium:
The STM32G0 compile all fails because the multi-bonded pins can also be remapped!! Yay!1!!
(This is a bug, the other pins cannot be remapped).

```xml
      <package device-pin="j" name="SO8N">
        <pin position="1" name="PC14-OSC32_IN (PC14)"/>
        <pin position="1" name="PB7"/>
        <pin position="1" name="PB8"/>
        <pin position="1" name="PB9"/>
        <pin position="2" name="VDD/VDDA" type="power"/>
        <pin position="3" name="VSS/VSSA" type="power"/>
        <pin position="4" name="NRST" type="reset"/>
        <pin position="4" name="PA0"/>
        <pin position="4" name="PA1"/>
        <pin position="4" name="PA2"/>
        <pin position="5" name="PA9 [PA11]" variant="remap"/>
        <pin position="5" name="PB0" variant="remap-default"/>
        <pin position="5" name="PB1" variant="remap-default"/>
        <pin position="5" name="PA8" variant="remap-default"/>
        <pin position="5" name="PA11 [PA9]" variant="remap-default"/>
        <pin position="6" name="PA10 [PA12]" variant="remap"/>
        <pin position="6" name="PA12 [PA10]" variant="remap-default"/>
        <pin position="7" name="PA13"/>
        <pin position="8" name="PA14-BOOT0"/>
        <pin position="8" name="PA15"/>
        <pin position="8" name="PB5"/>
        <pin position="8" name="PB6"/>
      </package>
```

salkinium:
Ok, I think I'm done with this now. I've checked that STM32G071 and STM32F042 now also generate the right A9 and A10 pins.

Can you check again? Then I'll squash and merge everything.

chris-durand:
Looks good, go ahead!

PR #474, opened 2020-09-16
[freertos] Make malloc/free thread-safe

salkinium:
Partial fix for #447 making all calls to malloc thread-safe, not just the ones from within FreeRTOS.

TODO:
- [ ] Make example to test this thread-safety

cc @mikewolfram


salkinium:
Moved to #475.

PR #475, opened 2020-09-17
[freertos-plus-tcp] Add includes

strongly-typed:
There is no FreeRTOS plus TCP project yet in there so the code is not covered by the CI.
When adding `<module>modm:freertos:tcp</module>` the includes are missing

```
In file included from modm/ext/freertos_plus_tcp/FreeRTOS_DHCP.c:35:
modm/ext/freertos_plus_tcp/include/FreeRTOS_IP.h:138:10: fatal error: pack_struct_start.h: No such file or directory
  138 | #include "pack_struct_start.h"
```

What would be more elegant way to add the includes? Shall it depend on the compiler (GCC vs clang?) But only GCC is imported from the original source.

By the way, let's bump FreeRTOS module in modm to the current version.

### DIFF
--- a/ext/aws/modm_port.cpp
+++ b/ext/aws/modm_port.cpp
@@ -1,12 +1,66 @@
-
+/*
+ * Copyright (c) 2019-2020 Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
 
 #include <freertos/FreeRTOS.h>
 #include <freertos/task.h>
 #include <modm/architecture/interface/assert.hpp>
+#include <cstdlib>
 
 extern "C" void vApplicationStackOverflowHook(TaskHandle_t, char *);
-
 void vApplicationStackOverflowHook(TaskHandle_t /*pxTask*/, char *pcTaskName)
 {
 	modm_assert(false, "freertos.stack", "FreeRTOS detected a stack overflow!", pcTaskName);
+}
+
+// ----------------------------------------------------------------------------
+// Make the Newlib heap thread-safe with FreeRTOS
+
+extern "C" void __malloc_lock(struct _reent *);
+void __malloc_lock(struct _reent *)
+{
+	vTaskSuspendAll();
+}
+
+extern "C" void __malloc_unlock(struct _reent *);
+void __malloc_unlock(struct _reent *)
+{
+	xTaskResumeAll();
+}
+
+// ----------------------------------------------------------------------------
+// Make the FreeRTOS heap use Newlib heap
+
+extern "C" void *pvPortMalloc(size_t xWantedSize);
+void *pvPortMalloc(size_t xWantedSize)
+{
+	void *pvReturn = malloc(xWantedSize);
+	traceMALLOC(pvReturn, xWantedSize);
+
+#if configUSE_MALLOC_FAILED_HOOK == 1
+	if(pvReturn == nullptr)
+	{
+		extern "C" void vApplicationMallocFailedHook(void);
+		vApplicationMallocFailedHook();
+	}
+#endif
+
+	return pvReturn;
+}
+
+extern "C" void vPortFree(void *pv);
+void vPortFree(void *pv)
+{
+	if(pv)
+	{
+		free(pv);
+		traceFREE(pv, 0);
+	}
 }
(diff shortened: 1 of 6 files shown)


salkinium:
> What would be more elegant way to add the includes?

You could copy the files directly to the local folder. That way they don't spam up the include path.

> Shall it depend on the compiler (GCC vs clang?) But only GCC is imported from the original source.

We don't support any other compiler than GCC right now. We'll deal with this when we add LLVM support in the (far) future.

strongly-typed:
> You could copy the files directly to the local folder. That way they don't spam up the include path.

Like so? But I was not able to identify which command copies these files already to `portable`:

```
find . -name "pack_struct_start.h"
./modm/ext/freertos_plus_tcp/include/pack_struct_start.h
./modm/ext/freertos_plus_tcp/portable/Compiler/GCC/pack_struct_start.h
```

ext/aws/module.lb
```diff
@@ -102,6 +102,9 @@ def build(env):
     env.copy("{}/portmacro.h".format(path), "freertos/inc/freertos/portmacro.h")
     # Copy the head_3.c file that just wraps the libc malloc
     env.copy("freertos/FreeRTOS/Source/portable/MemMang/heap_3.c", "freertos/heap_3.c")
+    # Copy the compiler support files
+    env.copy("freertos/FreeRTOS-Plus-TCP/portable/Compiler/GCC/pack_struct_start.h", "freertos_plus_tcp/include/pack_struct_start.h")
+    env.copy("freertos/FreeRTOS-Plus-TCP/portable/Compiler/GCC/pack_struct_end.h", "freertos_plus_tcp/include/pack_struct_end.h")
```
salkinium: This has to go into the `:freertos:tcp` submodule above.
You can add the pack files to the ignore list:
```py
        env.copy("freertos/FreeRTOS-Plus-TCP", "freertos_plus_tcp",
            ignore=env.ignore_files("BufferAllocation_1.c"))
```

strongly-typed:
How are we dealing with the warnings?

image
image
image
image

ext/aws/module.lb
strongly-typed: Thanks, also resolved the duplicates

salkinium:
> How are we dealing with the warnings?

Yeah, that's an issue in general. You can attach compiler options to specific files ([see `:cmsis:dsp` module](https://github.com/modm-io/modm/blob/develop/ext/arm/dsp.lb#L49-L85)), however, the build system can only translate these to source files, not header files (obviously).

So if you include a FreeRTOS header file in main.cpp for example, you still get the same warnings.

strongly-typed:
Which warnings shall be fixed? Are these critical?

* `taking address of packed member of '' may result in an unaligned pointer value [-Waddress-of-packed-member]`
* `#pragma warning "ipconfigTCP_MEM_STATS_MAX_ALLOCATION undefined?"`

salkinium:
> Which warnings shall be fixed? Are these critical?

I don't know. FreeRTOS is fairly complicated and lacks clear documentation for most things.

strongly-typed:
> I don't know. FreeRTOS is fairly complicated and lacks clear documentation for most things.

`modm` ➕ `FreeRTOS` ⏩💘

😎

salkinium:
Could you cherry-pick the commit from #474 over to here?

salkinium:
What hardware are you testing this on? I think it would be neat to test this on the Nucleo-144 boards which have an integrated Ethernet port for like ~20€.

I have these lying around:
- Nucleo-F429ZI
- Disco-F746G
- Nucleo-H743ZI (but unsupported by modm)

strongly-typed:
> What hardware are you testing this on?

Not yet testing. Just get the compiler to work and read through Mike's code.


> * Nucleo-F429ZI 

I also have this board on my desk, but did not hook it up (yet) except for adjusting the example. I just moved the `FreeRTOS plus TCP` example to Nucleo-F429ZI, which makes more sense.

ext/aws/modm_port.cpp
```diff
@@ -1,12 +1,66 @@
...
 void vApplicationStackOverflowHook(TaskHandle_t /*pxTask*/, char *pcTaskName)
 {
 	modm_assert(false, "freertos.stack", "FreeRTOS detected a stack overflow!", pcTaskName);
 }
+
+// ----------------------------------------------------------------------------
+// Make the Newlib heap thread-safe with FreeRTOS
+
+extern "C" void __malloc_lock(struct _reent *);
+void __malloc_lock(struct _reent *)
```
salkinium, line 27: This doesn't work, because I intentionally wrapped `_malloc_r`, where `_r` stands for reentrant, to avoid the overhead of calling these functions 🤦 Foot meet bullet…

salkinium: I actually did it right, because we're still supporting the newlib allocator, so I just needed to add the calls to `__malloc_lock`/`__malloc_unlock` to the TLSF and Block Allocators.

salkinium (approved):
I've tested the malloc lock with all allocators in hardware.

salkinium:
Ready for squash and merge?

strongly-typed:
RTM

Issue #303, opened 2019-10-12
Callbacks for received data on CAN or UART?


WasabiFan:
It doesn't seem like there are callbacks available which are fired on receipt of, e.g., a CAN frame. As far as I can tell I need to check for items in the CAN queue manually.

Is there a way to get this without manually editing the modm interrupt handler? Was there a particular reason it was omitted?

salkinium:
modm is by design polling, there are only a few exceptions (like AdcInterrupt for some obscure reason).
The issue here is the static memory size on AVRs, providing an interrupt hook for many things quickly fills RAM. So you'd need a split API for AVR and Cortex-M. It's not impossible to do this smoothly (ie. only provide a functional API subset on AVR, which is compatible with an upgraded Cortex-M API), but it's not been done yet, due to prioritizing porting issues.

However, on Cortex-M you can hook the NVIC interrupt vector directly, iff [the interrupt table is in RAM](https://modm.io/reference/module/modm-platform-cortex-m/#vector_table_location).

Something along this pseudocode:
```c++
void uart_irq_rx()
{
    MODM_ISR_DECL(USART1_RX);
    MODM_ISR_CALL(USART1_RX);
    // your code after USART Rx call
}

// in main
NVIC_SetVector(USART1_RX_IRQn, (uint32_t)uart_irq_rx);
```

nesos:
I'm currently trying to implement a clock synchronization mechanism over CAN and I need the exact time a CAN message arrived (rather than the time I'm reading it from the queue). So interrupts would be very helpful on that.
I have to use a STM32F0, so I can't transfer the vector table to RAM.

This may be naive but wouldn't it be possible to have an abstract class as a callback class. The driver would then hold a pointer to an object to that class and if it is not null (meaning a handler is registered) the handle method will then be called.

Speaking in pseudo code the interface handler would look somehow like this:
```c++
class InterruptHandler
{
  void handleInterrupt(void) = 0;  //add parameters as needed for the callback
}
```
and the actual call could look like this:

```c++
MODM_ISR(CAN_RX0)
{
  // process the mailbox

  if (callbackHandler != nullptr)
  {
    callbackHandler->handleInterrupt();
  }
}
```
This way only the pointers have to be stored in ram. Also the callbacks could be thrown out by a generator options to keep the poll-only version of the driver.

salkinium:
[...]

PR #476, opened 2020-09-18
[driver] ENC28J60 Ethernet MAC/PHY with SPI

strongly-typed:
This is heavily WIP. Opened the PR for discussions ...

# Variable sized messages

What is the `modm` way of representing a message with variable data size, e.g. [Ethernet Frames](https://en.wikipedia.org/wiki/Ethernet_frame#Structure)? An Ethernet Frame begins with the MAC address of the  destination and then of the source. It can be between 64 and 1522 bytes long.

For the [CAN messages](https://github.com/modm-io/modm/blob/develop/src/modm/architecture/interface/can_message.hpp.in#L81), a fixed 8 bytes buffer is used.

For the [PR with FDCAN](https://github.com/rleh/modm/blob/86427ff072ba9e73a104944a3d64a713133dfde5/src/modm/architecture/interface/can_message.hpp.in#L102), the data size was not changed?! @rleh 
If it was changed to 64 bytes, it will increase memory usage for all normal CAN messages by a factor of 8 (!).

For AMNB, a mixture of fixed allocation and [dynamic allocation](https://github.com/modm-io/modm/blob/cb82eecc054b8607e80b337af7a03b4b9a6abb09/src/modm/communication/amnb/message.hpp.in#L183) is used.

What would be the best approach for Ethernet Frames?

### DIFF
--- a/examples/stm32f103c8t6_blue_pill/enc28j60/enc28j60_bank.hpp
+++ b/examples/stm32f103c8t6_blue_pill/enc28j60/enc28j60_bank.hpp
@@ -124,6 +124,73 @@ enum class Register : uint8_t
 	EPAUSH		= 0x19 | BANK_3,
 };
 
+/* ENC28J60 EIE Register Bit Definitions */
+static constexpr uint8_t EIE_INTIE	= 0x80;
+static constexpr uint8_t EIE_PKTIE	= 0x40;
+static constexpr uint8_t EIE_DMAIE	= 0x20;
+static constexpr uint8_t EIE_LINKIE	= 0x10;
+static constexpr uint8_t EIE_TXIE	= 0x08;
+/* static constexpr uint8_t  EIE_WOLIE	0x04 (reserved) */
+static constexpr uint8_t EIE_TXERIE	= 0x02;
+static constexpr uint8_t EIE_RXERIE	= 0x01;
+
+/* ENC28J60 EIR Register Bit Definitions */
+static constexpr uint8_t EIR_PKTIF	= 0x40;
+static constexpr uint8_t EIR_DMAIF	= 0x20;
+static constexpr uint8_t EIR_LINKIF	= 0x10;
+static constexpr uint8_t EIR_TXIF	= 0x08;
+/* static constexpr uint8_t EIR_WOLIF	0x04 (reserved) */
+static constexpr uint8_t EIR_TXERIF	= 0x02;
+static constexpr uint8_t EIR_RXERIF	= 0x01;
+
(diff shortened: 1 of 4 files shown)


rleh:
> For the PR with FDCAN, the data size was not changed?!

I had the same question with FDCAN and so far no answer. Therefore I did not implement the _long frame mode_ (yet).

What are the arguments against passing the data as pointer/reference (`std::span`) and let the user do the memory management?

chris-durand:
In FreeRTOS+TCP the user can choose between fixed size preallocated buffers or dynamic allocation: https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/Embedded_Ethernet_Buffer_Management.html

salkinium:
I'd make it an lbuild option to use dynamic allocation via `std::shared_ptr`, or use an integrated buffer of a certain max size.
But still awkward to fit shared memory and copied memory into the same API… (AMNB is internally reference counting to avoid copies when moving messages).

strongly-typed:
Ok, just realized that this picks up some problems from #303: "modm is by design polling"

Current implementations of CAN and UART copy every received message in the ISR to the host. This is necessary as the receive hardware buffers mostly consists of only a single frame (UART: single char, CAN: single message on most devices). Overrun could happen easily if the application is not fetching data quickly enough. The underlying assumption is host RAM >> hardware buffer.

The situation is different with ENC28J60. It has "whopping" 8 KiByte of buffer ("whopping" compared to e.g. 20 KiByte SRAM of an STM32F103RB). I'll probably delegate the memory management to the application/user.

salkinium:
@rleh and I have been wanting to move buffer/queue mangement out of the driver/lbuild into pure C++ for some time now.
The idea is to make the driver only execute a callback in the ISR and let a wrapper populate these callbacks with implementations of AtomicQueue or not.
This would move make the driver a pushing design, which can then be remodelled by the wrapper to be polling with a queue or provide even higher abstractions like directly receiving a AMNB message without a queue.

Something like an `AtomicQueueWrapper<Uart0, 1024>` would provide a predefined interface, but this would also require the callbacks to have a compatible signature (modulo packet type, ie. uint8_t or CanMessage etc). You could standardize this interface at least for packet-oriented protocols. Then you could also have a `CustomEthernetProtocolWrapper<Eth>`, which doesn't even buffer the message at all, but copies the content directly into the destination (and perhaps does defragmenting).

Not entirely sure about how to deal with DMA in this scenario tho, that's why we didn't implement this yet.

PR #479, opened 2020-09-19
[freertos] Bump version and fix include of BufferAllocation_2.c

strongly-typed:
Sorry, did not test in a clean build. Need to adopt example so the CI would have caught this issue.


ext/aws/module.lb
```diff
@@ -35,6 +35,7 @@ the config macros, then redefine them with your options, for example:
         env.outbasepath = "modm/ext"
         env.copy("freertos/FreeRTOS-Plus-TCP", "freertos_plus_tcp",
             ignore=env.ignore_files("BufferAllocation_1.c", "portable"))
```
salkinium:
```suggestion
            ignore=env.ignore_files("portable"))
```

salkinium (approved):

PR #484, opened 2020-09-27
[driver] :bug: Fix ILI9341 and add SPI example for it

### LINKED ISSUE #483: Ili9341 SPI not compiling
The Ili9341 doesn't compile for me anymore when using with SPI interface. In extension to my implementation #464 the code in the #437 adds a parallel interface. But the templated constructor prevents from compiling. Also for SPI it expects an argument to the constructor.

Maybe you could have a look into it?

asmfreak:
This closes #483 and adds an example for ILI9341 in SPI mode for Blue Pill.
Example compiles. It **should** work on real hw, but I cannot confirm this.

### DIFF
--- a/examples/stm32f103c8t6_blue_pill/graphics/main.cpp
+++ b/examples/stm32f103c8t6_blue_pill/graphics/main.cpp
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2016, Sascha Schade
+ * Copyright (c) 2017, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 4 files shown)


salkinium (approved):
Thanks!

salkinium:
Just one thing: The emojis in the commit message are not rendered as emojis in git, but as ascii. This is an issue, especially for emojis like 👷= `:construction_worker:` which make the commit message super long.

asmfreak:
Welp. I wanted to try and use [gitmoji](https://github.com/carloscuesta/gitmoji), but it does not work in cli, I agree.

salkinium:
Thanks!

mikewolfram:
Tested and works. Thanks.

PR #486, opened 2020-10-04
[CI] Github Actions instead of CircleCI

rleh:
- [x] Github Actions instead of CircleCI
- [x] Modularized docker images from modm-ext/docker-modm-build#14
- [ ] Remove CircleCI workflow

Let's see if Github Actions is faster!

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -5,9 +5,27 @@ on: [pull_request]
 jobs:
   unittests-linux-generic:
     runs-on: ubuntu-20.04
-    container:
-      image: modm/modm-build:cortex-m
+    #container:
+    #  image: modm/modm-build:cortex-m
     steps:
+      - name: Setup build environment
+        env:
+          TZ: Europe/Berlin
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq build-essential git bzip2 wget python3 python3-dev python3-pip libboost-all-dev libsdl1.2-dev libsdl-image1.2-dev libgtkmm-2.4-dev libmpc-dev libmpfr-dev libgmp-dev cmake locales texlive-latex-base texlive-fonts-recommended doxygen graphviz curl
+          sudo locale-gen en_US.UTF-8
+          pip3 install modm scons configparser numpy pyx graphviz mkdocs mkdocs-material && echo "::add-path::~/.local/bin"
+          #mkdir /opt/doxypress
+          #wget -qO- https://download.copperspice.com/doxypress/binary/doxypress-1.4.0-ubuntu20.04-x64.tar.bz2 | tar xj -C /opt/doxypress
+          #echo "::add-path::/opt/doxypress"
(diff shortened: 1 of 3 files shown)


rleh:
> Let's see if Github Actions is faster!
* CPU performance ~(IPC/Instruction per Cycle)~ seems to be way better at CircleCI (half the compile time)
* Github Actions allows infinite parallelism
* Github Actions allows self-hosted runners. Maybe @roboterclubaachen could host some runners?

salkinium:
> IPC seems to be way better as CircleCI (half the compile time)

What does IPC mean in this context? 

`stm32f4-compile-all` runs for 29mins on GitHub Actions, but only 10mins on CircleCI.
If you compare the GitHub Actions times with CircleCI its like 3x slower right?

salkinium:
I mean the whole CircleCI dependencies are structured around failing obvious mistakes quickly in less than 5mins. I'm not sure if you can beat this if even just compiling stm32f4-examples takes 9mins…

rleh:
> What does IPC mean in this context?

CPU performance

>  If you compare the GitHub Actions times with CircleCI its like 3x slower right?

Yes. 2.5 ... 3 times slower. Maybe we have to increase the parallelism (currently `-j4`) to 8x or 16x to fully utilize the CPU.

salkinium:
[GitHub Actions Resources](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources)

> - 2-core CPU
> - 7 GB of RAM memory
> - 14 GB of SSD disk space

[CircleCI Resources for medium](https://circleci.com/docs/2.0/configuration-reference/#resource_class)

> - 2 vCPU
> - 4 GB RAM
> - Not sure how much disk space

But I remember experimenting with the number of CPU jobs and found 4 to be best.

rleh:
The point with Github Actions is: We are not limited to 4x parallelism.

We can split the CI jobs in more but smaller ones and thus get the results not only of some selected tests, but of all tests quickly.

rleh:
> But I remember experimenting with the number of CPU jobs and found 4 to be best.

For CircleCI. Might be different in the Github Actions environment...

salkinium:
> We can split the CI jobs in more but smaller ones and thus get the results not only of some selected tests, but of all tests quickly.

Sure, but you're down a fixed <1m30 for environment initialization per job, so you'll have to split the F4 Examples into a lot of parts to get to <5mins.

rleh:
>  fixed <1m30 for environment initialization

From my experience this goes down to few (<30) seconds once the docker images are in Githubs cache.

From my point of view the STM32F4 examples do not have to be ready in <5 minutes.
I think it is more important to reduce the total time and to make the results and logs more accessible.

salkinium:
> I think it is more important to reduce the total time and to make the results and logs more accessible.

I mean the entire CircleCI job is done in 25mins. The GHA one takes 40mins. You will have to split this into a lot of jobs.

salkinium:
COLOR ME SKEPTICAL!1!! \*looks at spraycans\*

PR #487, opened 2020-10-05
[CI] CirlceCI: Use modularized docker containers

rleh:
from modm-ext/docker-modm-build#14

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,41 +26,50 @@ jobs:
             lbuild --version
       - run:
           name: Hosted Unittests
+          when: always
           command: |
             (cd test && make run-hosted-linux)
       - run:
           name: Compile STM32 Unittests
+          when: always
           command: |
             (cd test && make compile-nucleo-f411)
-            (cd test && make compile-nucleo-f103)
             (cd test && make compile-nucleo-l432)
+            (cd test && make compile-nucleo-f103)
       - run:
           name: Linux Examples
+          when: always
           command: |
             (cd examples && ../tools/scripts/examples_compile.py linux)
       - run:
           name: Generic Examples
+          when: always
           command: |
             (cd examples && ../tools/scripts/examples_compile.py generic)
       - run:
           name: Examples SAM Devices
+          when: always
           command: |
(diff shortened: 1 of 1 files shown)


salkinium:
I'd like to also unify these three commands into one, so that all three always get run, even if they fail. Otherwise you have to fix trailing whitespace, then wait 5mins, then fix docs, then wait 5mins, etc, which is dumb.

```yml
      - run:
          name: Check for Trailing Whitespace
          command: |
            python3 tools/scripts/rm_whitespace.py
      - run:
          name: Synchronize Documentation
          command: |
            python3 tools/scripts/synchronize_docs.py -d
      - run:
          name: Check for Unique Build Paths
          command: |
            python3 tools/scripts/examples_check.py examples
```

Into this? I think this still fails.

```yml
      - run:
          name: Sanity Checks: Whitespace, Docs, Build Paths
          command: |
            python3 tools/scripts/rm_whitespace.py
            python3 tools/scripts/synchronize_docs.py -d
            python3 tools/scripts/examples_check.py examples
```

I'll try this in another PR though, since this is off topic.

rleh:
`when: always`?

salkinium:
> `when: always`?

https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute

Ah, that's obviously the right™ way. 🤦 

We should probably add this to all relevant steps, right?

salkinium:
Multiple commands still abort obviously, so I pulled the relevant ones apart. Just a small optimization, but this works exactly like I want it too. EXCELLENT!

examples/avr/assert/main.cpp
```diff
@@ -29,7 +29,7 @@ modm_abandon(const modm::AssertionInfo &info)
 	serialStream << IFS(" failed!\nAbandoning...\n");
 #endif
 
-	Leds::setOutput();
+	Leds::setOutput();asdfasdfasdf
```
rleh: Now remove all the `asdfasdfasdf` and we are ready to merge!

salkinium: \*whip\*

rleh (commented):
Approve :heavy_check_mark:

salkinium (approved):

PR #489, opened 2020-10-06
Fix Rcc:PllSource for some STM32F3 devices and add Nucleo-F334R8 board

rleh:
@hshose [discovered](https://github.com/modm-io/modm/pull/414#commitcomment-42985517) that there is no `Hsi` element in `Rcc::PllSource` on STM32F373 devices, d89c06e should fix it.

Sadly there is no STM32F373 Nucleo or Discovery board, but I found a Nucleo-F334R8 board in my dev-board box, so I added it to modm.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
           name: Examples STM32F3 Series
           when: always
           command: |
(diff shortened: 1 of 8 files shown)


salkinium (approved):
Thanks!

PR #493, opened 2020-10-12
Add getters for basic timer period settings

mcbridejc:
After configuring the timer, I needed to set a period in real units, but I did not see any way to determine the frequency after the timer configuration code picks the best settings from the target frequency. This adds accessors to allow that.


salkinium (approved):

salkinium:
Squashed and rebased, I'll merge this after the CI passes.

PR #494, opened 2020-10-12
Rename unittest.py to unit_test.py

### LINKED ISSUE #492: Python tools fail with cmake: "cannot import name 'SkipTest' from 'unittest'"
When executing `python3 modm/modm_tools/openocd.py -f modm/openocd.cfg program.elf`, I get the following error:

```
  File "modm/modm_tools/openocd.py", line 33, in <module>
    from modm_tools import utils, backend as bem
  File "modm/modm_tools/__init__.py", line 25, in <module>
    from . import info
  File "modm/modm_tools/info.py", line 20, in <module>
    from git import Repo
  File "/home/jeff/.local/lib/python3.8/site-packages/git/__init__.py", line 40, in <module>
    from git.config import GitConfigParser  # @NoMove @IgnorePep8
  File "/home/jeff/.local/lib/python3.8/site-packages/git/config.py", line 24, in <module>
    from git.util import LockFile
  File "/home/jeff/.local/lib/python3.8/site-packages/git/util.py", line 18, in <module>
    from unittest import SkipTest
ImportError: cannot import name 'SkipTest' from 'unittest' (/home/jeff/dev/touchdemo-c++/modm/modm_tools/unittest.py)
```

The problem seems clear: `sys.path[0]` points to `<projectdir>/modm/modm_tools`, so the modm provided `unittest.py` is being imported instead of the python standard package. I don't see anything special about my system to cause this, except that I am using cmake. Using scons, `scons program` works fine, I believe because in this case the `openocd.py` is imported, not run as an executable, and therefor its path is not set in `sys.path[0]` by the python interpreter.

mcbridejc:
The unittest.py module overrides the python standard package when executing scripts in the same directory (e.g. `gdb.py`).
This addresses #492.

### DIFF
--- a/tools/build_script_generator/module.lb
+++ b/tools/build_script_generator/module.lb
@@ -25,7 +25,7 @@ class BuildModuleDocs:
         if self._content is None:
             self._content = Path(localpath("module.md")).read_text(encoding="utf-8").strip()
             tools = ["avrdude", "openocd", "bmp", "gdb", "size",
-                     "unittest", "log", "build_id", "bitmap"]
+                     "unit_test", "log", "build_id", "bitmap"]
 
             for tool in tools:
                 tpath = Path(repopath("tools/modm_tools/{}.py".format(tool)))
@@ -171,7 +171,7 @@ def build(env):
         "bitmap",
         "find_files",
         "info",
-        "unittest",
+        "unit_test",
         "utils",
     }
     if is_cortex_m:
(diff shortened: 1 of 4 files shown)


tools/build_script_generator/module.lb
```diff
@@ -25,7 +25,7 @@ class BuildModuleDocs:
         if self._content is None:
             self._content = Path(localpath("module.md")).read_text(encoding="utf-8").strip()
             tools = ["avrdude", "openocd", "bmp", "gdb", "size",
-                     "unittest", "log", "build_id", "bitmap"]
+                     "unit_test", "log", "build_id", "bitmap"]
```
salkinium, line 28: There's another occurrence at line 174.

salkinium:
Oh wow, you found a completely unrelated bug in git info. I'll push a fix onto your branch in a second.

salkinium (approved):

PR #496, opened 2020-10-15
Miscellaneous fixes/features/tweaks

salkinium:
This is a collection PR of small improvements, fixes, tweaks that accumulated over some time while working on real world projects.

- [architecture] Add binary import helper macro. Unfortunately it does declare any dependency on the the binary, so SCons will not recompile the inclusion file if the binary changed. However, this is a generic problem that's not really an issue of the binary inclusion mechanism.
- [ui] Animation should always run for the fade time, even if the value change is zero. This is necessary for RGB LEDs, where it makes sense to have one LED d
- [gpio] Gpio::toggle() returns previous output state on STM32. This is not part of the interface, since it may cost extra to read the previous state, but on STM32 it must be done, since there is no hardware toggle.o nothing.
- [amnb] Fine-tune retransmission attempts for a busier bus. Testing this with 10+ nodes requires significantly more attempts before discarding the message.
- [amnb] Allow actions and listeners with binary data (Action return types must still be fixed size though).
- [amnb] Call all compatible listeners for a command.
- [color] Make color constexpr.
- [math] Make modm::enumerate constexpr.

### DIFF
--- a/src/modm/communication/amnb/node.hpp
+++ b/src/modm/communication/amnb/node.hpp
@@ -224,7 +224,7 @@ protected:
 					const auto listener = listenerList[ii];
 					if (rx_msg.command() == listener.command) {
 						if (complete) listener.call(rx_msg);
-						return true;
+						else return true;
 					}
 				}
 				break;
--- a/src/modm/communication/amnb/node.hpp
+++ b/src/modm/communication/amnb/node.hpp
@@ -71,16 +71,21 @@ public:
 		if (tx_queue.isFull()) return false;
 		return tx_queue.push(std::move(Message(address, command, Type::Broadcast)));
 	}
+	bool
+	broadcast(uint8_t command, const uint8_t *data, size_t length)
+	{
+		if (tx_queue.isFull()) return false;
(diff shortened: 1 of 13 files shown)


src/modm/communication/amnb/node.hpp
```diff
@@ -314,7 +327,8 @@ class Node : public modm::Resumable<6>
 	}
 	response_status{ResponseStatus::NotWaiting};
 
-	static constexpr uint8_t PRIORITY_BITS{5};
+	static constexpr uint8_t MIN_TX_TRIES{20};
+	static constexpr uint8_t PRIORITY_BITS{6};
 	static constexpr uint8_t RESCHEDULE_MASK_SHORT{7};
 	static constexpr uint8_t RESCHEDULE_MASK_LONG{11};
```
salkinium, line 333: I'm not happy about the magic numbers here, but I don't have a proper test setup here and currently lack instrumentation of this code. This isn't even true p-persistent CSMA, and the timings depend on the baudrate and median message size. I'm going to have to do some more research once I have time.

Issue #495, opened 2020-10-13
delay_ns off by a factor of two


mcbridejc:
I think this must be a bug, though I haven't yet put my finger on what's going on.

Board: Nucleu-F303K8 (I'm using the modm board config, and haven't messed with clock setup, so it's running from HSI @ 64MHz)

My test: 
```
{
   modm::atomic::Lock lck;
   GpioA0->setOutput(true);
   modm::delay_ns(pulse_width);
   GpioA0->setOutput(false);
   modm::delay_ns(delay);
}
```
I'm observing with scope that the delay is double the requested value. This is consistent for values ranging from 1500 to 15000 ns; it always runs twice as long as requested. 

The delay_ns_per_loop appears to have been calculated correctly for 64MHz:
```
(gdb) p modm::platform::delay_ns_per_loop
$1 = 47
```

I have a timer PWM output that has the expected timing, so I believe that I am actually running with the correct clock frequency. 

I've tried stepping through delay loop in the debugger, and monitoring `DWT.CYCCNT`. This only increments 2 per loop iteration, not the expected 3, which is an error in the wrong direction to explain this problem. I'm not sure though if the debugger is affecting branch timing. 

Disassembled function (from gdb) looks good to me: 
```
0x8000498 <modm::delay_ns(unsigned long)>       ldr     r3, [pc, #12]   ; (0x80004a8 <modm::delay_ns(unsigned long)+16>)         
0x800049a <modm::delay_ns(unsigned long)+2>     movs    r2, #10                                                                  
0x800049c <modm::delay_ns(unsigned long)+4>     ldrh    r3, [r3, #0]                                                             
0x800049e <modm::delay_ns(unsigned long)+6>     muls    r2, r3                                                                   
0x80004a0 <modm::delay_ns(unsigned long)+8>     subs    r0, r0, r2                                                               
0x80004a2 <modm::delay_ns(unsigned long)+10>    subs    r0, r0, r3                                                               
0x80004a4 <modm::delay_ns(unsigned long)+12>    bpl.n   0x80004a2 <modm::delay_ns(unsigned long)+10>                             
0x80004a6 <modm::delay_ns(unsigned long)+14>    bx      lr                                                                       
0x80004a8 <modm::delay_ns(unsigned long)+16>    lsls    r0, r5, #6                                                               
0x80004aa <modm::delay_ns(unsigned long)+18>    asrs    r0, r0, #32                                                              
```
[...]

PR #498, opened 2020-11-14
Add fastcode attribute to delay_ns

mcbridejc:
See issue #495. 

I found the delay_ns to be off by scale factor of 2 on stm32f303. Putting the code into fast code section fixes this, at least for this part.

### DIFF
--- a/src/modm/platform/core/cortex/delay_impl.hpp.in
+++ b/src/modm/platform/core/cortex/delay_impl.hpp.in
@@ -31,7 +31,7 @@ extern uint16_t delay_ns_per_loop;
 extern uint16_t delay_fcpu_MHz;
 }
 
-inline void delay_ns(uint32_t ns)
+inline void modm_fastcode delay_ns(uint32_t ns)
 {
 	volatile uint32_t overhead_cycles;
 	// ns_per_loop = nanoseconds per cycle times cycles per loop ({{loop}} cycles)
--- a/src/modm/platform/core/cortex/delay_impl.hpp.in
+++ b/src/modm/platform/core/cortex/delay_impl.hpp.in
@@ -33,14 +33,15 @@ extern uint16_t delay_fcpu_MHz;
 
 inline void delay_ns(uint32_t ns)
 {
+	volatile uint32_t overhead_cycles;
 	// ns_per_loop = nanoseconds per cycle times cycles per loop ({{loop}} cycles)
 	asm volatile (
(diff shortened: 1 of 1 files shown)


salkinium:
This will only work on devices with a single-cycle access instruction cache, like the F3 (ICCM), F7 (I-Cache) and H7 I-Cache).
For all other devices the `delay_ns_per_loop` is likely still wrong.
I'm going to keep this open until I have the time to do some measurements.

mcbridejc:
Since I had a STM32G431 board handy, I decided to do a quick test on that. Here, the delay_ns function works roughly as expected, both with and without the modm_fastcode attribute enabled.

But I think I found another issue with the delay_ns function. In my test code, I happened to also use a delay_us as well: 

```
while(true) {
    GpioA0::setOutput(true);
    modm::delay_ns(5000);
    GpioA0::setOutput(false);
    modm::delay_us(10);
}
```
With a delay_us value of 10, I got a delay of 180us. With any other value I tried, I got the expected delay. Debugger showed the argument was in fact passed as 180, when I tried to pass 10. Inspecting disassembly, I found that when I used 10, the compiler passed the value of register r6 as first argument (`mov    r0, r6`). Any other value resulted in loading a literal, e.g. `movs    r0, #9`. After some hair pulling, I realized this is because of the [assembly block](https://github.com/modm-io/modm/blob/b0107758478a06a3aa02592f438125da25802aa1/src/modm/platform/core/cortex/delay_impl.hpp.in#L37) in the delay_ns function above. Because the argument value matched the value loaded into the register there (10) it re-used it, but actually it had been multiplied by 18 in the inline asm.

From https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#InputOperands:

> Warning: Do not modify the contents of input-only operands (except for inputs tied to outputs). The compiler assumes that on exit from the asm statement these operands contain the same values as they had before executing the statement. It is not possible to use clobbers to inform the compiler that the values in these inputs are changing. One common work-around is to tie the changing input variable to an output variable that never gets used. Note, however, that if the code that follows the asm statement makes no use of any of the output operands, the GCC optimizers may discard the asm statement as unneeded (see Volatile). 

I'm about to push a change that fixes it, by adding an output to the asm block.

salkinium (approved):
Excellent catch!

salkinium:
Will merge next after CI passes. (Rebased on develop and reworded the commit messages slightly)

PR #502, opened 2020-11-17
Github Actions: add-path command was deprecated

rleh:
Now fixed.

New syntax: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path
(Background: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)


salkinium (approved):
Thanks!

PR #503, opened 2020-11-17
ADS7843: Use pressure measurement for touch detection and different filtering

mcbridejc:
I tried using the ADS7843 driver with this display: http://hiletgo.com/ProductDetail/2157216.html. I found the touch detection to not be very good; if I held my finger firmly on the display it would go back and forth between pressed or not. 

With a little reading, it seems that the recommended approach is to use the Z registers for a pressure calculation, and this seems to work much better at identifying when the screen is touched, at least for this display. 

I'm hesitant to recommend this, because even though this works much better for me, it is changing the driver behavior in ways that may affect other projects. At the same time, I found this driver to be essentially unusable as-is. I'm interested in feedback on how best to proceed from anyone interested.

### DIFF
--- a/src/modm/driver/touch/ads7843.hpp
+++ b/src/modm/driver/touch/ads7843.hpp
@@ -44,7 +44,7 @@ namespace modm
 		 * 			stable enough to provide a reading, otherwise `false`.
 		 */
 		static bool
-		read(modm::glcd::Point * point);
+		read(modm::glcd::Point * point, uint16_t pressure_threshold=defaultThreshold);
 
 		static inline uint16_t
 		readX()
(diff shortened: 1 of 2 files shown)


salkinium:
Glancing over the datasheet it seems to me that the Z-registers are the two axillery analog input channels IN3 and IN4 of the driver? So it seems like this is a different hardware setup which also exposes a resistive Z axis to the driver?

Do you want to create a new class, which inherits from the base class, to provide the additional functionality? That way it's backwards compatible and also allows future customization for other uses of IN3, IN4.

mcbridejc:
No it's not that; the IN3 and IN4 inputs are not connected on this display. 

Atmel has a good app note describing the the pressure/touch resistance calculation: http://ww1.microchip.com/downloads/en/Appnotes/doc8091.pdf

I am actually using an XPT2046 controller (and this datasheet does refer to pressure sensing). I was under the impression that it was a 100% clone of the ADS7843, but now I am not so sure. The ADS7843 does not mention the Z1/Z2 input configurations (though the fact that it has the same 3 bits for the mode but lists 4 possible modes is suspicious), and the [ADS7846 datasheet](https://www.ti.com/lit/ds/symlink/ads7846.pdf) calls pressure sensing an added feature over the ADS7843. 

> Do you want to create a new class, which inherits from the base class, to provide the additional functionality? That way it's backwards compatible and also allows future customization for other uses of IN3, IN4.

Personally I don't think this makes sense, as it isn't really added functionality, it's just a different way to implement the same functionality, and adds the ability to adjust the touch threshold. One approach is that the base driver should just be register level access, and not have any thresholding or filtering, but that breaks the API compatibility. 

Given that I don't have an actual ADS7843 to test on, I'm inclined to close this and leave it as-is for now.

PR #512, opened 2020-11-24
[architecture] Make operator<< for can::Message usable in namespace modm

### LINKED ISSUE #509: << operator for modm::can::Message in generated code
I have a problem with `modm::can::Message`

It works in a file included from `main.cpp` but not in generated code.

```
./modm/src/modm/communication/other/other_impl.hpp:15:17: error: no match for 'operator<<' (operand types are 'modm::log::Logger' and 'modm::can::Message')
   15 |  MODM_LOG_DEBUG << tx;
      |                 ^~ ~~
      |                    |
      |                    modm::can::Message
```

I have an example [here](https://github.com/strongly-typed/modm/tree/logger-canmessage/examples/linux/logger)

Compiles from [`main`](https://github.com/strongly-typed/modm/blob/logger-canmessage/examples/linux/logger/main.cpp#L39)

and from [`examples/linux/logger/other_impl.hpp`](https://github.com/strongly-typed/modm/blob/logger-canmessage/examples/linux/logger/other_impl.hpp#L17)

but not in [`modm/communication/other/other_impl.hpp`](https://github.com/strongly-typed/modm/blob/logger-canmessage/src/modm/communication/other/other_impl.hpp#L15) 

The generated code in `./modm/src/modm/architecture/interface/can_message.hpp` has the operator `<<` included.

I probably do something completely wrong here ... but I cannot see it. Any suggestions?

chris-durand:
Fixes #509. The operator was defined in the global namespace and is not found by name lookup when used from within namespace modm.

@strongly-typed Could you please check if this works for you?

### DIFF
--- a/src/modm/architecture/interface/can_message.cpp.in
+++ b/src/modm/architecture/interface/can_message.cpp.in
@@ -35,8 +35,11 @@ modm::can::Message::operator == (const modm::can::Message& rhs) const
 #include <inttypes.h>
 #include <modm/io/iostream.hpp>
 
+namespace modm
+{
+
 modm::IOStream&
-operator << (modm::IOStream& s, const modm::can::Message m)
+operator << (modm::IOStream& s, const modm::can::Message& m)
 {
 	s.printf("id = %04" PRIx32 ", len = ", m.identifier);
 	s << m.length;
@@ -49,5 +52,7 @@ operator << (modm::IOStream& s, const modm::can::Message m)
 		}
 	}
 	return s;
+}
+
 }
 %% endif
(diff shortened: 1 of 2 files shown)


salkinium:
There are many more `operator <<` like this in the global namespace (all of them I think), so does this mean we did it wrong for everything?

Should we fix it for everything then?

strongly-typed:
> There are many more operator <<

It is not unlikely that I copied some of them from the initial example 🤓

> Should we fix it for everything then?

Yes, who's doing it? I could if you don't mind?

salkinium:
> Yes, who's doing it? I could if you don't mind?

Please do.

salkinium:
Closing this in favor of #513.

PR #518, opened 2020-12-14
FreeRTOS example on stm32f4_discovery board

ayoubSoussi:
Basic example of using FreeRTOS on an STM32F407VG board similar to the one that exists in the NUCLEO-F429zi examples.
It uses 3 threads to make 3 LEDs blink at different rates.
The plan is to add later another thread for an SSD1306 LCD screen (still working on it).


salkinium:
> The plan is to add later another thread for an SSD1306 LCD screen (still working on it).

Ok, I'm converting this to a Draft and wait until you're ready.

salkinium:
I'm closing this due to inactivity, but feel free to open a new PR when you're ready.

PR #519, opened 2020-12-14
[freertos] Add enter/exit critical section functions

ayoubSoussi:
Wrapping taskENTER_CRITICAL and taskEXIT_CRITICAL of FreeRTOS to deal with critical sections.

I tested it on an STM32F4 discovery board.

### DIFF
--- a/src/modm/processing/rtos/freertos/thread.cpp
+++ b/src/modm/processing/rtos/freertos/thread.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2009-2012, Fabian Greif
  * Copyright (c) 2012, 2018, Niklas Hauser
+ * Copyright (c) 2020, Ayoub SOUSSI
  *
  * This file is part of the modm project.
  *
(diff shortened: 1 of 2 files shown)


salkinium:
Ah interesting: [`modm::atomic::Lock`](https://github.com/modm-io/modm/blob/develop/src/modm/platform/core/cortex/atomic_lock_impl.hpp.in) sets the global interrupt enable/disable flags, but [`taskENTER_CRITICAL()`](https://github.com/modm-ext/freertos-partial/blob/a5317393d92e283980f809c9fc7a6510c4a57a6e/FreeRTOS/Source/portable/GCC/ARM_CM3/port.c#L366-L392) calls `portDISABLE_INTERRUPTS()`, which is implemented by raising/lowering BASEPRI. That's a notable semantic difference, since the scheduler will be blocked with `modm::atomic::Lock`, while `taskENTER_CRITICAL()` will allow the scheduler to continue.

The `modm::rtos` code is meant to provide an abstraction over `std::thread`, so that we can run threaded code on an hosted OS, for our CI and easier algorithm testing/debugging. So it's not intended to abstract FreeRTOS, but provide a `std::thread` compatible implementation *using* FreeRTOS.

So for more advanced use-cases I recommend simply using FreeRTOS directly, instead of completing the API, otherwise you'd have to add a lot more wrapper code, which just adds more indirection…

ayoubSoussi:
Oh thanks for the clarifications

PR #520, opened 2020-12-15
[cmake] Fix typo in CMakeLists template

rleh:


salkinium (approved):
Thanks.

This grep pattern is not particularly useful tho, but I couldn't find out how to do it properly recursively…

rleh:
I think (with my CMake knowledge) that the CMake way here would be a different one anyway, but as long as it works....

PR #522, opened 2020-12-19
[stm32] Fix [[nodiscard]] for Can::initialize()

chris-durand:
Put the [[nodiscard]] attribute in the right location.

### DIFF
--- a/src/modm/platform/can/stm32/can.hpp.in
+++ b/src/modm/platform/can/stm32/can.hpp.in
@@ -106,9 +106,9 @@ public:
 	 * 			other function from this class!
 	 */
 	template< class SystemClock, bitrate_t bitrate=kbps(125), percent_t tolerance=pct(1) >
-	static inline bool
+	[[nodiscard]] static inline bool
 	initialize(	uint32_t interruptPriority, Mode startupMode = Mode::Normal,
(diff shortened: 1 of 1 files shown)


rleh (approved):
Thanks :+1:

chris-durand:
Here we get the same linker error with gcc10 as in the C++20 pull request.
The line of code triggering it [looks a bit unsual](https://github.com/modm-io/modm/blob/develop/ext/aws/FreeRTOSConfig.h.in#L39):
`modm_extern_c uint32_t _ZN4modm8platform4fcpuE;`

(a C++ mangled name inside an extern "C" block to be also used from C files, where `modm_extern_c` is empty)

This is a definition, not a declaration when compiled as C. This should fail if this file is included in more than one translation unit.

salkinium:
So it should be like this?

```c++
extern
#ifdef __cplusplus
"C" 
#endif
uint32_t _ZN4modm8platform4fcpuE;
```

chris-durand:
I would have suggested that, but when I looked closer I didn't find another definition (see #523). We might have to add one in a .c/.cpp file.

rleh:
All CircleCI tests have passed, ~but for some reason Github does not know about it~.

PR #528, opened 2021-01-05
[test] Fix timestamp test

chris-durand:
std::chrono::duration is uninitialized when default constructed, for performance reasons. This fix adds two `{}` to zero-initialize the duration objects in `DurationTest::testConstructors()`.


rleh (approved):

salkinium (approved):
Whoops, thanks!

PR #531, opened 2021-01-06
[test] Fix angle_test on AVR

chris-durand:
`angle_test` fails on AVR because an angle wraps at pi due to numerical reasons.


salkinium (approved):
Thanks!

PR #532, opened 2021-01-08
[stm32] Fix F0/F1 PLL pre-divider setting

chris-durand:
This PR fixes the issue of PR https://github.com/modm-io/modm/pull/99 opened by @daniel-k in 2018.
Bit 17 in register `RCC->CFGR` (`RCC_CFGR_PLLXTPRE`) aliases the LSB of the PLL pre-divider setting in `RCC->CFGR2`. They were written in such a way that the LSB of the pre-divider was always zero. This is fixed by setting `CFGR2` after `CFGR`.

Furthermore a parameter is added to set the `RCC_CFGR_PLLXTPRE` bit on  F10{1, 2, 3} devices which acts as a HSE clock /2 pre-divider. These devices don't include the /1-/16 divider.

I don't have access to any of the affected hardware right now, but I am quite confident the changes will work.

### DIFF
--- a/src/modm/platform/clock/stm32/module.lb
+++ b/src/modm/platform/clock/stm32/module.lb
@@ -3,6 +3,7 @@
 #
 # Copyright (c) 2016-2019, Niklas Hauser
 # Copyright (c) 2017, Fabian Greif
+# Copyright (c) 2021, Christopher Durand
 #
 # This file is part of the modm project.
 #
@@ -57,6 +58,7 @@ def build(env):
     properties["pllprediv"] = \
         (target["family"] in ["f0", "f3"] or (target["family"] == "f1" and target["name"] in ["00", "05", "07"]))
     properties["pllprediv2"] = False    # FIXME: not sure what value this should have
+    properties["pll_hse_prediv2"] = target["family"] == "f1" and target["name"] in ["01", "02", "03"]
     properties["hsi48"] = \
         target["family"] == "f0" and target["name"] in ["42", "48", "71", "72", "78", "91", "98"]
     properties["pll_p"] = ((target["family"] == "l4" and target["name"] not in ["12", "22"]) or target["family"] == "g4")
--- a/src/modm/platform/clock/stm32/rcc.cpp.in
+++ b/src/modm/platform/clock/stm32/rcc.cpp.in
@@ -261,6 +261,12 @@ modm::platform::Rcc::enablePll(PllSource source, const PllFactors& pllFactors, u
 	// Pll multiplication factor
 	tmp |= (static_cast<uint32_t>(pllFactors.pllMul - 2) << 18) & {{pullmul}};
 
+%% if pll_hse_prediv2
+	if(pllFactors.enableHsePllPrediv2) {
+		tmp |= RCC_CFGR_PLLXTPRE;
+	}
+%% endif
+
 	RCC->CFGR = tmp;
 
 %% if pllprediv
--- a/src/modm/platform/clock/stm32/rcc.cpp.in
+++ b/src/modm/platform/clock/stm32/rcc.cpp.in
@@ -5,7 +5,7 @@
  * Copyright (c) 2012, 2016, Sascha Schade
  * Copyright (c) 2012, 2014-2019, Niklas Hauser
  * Copyright (c) 2013-2014, Kevin Läufer
- * Copyright (c) 2018, Christopher Durand
+ * Copyright (c) 2018, 2021, Christopher Durand
  *
  * This file is part of the modm project.
  *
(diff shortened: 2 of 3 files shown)


rleh:
Related: #414

rleh (commented):
I think we should rename the `PllSource` option on devices with a fixed HSE prescaler of /2 from `Hse` to `HseDiv2`. This is not the case as far as I can see.

What do you think @chris-durand?

chris-durand:
>I think we should rename the `PllSource` option on devices with a fixed HSE prescaler of /2 from `Hse` to `HseDiv2`.

Hm, I am not aware of any such devices. This addition only applies to F10{1, 2, 3} which have a configurable /2 `Hse` divider.

A design alternative would be to add `HseDiv2` as an additional clock source. There are pros and cons to both approaches. `HseDiv2` would be consistent with `HsiDiv2` but wouldn't reflect what the hardware actually does. The PllSource enum directly maps to the `PLLSRC` bits in the `PLLCFGR` register but the HSE divider is configured in a different place. As the PLL configuration is a shallow abstraction to registers it seems preferable to me to maintain consistency with the hardware implementation.

chris-durand:
What just came to my mind is that we have C++20 with non-type template parameters soon. It will be quite easy to add some additional checks on the PLL constants at compile time. The PLL factors are plain `int`s, now you can specify any unsupported value and it will silently fail at runtime. We could pass the constant struct object as an NTTP and check the constants with `static_assert`. Short example on how that would work: https://gcc.godbolt.org/z/shbnc8

With a bit of work on the CubeMX data it could be even possible to have some information on valid frequency ranges in the device files and check them at compile time, too.

What do you think? @rleh @salkinium

salkinium:
>  CubeMX data it could be even possible to have some information on valid frequency ranges

Yes definitely, the clock graphs already have some of these constraints, incl. prescaler ranges: https://salkinium.com/stm32/clock/F105.svg

However, the CubeMX data also contains a constraint solver (or filter?) with conditions written as Java Code in XML comments that get filtered by the GUI somehow. I couldn't figure it out by just looking at the data, there would need to be some reverse engineering of the App, plus reimplementing of that solver/filter.

src/modm/platform/clock/stm32/module.lb
```diff
@@ -57,6 +58,7 @@ def build(env):
     properties["pllprediv"] = \
         (target["family"] in ["f0", "f3"] or (target["family"] == "f1" and target["name"] in ["00", "05", "07"]))
     properties["pllprediv2"] = False    # FIXME: not sure what value this should have
+    properties["pll_hse_prediv2"] = target["family"] == "f1" and target["name"] in ["01", "02", "03"]
```
salkinium, line 61: The CMSIS header suggest there are more devices with this predivider?
```
 $ ack PLLXTPRE -l
stm32f0xx/Include/stm32f030x6.h
stm32f0xx/Include/stm32f030x8.h
stm32f0xx/Include/stm32f030xc.h
stm32f0xx/Include/stm32f031x6.h
stm32f0xx/Include/stm32f038xx.h
stm32f0xx/Include/stm32f042x6.h
stm32f0xx/Include/stm32f048xx.h
stm32f0xx/Include/stm32f051x8.h
stm32f0xx/Include/stm32f058xx.h
stm32f0xx/Include/stm32f070x6.h
stm32f0xx/Include/stm32f070xb.h
stm32f0xx/Include/stm32f071xb.h
stm32f0xx/Include/stm32f072xb.h
stm32f0xx/Include/stm32f078xx.h
stm32f0xx/Include/stm32f091xc.h
stm32f0xx/Include/stm32f098xx.h

stm32f1xx/Include/stm32f100xb.h
stm32f1xx/Include/stm32f100xe.h
stm32f1xx/Include/stm32f101x6.h
stm32f1xx/Include/stm32f101xb.h
stm32f1xx/Include/stm32f101xe.h
stm32f1xx/Include/stm32f101xg.h
stm32f1xx/Include/stm32f102x6.h
stm32f1xx/Include/stm32f102xb.h
stm32f1xx/Include/stm32f103x6.h
stm32f1xx/Include/stm32f103xb.h
stm32f1xx/Include/stm32f103xe.h
stm32f1xx/Include/stm32f103xg.h
stm32f1xx/Include/stm32f105xc.h
stm32f1xx/Include/stm32f107xc.h

stm32f3xx/Include/stm32f301x8.h
stm32f3xx/Include/stm32f302x8.h
stm32f3xx/Include/stm32f302xc.h
stm32f3xx/Include/stm32f302xe.h
stm32f3xx/Include/stm32f303x8.h
stm32f3xx/Include/stm32f303xc.h
stm32f3xx/Include/stm32f303xe.h
stm32f3xx/Include/stm32f318xx.h
stm32f3xx/Include/stm32f328xx.h
stm32f3xx/Include/stm32f334x8.h
stm32f3xx/Include/stm32f358xx.h
stm32f3xx/Include/stm32f373xc.h
stm32f3xx/Include/stm32f378xx.h
stm32f3xx/Include/stm32f398xx.h
```

salkinium: It's all the exact same entry:
```
#define RCC_CFGR_PLLXTPRE_Pos                    (17U)
#define RCC_CFGR_PLLXTPRE_Msk                    (0x1UL << RCC_CFGR_PLLXTPRE_Pos) /*!< 0x00020000 */
#define RCC_CFGR_PLLXTPRE                        RCC_CFGR_PLLXTPRE_Msk         /*!< HSE divider for PLL entry */
#define RCC_CFGR_PLLXTPRE_HSE_PREDIV_DIV1        (0x00000000U)                 /*!< HSE/PREDIV clock not divided for PLL entry */
#define RCC_CFGR_PLLXTPRE_HSE_PREDIV_DIV2        (0x00020000U)                 /*!< HSE/PREDIV clock divided by 2 for PLL entry */
```

salkinium: Oh now I understand, this is an alias bit, and this is only relevant for the F1 devices that do *not* have the /1-/16 divider.

salkinium (approved):
Thanks! This is all rather confusing…

PR #535, opened 2021-01-10
[libstdc++] Update avr libstdc++ submodule to C++20 version

chris-durand:
Fix submodule commit

### DIFF
--- a/src/modm/platform/i2c/at90_tiny_mega/i2c_master.hpp.in
+++ b/src/modm/platform/i2c/at90_tiny_mega/i2c_master.hpp.in
@@ -78,10 +78,10 @@ public:
 
 		// calculate the fractional prescaler value
 		constexpr float pre_part_raw = float(SystemClock::I2c) / ( 2 * baudrate );
-		constexpr float pre_raw = std::floorf(pre_part_raw) < 8 ? 0 : (pre_part_raw - 8) / pre;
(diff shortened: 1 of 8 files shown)


salkinium (approved):

chris-durand:
```
modm/src/modm/platform/i2c/i2c_master.hpp:81:34: error: 'floorf' is not a member of 'std'; did you mean 'floor'?
   81 |   constexpr float pre_raw = std::floorf(pre_part_raw) < 8 ? 0 : (pre_part_raw - 8) / pre;
```

It's correct that it doesn't work. libstdc++ [does not provide `std::*f` C99 math functions](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79700) in the std namespace, this is a gcc bug. Nasty macros in the AVR math.h C header like `#define floorf floor` made `std::floorf` accidentally compile. To fix a math function issue I undefined the macros in avr-libstdc++ `<cmath>` and replaced them with proper functions. Now the code above does not compile anymore.

To keep consistency between platforms I will not add the float variants but fix the modm AVR code to use the proper C++ native overloads.

chris-durand:
The CI passes now, but there is still an [issue to be fixed](https://github.com/modm-io/avr-libstdcpp/pull/8). Some math functions are still missing in namespace std but they are not used within the modm codebase. I'll add a test to trigger the problem.

PR #536, opened 2021-01-12
[ci] MacOS: Enable AVR testing because GCC 10 toolchain is now available

rleh:


.github/workflows/macos.yml
```diff
@@ -9,15 +9,13 @@ jobs:
...
         run: |
-          brew tap modm-ext/arm
-        # brew tap modm-ext/avr
+          brew tap osx-cross/avr
+          brew tap osx-cross/arm
 
       - name: Setup environment - Brew install
         run: |
-          brew install git doxygen boost gcc arm-gcc-bin cmake || true
+          brew install git doxygen boost gcc avr-gcc arm-gcc-bin cmake || true
```
salkinium:
```suggestion
          brew install git doxygen boost gcc avr-gcc@10 arm-gcc-bin cmake || true
```
Unfortunately v10 is not the default.

salkinium: We also need to update the install instructions. 😒

rleh:
Looks to me that the `avr-gcc@10` homebrew package does not provide the `avr-g++` command. Maybe we have to create a symlink `avr-g++-10` -> `avr-g++` or similar?

@salkinium You are the macOS expert, could you please help out?

salkinium:
> Maybe we have to create a symlink avr-g++-10 -> avr-g++ or similar?

No, we have to [detect the version](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/resources/SConscript.in#L20) correctly. We also need to removed the outdated versions of GCC here for C++20 support (and add `gcc-11` while we're at it).

Let's wait a little until homebrew-avr has settled. They just removed the pre-compiled version again, and the formula must now be compiled. Maybe we'll have to fork it to get better control over it.

salkinium:
They only added a bottle for the latest macOS versions, which we should be using in the CI, only I need to compile avr-gcc locally, since I didn't yet update to the lastest version.

Let's see if this works in the CI.

salkinium:
It's keg-only (ie. not in the path by default).
```
==> Caveats
avr-gcc@10 is keg-only, which means it was not symlinked into /usr/local,
because it might interfere with other version of avr-gcc.
This is useful if you want to have multiple version of avr-gcc
installed on the same machine.

If you need to have avr-gcc@10 first in your PATH run:
  echo 'export PATH="/usr/local/opt/avr-gcc@10/bin:$PATH"' >> ~/.zshrc

For compilers to find avr-gcc@10 you may need to set:
  export LDFLAGS="-L/usr/local/opt/avr-gcc@10/lib"
```

salkinium:
Well… that was a shitty experience: Homebrew has a bug, where it "checked the depends" (whatever that means)

```
> ==> No outdated dependents to upgrade!
> ==> Checking for dependents of upgraded formulae...
> ==> Reinstalling 1 broken dependent from source:
> osx-cross/avr/avr-gcc
```

which would replace the `avr-gcc/10.2.0` folder with a `avr-gcc/10.2.0.reinstall` file, which then isn't reinstalled or something bullshitty like that. I disabled that with `export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1`. JFC 🙄

salkinium:
Well… I guess it's not ready yet.
```
Compiling C++·· build/release/modm/src/modm/utils/dummy.o
scons: building terminated because of errors.dyld: Library not loaded: /usr/local/opt/isl/lib/libisl.23.dylib
  Referenced from: /usr/local/Cellar/avr-gcc@10/10.2.0/libexec/gcc/avr/10.2.0/cc1plus
  Reason: image not found
avr-g++: internal compiler error: Abort trap: 6 signal terminated program cc1plus
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://github.com/osx-cross/homebrew-avr/issues> for instructions.
scons: *** [build/release/modm/src/modm/utils/dummy.o] Error 4
```

salkinium:
The issue was that we didn't do `brew update` so the package manager worked with out-of-date information, while the avr-gcc@10 recipe required an up-to-date `isl` installation etc.

rleh (commented):
✅ ✔️ 

Thanks for your effort @salkinium!

PR #537, opened 2021-01-14
[stm32] fix adc clock source configured incorrectly on stm32g4

nicoBruhn:
The adc clock select was configured incorrectly for stm32g4. According to the reference manual it should be as follows:
image

### DIFF
--- a/examples/nucleo_g474re/adc_basic/main.cpp
+++ b/examples/nucleo_g474re/adc_basic/main.cpp
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2021, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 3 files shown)


rleh:
Yes. Same issue I started to fix in #324.

rleh (approved):
Looks fine. Thanks!

Does it work on hardware @nicoBruhn?

nicoBruhn:
I won't have access to a stm32g4 until later this week.

rleh:
Ok, then let's wait with the merge until this has been tested on hardware.
I should also have a Nucleo-G474 lying around, but I will not find time to test it before the weekend.

rleh:
I added an example for the Nucleo_G474RE board and tested it: It works!

rleh (approved):

rleh:
Oops, I was too impatient for TravisCI and merged too early.

salkinium:
Didn't we replace the Travis macOS runner with GH Actions? Did I forget to disable that?

rleh:
The TravisCI job is not for macos but for linux on arm64/Aarch64, because Travis is currenttly the only CI service with arm64 VMs.

PR #538, opened 2021-01-14
[stm32] Fixed TIM2 and TIM5 generated as 32 bit timer on stm32g4

nicoBruhn:
TIM2 and TIM5 are generated as 16 bit timers on stm32g4. They are 32 bit timers.
image


salkinium (approved):
Thanks!

rleh (approved):

PR #540, opened 2021-01-16
[delay] Fix register constraints for subs.n instruction

### LINKED ISSUE #539: Debug build with GCC10.2.0
When building a debug build with GCC 10.2.0, the following assembler error is emitted:
```
/tmp/cc7OQ88c.s: Assembler messages:
/tmp/cc7OQ88c.s:7806: Error: Thumb-2 MUL must not set flags -- `muls.n r2,r2,ip'
/tmp/cc7OQ88c.s:7808: Error: cannot honor width suffix -- `subs.n r4,r4,ip'
/tmp/cc7OQ88c.s:8015: Error: cannot honor width suffix -- `subs.n lr,lr,r1'
/tmp/cc7OQ88c.s:8016: Error: cannot honor width suffix -- `subs.n lr,lr,r4'
```
Release builds do not suffer from the same problem.
It seems to be related to delay_impl.hpp

salkinium:
Fixes #539.

cc @chris-durand @rleh @XDjackieXD

### DIFF
--- a/src/modm/platform/core/cortex/delay_impl.hpp.in
+++ b/src/modm/platform/core/cortex/delay_impl.hpp.in
@@ -33,15 +33,16 @@ extern uint16_t delay_fcpu_MHz;
 
 inline void modm_fastcode delay_ns(uint32_t ns)
 {
-	volatile uint32_t overhead_cycles;
+	volatile uint32_t cycles;
 	// ns_per_loop = nanoseconds per cycle times cycles per loop ({{loop}} cycles)
 	asm volatile (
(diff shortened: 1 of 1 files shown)


chris-durand (approved):
works

salkinium:
(I'm waiting on @XDjackieXD to confirm before merging).

XDjackieXD:
Thank you for the quick fix! I just confirmed that it fixes the problem for both my "production" code and the minimal testing code I made yesterday

rleh (approved):

Issue #485, opened 2020-09-29
Enable D-Cache for Cortex-M7


salkinium:
The Data Cache for Cortex-M7 devices is currently disables, due to a lack of in-depth understanding of cache policies during porting. The Instruction Cache however is enabled, since it only gets read.

See: https://github.com/modm-io/modm/blob/develop/src/modm/platform/core/cortex/startup.c.in#L95-L98

In addition to enabling the cache, the caches must be invalidated manually on certain operations (writing Flash for example).
This is however not just an issue on Cortex-M7, since most Cortex-M devices have some sort of vendor specific cache implementation for their Flash reads, which must also be manually invalided.

cc @mikewolfram

mikewolfram:
I had this already on my list since I was wondering why the I-Cache is enabled, but not the D-Cache.

First time I saw this when porting the FreeRTOS HAL based network interface, but at least for the M7 the macro does nothing to invalidate the cache.

salkinium:
Yes, I just remembered this due to getting confused about DTCM with D-Cache while fixing the DTCM size bug. I wasn't sure during porting what the invalidation required from an modm API perspective, so I just disabled it. Not great, not terrible.

salkinium:
Leaving this here for my future self:
https://blog.feabhas.com/2020/10/introduction-to-the-arm-cortex-m7-cache-part-1-cache-basics/

salkinium:
Also good: https://alexkalmuk.medium.com/cpu-caches-with-examples-for-arm-cortex-m-2c05a339246e

We should enable the D-Cache at least with write-through. Write-back can then be enabled by the user if required.

salkinium:
> Disadvantages of the ‘write-through’ mode are the following:
> - Sequential and frequent access to the same memory address can degrade performance.
> - You still need to do a cache invalidate after the end of DMA operations.
> - There is the “Data corruption in a sequence of Write-Through stores and loads” bug in some versions of Cortex-M7

Well… maybe not. Sounds like we would need support from the DMA API.

mikewolfram:
I recently had an issue on a F765, where I enabled the D cache and Ethernet stopped working. Probably related to the DMA.

salkinium:
Yeah, it seems to require application support to manually invalidate the cache when required (ie. after (during?) DMA transfers), and tbh the user can still enable the DCache in main() so there's no need for modm to enable it (wrongly).

[...]

PR #542, opened 2021-01-26
[board] Nucleo-F767ZI support

rleh:
- [x] It compiles
- [x] Verify clock config (CubeMX)
- [x] Update README 🤦🏽‍♀️ 
- [ ] Tested on hardware

### DIFF
--- a/src/modm/board/nucleo_f767zi/board.hpp
+++ b/src/modm/board/nucleo_f767zi/board.hpp
@@ -1,0 +1,150 @@
+/*
+ * Cpoyright (c) 2021, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_STM32_NUCLEO_F767ZI_HPP
+#define MODM_STM32_NUCLEO_F767ZI_HPP
+
+#include <modm/platform.hpp>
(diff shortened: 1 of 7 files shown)


src/modm/board/nucleo_f767zi/board.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <modm/platform.hpp>
+#include <modm/architecture/interface/clock.hpp>
+#include <modm/debug/logger.hpp>
+/// @ingroup modm_board_nucleo_F767ZI
+#define MODM_BOARD_HAS_LOGGER
+
+using namespace modm::platform;
+
+/// @ingroup modm_board_nucleo_F767ZI
```
salkinium: FYI: The `:docs` generator expects this to be lower case, not sure if Doxypress is case-sensitive and would link this correctly.

chris-durand:
Should we add USB support for this board? We didn't test USB on any F7 so far. Should be quite trivial to do. I could have a look at it later.

salkinium:
I tested it on the F7 Disco board, but could only get the low speed interface working. Not sure why the HS interface didn't work…

salkinium (approved):
LGTM so far, but I don't have the hardware to test this on.

salkinium:
Btw, since this is a Cortex-M7 it may be useful to enable the Data-Cache at some point: #485.

mikewolfram:
The pull request #466 already introduced this board long time ago. But we never got it to the end. That one also includes the example how to use ethernet.

BTW; The board.hpp in this PR looks pretty much like a copy of mine. ;-)

salkinium:
But… don't we want to merge this one instead? The other PR is still a draft.

rleh:
I currently dont't have the Nucleo board available for testing and #466 isn't a draft anymore, so let's merge #466.

PR #548, opened 2021-02-01
Align the tables to copy in linker script to 8 bytes

mikewolfram:
When building with -Os flag there will be a `ldrd` instruction in 
```c++
static inline void
table_copy(const uint32_t *const start, const uint32_t *const end)
```
The requirement of that instruction is that the source address must be aligned to 8 bytes.

Using -Os will not inline the function.

### DIFF
--- a/src/modm/platform/core/cortex/linker.macros
+++ b/src/modm/platform/core/cortex/linker.macros
@@ -102,7 +102,7 @@ TOTAL_STACK_SIZE = MAIN_STACK_SIZE + PROCESS_STACK_SIZE;
 	%% if vector_table_location == "ram"
 		%% do sections.append("vector_table_ram")
 	%% endif
-	.table.copy.intern : ALIGN(4)
+	.table.copy.intern : ALIGN(8)
 	{
 		__table_copy_intern_start = .;
 		LONG(__data_load)


salkinium:
But then we should probably align all tables to 8 bytes? Perhaps we should instead specify the pointer arguments to be 4B aligned, since 8B alignment will waste both flash and RAM memory (~2\*4\*7B = 56B in the worst case), which is definitely more than what is saved by the double load instructions. 

Skipping though the AAPCS I don't see any specific requirement of any pointers to be 8B aligned? The stack pointers should be 8B aligned across compilation units, but that doesn't seem to be a general alignment requirement for all memory pointers?

salkinium:
I mean, specifically pointers are 4B aligned, are we perhaps using the wrong type somewhere to reference the memory section?

chris-durand:
> The requirement of that instruction is that the source address must be aligned to 8 bytes.

  Do you have a reference for that? ~~[Some ARM documentation](https://developer.arm.com/documentation/dui0489/c/arm-and-thumb-instructions/memory-access-instructions/ldr--pc-relative-) specifies "In ARMv7-M, `LDRD` (PC-relative) instructions must be on a word-aligned address." Word-aligned should mean by 4 bytes or am I mistaken?~~

EDIT: Oh sorry, that refers to the instruction alignment not the source data, sorry. :facepalm:

salkinium:
I'm still confused where the compiler gets the idea that it can just assume the pointer to be 8B aligned. How did you stumble on this? Did this cause a hardfault? The compiler cannot know the alignment beforehand, only the linker does.

chris-durand:
[Here](https://developer.arm.com/documentation/dui0473/m/using-the-assembler/address-alignment) it states for ARM-v7:

"For `STRD` and `LDRD`, the specified address must be word-aligned.", so 4 bytes should be enough.

chris-durand:
@mikewolfram What cpu are you compiling this for? Is it a Cortex M0?

salkinium:
[ARMv7-M LDRD](https://developer.arm.com/documentation/ddi0403/latest/) only requires word alignment:


[ARMv6-M](https://developer.arm.com/documentation/ddi0419/latest/) does not have the LDRD instruction at all.

mikewolfram:
It's kind of weird. Yes I was looking into the manuals as well and was expecting a 4 bytes alignment.

But building with -Os screwed up and checking with the debugger I figured out, that in table_copy the addresses loaded were totally wrong. Googled a bit and found a post about GCC producing a misaligned LDRD, so I changed the alignment of the copy table to 8 and suddenly the application started to work.

I'm on a STM32L4 (Cortex-M4). Using GCC 9.2.1.

Trying to reproduce, but no matter what I do, now it seems to work with 4 bytes alignment. 

So I have no clue what happened, why it was resolved when changing alignment.

Sorry.

salkinium:
So I guess that the GCC 9.2.1 was also from xpack? 🤨 (related to #549)

mikewolfram:
Well, it's xpack GCC 9.2.1. But in this project I didn't switch the compiler yet.

Anyway, let's close it since it works.

Issue #554, opened 2021-02-10
STM32F407 Discovery Board - USB


lukh:
Hi ! 
I was trying to run the sample project (examples/generic/usb) with the "modm:disco-f407vg" target.
The board is the stm32f4-discovery obviously, rev MB997C.

I am really annoyed to come and seek help, but this is the first time I work with USB.
I built a basic example code, build with STM32CubeMX, seen as a HID device, it does work (after adding the pullup 1.5k on one of the Dataline), so I doubt the hw is not working properly...

This example should work out of the box right ? No need to implement weak callbacks, or fill descriptors somewhere in the tinyusb sub folder ? (at least, to make this example running)

It is seen as a USB device, but windows tell me a Code43 error, could not get the usb descriptor.

What did I miss ?
Thanks for your help !
Vivien

rleh:
> This example should work out of the box right?

Yes.

I just tested the  `examples/generic/usb` example on `modm:disco-f407vg` target and it works fine without any hardware modification on my (linux) computer:

```bash
$ lsusb
# ...
Bus 003 Device 006: ID cafe:4005 TinyUSB TinyUSB Device

$ dmesg
# ...
[118546.698426] usb 3-1: New USB device found, idVendor=cafe, idProduct=4005, bcdDevice= 1.00
[118546.698428] usb 3-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[118546.698429] usb 3-1: Product: TinyUSB Device
[118546.698430] usb 3-1: Manufacturer: TinyUSB
[118546.698431] usb 3-1: SerialNumber: 004C003B3434511034383735C000FCC0
[118546.699327] cdc_acm 3-1:1.0: ttyACM0: USB ACM device
[118546.699659] usb-storage 3-1:1.2: USB Mass Storage device detected
[118546.700468] scsi host7: usb-storage 3-1:1.2
[118547.760705] scsi host7: scsi scan: INQUIRY result too short (5), using 36
[118547.760712] scsi 7:0:0:0: Direct-Access     TinyUSB  Mass Storage     1.0  PQ: 0 ANSI: 2
[118547.761026] sd 7:0:0:0: Attached scsi generic sg5 type 0
[118547.761329] sd 7:0:0:0: [sde] 16 512-byte logical blocks: (8.19 kB/8.00 KiB)
[118547.761529] sd 7:0:0:0: [sde] Write Protect is off
[...]

PR #555, opened 2021-02-11
fix: configure USB::Id as PullUp in add to OpenDrain

lukh:
After issues with the USB example code (generic/usb) on a STM32F407-Disco board,
Found that I needed to set up Id line as a pullup in addition to Output/OpenDrain (#554 )

(Same as in the tinyusb implementation)

One question remains: why @rleh  was able to run the example without any problems ?
No clue...


salkinium (approved):

rleh:
> One question remains: why @rleh was able to run the example without any problems?

Exciting question ... but unfortunately I don't have an answer to it either.
Maybe there was enough dust on my dev board so that the pin was somehow pulled to 3.3V?

lukh:
Ah ah ah ! I was not expecting that ! thanks for the help anyway !

PR #556, opened 2021-02-11
[driver] Add MMC5603NJ Compass driver

salkinium:
1. Adds a driver for the cheapest and smallest compass on [Digikey: the MMC5603NJ](https://www.digikey.de/product-detail/de/memsic-inc/MMC5603NJ/1267-MMC5603NJCT-ND/10452803).
2. Makes `modm::ui::Animation` virtual, so that it can be used for motor position animation.
3. Fixes an issue with `Apa102::getColor()`

### DIFF
--- a/src/modm/driver/inertial/mmc5603.hpp
+++ b/src/modm/driver/inertial/mmc5603.hpp
@@ -1,0 +1,374 @@
+/*
+ * Copyright (c) 2020, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include <modm/architecture/interface/i2c_device.hpp>
+#include <modm/processing/protothread.hpp>
+#include <modm/math/utils.hpp>
+
+namespace modm
+{
+
+// forward declaration for friending with mmc5603::Data
+template < class I2cMaster >
+class Mmc5603;
+
+/// @ingroup modm_driver_mmc5603
+struct mmc5603
+{
+	/// The MEMSIC device 7-bit device address is 0110000 where the three
+	/// LSB’s are pre-programmed into the MMC5603NJ by the factory.
+	static constexpr
+	uint8_t addr(uint8_t factory=0)
+	{ return 0b0110000 | (factory & 0b111); }
+
+	static constexpr uint8_t ProductID{0x10};
+
+	enum class
+	Register : uint8_t
+	{
+		Xout0 = 0x00,				// Xout[19:12]
+		Xout1 = 0x01,				// Xout[11:4]
+		Yout0 = 0x02,				// Yout[19:12]
+		Yout1 = 0x03,				// Yout[11:4]
+		Zout0 = 0x04,				// Zout[19:12]
+		Zout1 = 0x05,				// Zout[11:4]
+		Xout2 = 0x06,				// Xout[3:0]
+		Yout2 = 0x07,				// Yout[3:0]
+		Zout2 = 0x08,				// Zout[3:0]
+		Tout = 0x09,				// Temperature output
+		Status1 = 0x18,				// Device status1
+		ODR = 0x1A,					// Output data rate
+		InternalControl0 = 0x1B,	// Control register 0
+		InternalControl1 = 0x1C,	// Control register 1
+		InternalControl2 = 0x1D,	// Control register 2
+		ST_X_TH = 0x1E,				// X-axis selftest threshold
+		ST_Y_TH = 0x1F,				// Y-axis selftest threshold
+		ST_Z_TH = 0x20,				// Z-axis selftest threshold
+		ST_X = 0x27,				// X-axis selftest set value
+		ST_Y = 0x28,				// Y-axis selftest set value
+		ST_Z = 0x29,				// Z-axis selftest set value
+		ProductID = 0x39,			// Product ID
+	};
+
+public:
+	enum class
+	Status1 : uint8_t
+	{
+		/// This bit is an indicator of successfully reading its OTP memory
+		/// either as part of its power up sequence, or after an I2C command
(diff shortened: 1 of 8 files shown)


src/modm/driver/inertial/mmc5603.hpp
```diff
@@ -0,0 +1,342 @@
...
+// forward declaration for friending with mmc5603::Data
+template < class I2cMaster >
+class Mmc5603;
+
+/// @ingroup modm_driver_mmc5603
+struct mmc5603
+{
+	static constexpr
+	uint8_t addr(uint8_t factory=0)
+	{ return 0b0110000 | (factory & 0b111); }
```
rleh, line 32: Would you like to add an explanation as a comment?

("The MEMSIC device 7-bit device address is `0110000` where the three LSB’s are pre-programmed into the MMC5603NJ by the factory.")

src/modm/driver/inertial/mmc5603.hpp
```diff
@@ -0,0 +1,342 @@
...
+	};
+	MODM_FLAGS8(Control1);
+
+	enum class
+	Bandwidth : uint8_t
+	{
+		Ms6_6 = 0x00,
+		Ms3_5 = 0x01,
+		Ms2_0 = 0x02,
+		Ms1_2 = 0x03,
```
rleh:
```suggestion
		Time6600us = 0x00,
		Time3500us = 0x01,
		Time2000us = 0x02,
		Time1200us = 0x03,
```

... could be a little bit more self explaining.

src/modm/driver/inertial/mmc5603.hpp
```diff
@@ -0,0 +1,342 @@
...
+		float inline
+		y() const { return swapData(1); }
+
+		float inline
+		z() const { return swapData(2); }
+		///@}
+
+		/// returns the temperature in Celcius
+		int8_t inline
+		t() const { return int16_t(data[9]) * 4/5 - 75; }
```
rleh:
```suggestion
		getTemperature() const { return int16_t(data[9]) * 4/5 - 75; }
```

or

```suggestion
		temperature() const { return int16_t(data[9]) * 4/5 - 75; }
```

src/modm/driver/inertial/mmc5603.hpp
```diff
@@ -0,0 +1,342 @@
...
+	{ return update(Register::InternalControl0, Control0::Take_meas_T); }
+
+	modm::ResumableResult<bool>
+	configureContinuousMode(uint8_t frequency)
+	{
+		RF_BEGIN();
+		if (not RF_CALL(write(Register::ODR, frequency)))
+			RF_RETURN(false);
+		if (not RF_CALL(update(Register::InternalControl1, Register_t(i(Bandwidth::Ms1_2)))))
+		    RF_RETURN(false);
```
rleh: Wouldn't it make sense to be able to configure the bandwidth if the measurement frequency is configurable?

```suggestion
	modm::ResumableResult<bool>
	configureContinuousMode(uint8_t frequency, Bandwidth bandwidth = Bandwidth::Ms1_2)
	{
		RF_BEGIN();
		if (not RF_CALL(write(Register::ODR, frequency)))
			RF_RETURN(false);
		if (not RF_CALL(update(Register::InternalControl1, Register_t(i(bandwidth)))))
		    RF_RETURN(false);
```

src/modm/driver/inertial/mmc5603.lb
```diff
@@ -0,0 +1,45 @@
...
+def init(module):
+    module.name = ":driver:mmc5603"
+    module.description = """
+# MMC5603 3-Axis Digital Magnetometer
+
+The MMC5603NJ is a monolithic complete 3-axis AMR magnetic sensor with on-chip
+signal processing and integrated I2C bus.
+
+It can measure magnetic fields within the full scale range of ±30 Gauss (G),
+with up to 0.0625mG per LSB resolution at 20bits operation mode and 2mG total
```
rleh, line 23: But this driver only used the 16bit mode, correct?

rleh (changes requested):

src/modm/ui/animation/base.hpp
```diff
@@ -23,6 +23,39 @@ namespace modm
...
+	///			`false` if otherwise
+	virtual bool
+	animateTo(T value, uint32_t time) = 0;
+
+	/// @return	`true` if the value has been changed,
+	///			`false` otherwise
+	virtual bool
+	update()
+	{ return false; }
+};
```
chris-durand, line 60: This base class should either have a public virtual or protected non-virtual destructor, see http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-dtor-virtual.

chris-durand (changes requested):

src/modm/driver/inertial/mmc5603.lb
salkinium: No the driver reads all 20-bits, but the Data object only used 16. I change this to use all 20 now.

src/modm/driver/inertial/mmc5603.hpp
salkinium: Added some docs from the datasheet.

src/modm/driver/inertial/mmc5603.lb
rleh: I don't know if that would be useful, so do what works best for you.

rleh (approved):

src/modm/driver/inertial/mmc5603.lb
salkinium: I mean the data is transferred anyways , might as well access them for a few instructions more.
I made the expensive float optional, so the 20-bit int32_t output is going to be cheap…

src/modm/driver/inertial/mmc5603.hpp
salkinium: Yes, also the 1.2ms measurement time was only necessary for 1000Hz operation, however, that does not work with Automatic Set/Reset as enabled in the function, hence the frequency only being 8-bit for a max of 255Hz. If you need 1kHz update rate, then you need to do it manually or update the driver later.

salkinium:
Still have to do:
- [x] Test the changes to `mmc5603::Data` class in hardware.

PR #559, opened 2021-02-15
adding support for usb exemple on STM32F072B-Disco

lukh:
Usb example is running on a disco f072,

### DIFF
--- a/examples/generic/usb/project.xml
+++ b/examples/generic/usb/project.xml
@@ -1,5 +1,6 @@
 <library>
   <extends>modm:blue-pill</extends>
+  <!-- <extends>modm:disco-f072rb</extends> -->
   <!-- <extends>modm:mini-f401</extends> -->
   <!-- <extends>modm:mini-f411</extends> -->
   <!-- <extends>modm:feather-m0</extends> -->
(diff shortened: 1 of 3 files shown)


lukh:
I have one question,
The usb clock is set up by default on the HSI48, instead of the PLL, (which make it work with the nucleo booard)

Should we add the option, in rcc, to change the usb clock ?

I started the work in rcc.hpp.in, l144
`	
// enum class
// UsbClockSource : uint32_t
// {
// 	Hsi48 = RCC_CFGR3_USBSW_HSI48,
// 	Pll = RCC_CFGR3_USBSW_PLLCLK
// };
`

But not sure about how to integrate properly in modm generator.. (and if it is necessary) (but i guess it  it is)

Bests

salkinium (changes requested):
You need to add the `:platform:usb` module as a dependency to the F071 board `module.lb`, otherwise the `modm::platform::UsbFs` won't be available for examples that do not use the USB stack.

salkinium:
> Should we add the option, in rcc, to change the usb clock ?

If I read the [User Manual](https://www.st.com/resource/en/user_manual/dm00099401-discovery-kit-for-stm32f0-series-microcontrollers-with-stm32f072rb-stmicroelectronics.pdf) correctly, the crystal in not populated and neither is SB19 to connect MCO from the ST-Link. So we have to use the internal clock anyways, and the accuracy is probably not better with the PLL from HSI8 compared to the HSI48.


So… meh?

salkinium (approved):
I will squash your commits into one and merge.

lukh:
Great ! Thanks !

PR #565, opened 2021-02-21
[docs] Fix docs for STM32F4 with ethernet

salkinium:
Unfortunately I didn't set the `:freertos:tcp:lan8720a` module enabled correctly and thus F2/F4 devices with ethernet could not resolve the `:platform:eth` module, which is only implemented for F7.

cc @rleh


rleh (approved):

salkinium:
In case you're wondering why the CI is this slow (stm32f4-compile-all takes 40mins vs 9mins before): We ran out of the 2500 free credits per month, we're currently at ~7500~ 7700 credits. It's mostly because I pushed so many PRs this week…

PR #569, opened 2021-02-28
fixed warnings and excessive use of void* and cast

ghost:
- Fixed few warnings in core/build_id.cpp _"ISO C++ forbids flexible array"_ and _"array subscript <unknown> is outside array bounds"_ It actually not allowed in C++ only C. 
- Fixed type casts in heap_table.cpp. Has a problematic design: _"struct table_pool_t"_ is defined in several compilations units.

### DIFF
--- a/src/modm/platform/core/cortex/build_id.cpp
+++ b/src/modm/platform/core/cortex/build_id.cpp
@@ -15,19 +15,17 @@ typedef struct {
     uint32_t namesz;
     uint32_t descsz;
     uint32_t type;
-    uint8_t data[];
+    uint8_t data[30];
 } ElfNoteSection_t;
-extern "C" const ElfNoteSection_t __build_id[];
+extern "C" const ElfNoteSection_t __build_id;
 
 namespace modm
 {
 
-const std::array<uint8_t, 20>&
+std::span<const uint8_t, 20>
 build_id()
 {
-	const uint8_t *const sha1 = &__build_id->data[__build_id->namesz];
-	auto *const hash = reinterpret_cast<const std::array<uint8_t, 20> *>(sha1);
-	return *hash;
+    return std::span<const uint8_t, 20>(&__build_id.data[__build_id.namesz], 20);
 }
 
 }
--- a/src/modm/architecture/interface/build_id.hpp
+++ b/src/modm/architecture/interface/build_id.hpp
@@ -12,7 +12,7 @@
 #pragma once
 
 #include <cstdint>
-#include <array>
+#include <span>
 
 namespace modm
 {
@@ -23,7 +23,7 @@ namespace modm
  * @ingroup modm_architecture_build_id
  */
 [[nodiscard]]
-const std::array<uint8_t, 20>&
+std::span<const uint8_t, 20>
 build_id();
 
 }
--- a/src/modm/platform/core/cortex/heap_table.cpp.in
+++ b/src/modm/platform/core/cortex/heap_table.cpp.in
@@ -14,12 +14,12 @@
 #include <stdint.h>
 #include "heap_table.hpp"
 
-typedef struct
+struct table_pool_t
 {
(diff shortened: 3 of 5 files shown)


src/modm/platform/core/cortex/build_id.cpp
```diff
@@ -15,19 +15,18 @@ typedef struct {
     uint32_t namesz;
     uint32_t descsz;
     uint32_t type;
-    uint8_t data[];
+    uint8_t data;
```
chris-durand: This change does not seem to be an improvement. Dereferencing `&__build_id.data + __build_id.namesz` is clearly undefined behaviour in C++ and very likely also in C. It is a pointer to some location where no actual objects exist. Beforehand it was at least using a gcc language extension for which gcc promises defined behaviour.

src/modm/platform/core/cortex/build_id.cpp
```diff
@@ -15,19 +15,18 @@ typedef struct {
...
 {
 
 const std::array<uint8_t, 20>&
 build_id()
 {
-	const uint8_t *const sha1 = &__build_id->data[__build_id->namesz];
-	auto *const hash = reinterpret_cast<const std::array<uint8_t, 20> *>(sha1);
-	return *hash;
+	return *reinterpret_cast<const std::array<uint8_t, 20> *>(
+        &__build_id.data + __build_id.namesz);
```
chris-durand: This is also undefined behaviour before and after your change. Casting arbitrary memory to `std::array` where no `std::array` has been created is a [strict aliasing violation](https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing) and therefore illegal. A conforming version would be returning a value of `std::span<const uint8_t, 20>` referring to the `uint8_t` data.

chris-durand (changes requested):
Thanks for cleaning up the `void*` mess.

src/modm/platform/core/cortex/build_id.cpp
ghost: Sorry, but span do not support const, it cannot compile. I have tried it once.

chris-durand: Hm, interesting. It works here: https://gcc.godbolt.org/z/5YEjvc

ghost: This is maybe the best solution with build-in types, It will break signedness but it's not a disaster for log output.  
` char data[30];`
`return std::string_view (&__build_id.data[__build_id.namesz], &__build_id.data[__build_id.namesz + 20]);`

ghost: thanks, never noticed the compile time length.

src/modm/platform/core/cortex/build_id.cpp
```diff
@@ -15,19 +15,17 @@ typedef struct {
     uint32_t namesz;
     uint32_t descsz;
     uint32_t type;
-    uint8_t data[];
+    uint8_t data[30];
 } ElfNoteSection_t;
-extern "C" const ElfNoteSection_t __build_id[];
+extern "C" const ElfNoteSection_t __build_id;
```
salkinium, line 20: Does this work in hardware? I found that the [] is necessary to get the right pointer.

ghost: Yes, it's tested, you can do both. You only need a pointer if the struct is an array.

salkinium (approved):
Thanks! I've tested these changes successfully in hardware on the F469 disco with tlsf and hard fault examples.

salkinium:
Let's squash the commits into two. How about:
- [build_id] Use std::span to avoid aliasing issues
- [heap] Fix void* type casts in heap table

rleh (approved):

chris-durand (approved):

PR #572, opened 2021-03-07
Fix/pre lto support

ghost:
LTO support is tested with these changes with success. The test project was reduced 12% in code size with lto support, assert and hardware init tables seems to work fine.
`src/modm/platform/heap/heap_tlsf.cpp` needs to be excluded from lto optimization for now, due to bug in gcc.

**Fixes**
- Used attribute added to linker tables.
- Type mismatches between translation units.
- Struct table_pool_t placed in own header. Was defined more than once.

### DIFF
--- a/src/modm/architecture/utils.hpp
+++ b/src/modm/architecture/utils.hpp
@@ -69,6 +69,9 @@
 	/// Specifies that a variable (or function) lives in a particular section.
 	#define modm_section(s)
 
+	/// Specifies that a variable (or function) lives in a particular section with used flag.
+	#define modm_section_used(s)
+
 	/// Marks a declaration as deprecated and displays a message.
 	#define modm_deprecated(msg)
 
@@ -124,6 +127,7 @@
 	#define modm_likely(x)			__builtin_expect(!!(x), 1)
 	#define modm_unlikely(x)		__builtin_expect(!!(x), 0)
 	#define modm_section(s)			__attribute__((section(s)))
+	#define modm_section_used(s)	__attribute__((section(s), used))
 	#define modm_fallthrough		__attribute__((fallthrough))
 	#define modm_noreturn			__attribute__((noreturn))
 	#define modm_warn_unused_result	__attribute__((warn_unused_result))
--- a/src/modm/platform/core/cortex/hardware_init.hpp
+++ b/src/modm/platform/core/cortex/hardware_init.hpp
@@ -18,25 +18,25 @@
 /// Call `function` during boot process.
 /// @hideinitializer
 #define MODM_HARDWARE_INIT(function) \
-	modm_section(".hardware_init") \
+	modm_section_used(".hardware_init") \
 	void* MODM_CONCAT(__modm_hardware_init_ptr_, function) = (void*)&function
 
 /// Call `function` during boot process with a unique name.
 /// @hideinitializer
 #define MODM_HARDWARE_INIT_NAME(name, function) \
-	modm_section(".hardware_init") \
+	modm_section_used(".hardware_init") \
 	void* MODM_CONCAT(__modm_hardware_init_ptr_, name) = (void*)&function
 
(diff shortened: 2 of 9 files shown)


salkinium:
Please add the LTO flags to the common.py, so I can test this.

ghost:
> Please add the LTO flags to the common.py, so I can test this.

This was only meant as a bug fix to eliminate problems that LTO don't like. In my head LTO should be a option in project.xml not a hard-coded thing, i have not looked at that part yet.

ghost:
LTO flags is added.  -flto -fno-fat-lto-objects
It seems to work without any special flags for`src/modm/platform/heap/heap_tlsf.cpp` at lest with GCC version 10.2.1 As the CI shows, old compilers is not a option. 
For deterministic builds a -lrandom-seed flag is needed. [explanation here](https://blog.conan.io/2019/09/02/Deterministic-builds-with-C-C++.html)

salkinium:
I cannot get this to work with arm-none-eabi-gcc 10.2.1 on macOS. 

I get the same errors are the CI regarding undefined `_sbrk_r`, even when not even using malloc anywhere.
```
/usr/local/Cellar/arm-gcc-bin/10-2020-q4-major/bin/../lib/gcc/arm-none-eabi/10.2.1/../../../../arm-none-eabi/bin/ld: nano-mallocr.c:(.text._malloc_r+0xaa): undefined reference to `_sbrk_r'
```
Even when explicitly defining `_sbrk_r` in code, or when using the `modm:platform:heap` module with allocator newlib. Linker function wrapping also does not work for me.

ghost:
> I cannot get this to work with arm-none-eabi-gcc 10.2.1 on macOS.
> 
> I get the same errors are the CI regarding undefined `_sbrk_r`, even when not even using malloc anywhere.
> 
> ```
> /usr/local/Cellar/arm-gcc-bin/10-2020-q4-major/bin/../lib/gcc/arm-none-eabi/10.2.1/../../../../arm-none-eabi/bin/ld: nano-mallocr.c:(.text._malloc_r+0xaa): undefined reference to `_sbrk_r'
> ```
> 
> Even when explicitly defining `_sbrk_r` in code, or when using the `modm:platform:heap` module with allocator newlib. Linker function wrapping also does not work for me.

I was not a feature/lto but a fix to remove ugly code and undefined behavior LTO complained about. Take it or leave it.

salkinium:
> I was not a feature/lto but a fix to remove ugly code and undefined behavior LTO complained about. Take it or leave it.

I'm fine with the common heap_table declaration changes, but I don't understand how I can evaluate the LTO changes if I or the CI cannot build with LTO enabled. What compiler are you using? Are you using AVRs?

ghost:
> I'm fine with the common heap_table declaration changes, but I don't understand how I can evaluate the LTO changes if I or the CI cannot build with LTO enabled. What compiler are you using? Are you using AVRs?

I'm using ARM and I was tricked by the build system, by accident `make` was switched with `ninja` and I discovered something was cached somehow. A file or two like `heap_tlsf.cpp` still needs special flags. I'm not sure how to solve this with lbuild the correct way.
One is that simple examples works, but LTO is probably not mature enough to everything, another project I'm currently working  on shows that inline assembler also can give issues. We need more experience and a easy way to enable it in lbuild.

salkinium:
I'm sorry, but I've decided not to merge this.

I think the `table_pool_t` struct header has some value, but it's a super private API that should only ever be accessed by the official `heap_table.hpp` API and I'm ok with the duplication in the fault module, since it reduces dependencies.

I don't think LTO is going to happen without a toolchain fix, and I couldn't even get a vanilla newlib (ie. without function wrapping) to work, since the linker won't find `_sbrk_r` for some reason (I too got warnings about assembly). I would be very suspicious of any firmware produced with LTO, and double check that all sections are in the right place (via the listing for example).

PR #574, opened 2021-03-09
STUSB4500 USB-PD chip driver

rleh:
Minimal but working implementation of a [STUSB4500](https://www.st.com/en/interfaces-and-transceivers/stusb4500.html) driver.

Not actually tested on the STM32F072-Disco board, but on real hardware (because STUSB4500 package is too small for jumper wires).

cc @hshose

### DIFF
--- a/src/modm/driver/usb/stusb4500.hpp
+++ b/src/modm/driver/usb/stusb4500.hpp
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2021, Henrik Hose
+ * Copyright (c) 2021, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef MODM_STUSB4500_HPP
+#define MODM_STUSB4500_HPP
+
+#include <modm/architecture/interface/i2c_device.hpp>
+
+namespace modm
+{
+
+/// @ingroup modm_driver_stusb4500
+struct stusb4500
+{
+public:
+	enum class
+	Register : uint8_t
+	{
+		RdoRegStatus = 0x91,
+		// PortStatus1 = 0x0E,
+		// CcStatus = 0x11,
+		// TxHeaderLow = 0x51,
+		// PdCommandCtrl = 0x1A,
+		DpmPdoNumb = 0x70,
+		DpmSnkPdo1 = 0x85,
+		DpmSnkPdo2 = 0x89,
+		DpmSnkPdo3 = 0x8D,
+	};
+
+	struct RdoRegStatusData {
+		uint32_t MaxCurrent; /// [mA]
+		uint32_t OperatingCurrent; /// [mA]
+		uint8_t ObjectPos;
+	};
+};
+
+/**
+ * @ingroup modm_driver_stusb4500
+ * @author Henrik Hose, Raphael Lehmann
+ */
+template<class I2cMaster>
+class Stusb4500 : public stusb4500, public modm::I2cDevice< I2cMaster, 4 >
+{
+public:
+	Stusb4500(uint8_t address=0x28):
+		modm::I2cDevice<I2cMaster,4>(address)
+	{};
+
+	/**
+	 * Configures a PDO.
+	 *
+	 * Note: According to the USB standard the first PDO (1) always has to set
+	 *  voltage to 5000 [mV].
+	 *
+	 * @param	pdoNumber Which PDO to configure, range 1 to 3.
+	 * @param	voltage	  in mV
+	 * @param	current	  in mA
+	 */
+	modm::ResumableResult<bool>
+	configurePdo(uint8_t pdoNumber, uint32_t voltage, uint32_t current)
+	{
+		RF_BEGIN();
+
+		if((pdoNumber < 1) || (pdoNumber > 3))
+			RF_RETURN(false);
+
+		RF_CALL(readRegister(DpmSnkPdos[pdoNumber - 1], buffer2.data(), 4));
+
+		current /= 10;
+		voltage /= 50;
+
+		buffer2[0] = current;
+		buffer2[1] = ((current >> 8) & 0b11) | ((voltage & 0b11'1111 ) << 2);
+		buffer2[2] = (buffer2[2] & 0b1111'0000) | ((voltage >> 6) & 0b1111);
+
+		RF_END_RETURN_CALL(updateRegister(DpmSnkPdos[pdoNumber - 1], buffer2.data(), 4));
+	}
+
+	/**
+	 * Sets the valid PDO.
+	 *
+	 * @param	pdoNumber Which PDO to set valid, range 1 to 3.
+	 */
+	modm::ResumableResult<bool>
+	setValidPdo(uint8_t pdoNumber)
+	{
+		RF_BEGIN();
+
+		if((pdoNumber < 1) || (pdoNumber > 3))
+			RF_RETURN(false);
+
+		RF_END_RETURN_CALL(updateRegister(Register::DpmPdoNumb, &pdoNumber, 1));
+	}
+
+	/**
+	 * Retrieve RDO status from STUSB4500 chip.
+	 *
+	 * @return	RdoRegStatusData object. Typically only
+	 *  RdoRegStatusData::MaxCurrent [mA] is of interest.
+	 */
+	modm::ResumableResult<RdoRegStatusData>
+	getRdoRegStatus()
+	{
+		RF_BEGIN();
+
+		RF_CALL(readRegister(Register::RdoRegStatus, buffer2.data(), 4));
+
+		// MaxCurrent: Bits 9..0; OperatingCurrent: Bits 19..10; ObjectPos: Bits 30..28
+		rdoRegStatusData.MaxCurrent = buffer2[0] | ((buffer2[1] & 0b11) << 8);
+		rdoRegStatusData.MaxCurrent *= 10;
+		rdoRegStatusData.OperatingCurrent = ((buffer2[1] & 0b1111'1100) >> 2) | ((buffer2[2] & 0b1111) << 6);
+		rdoRegStatusData.OperatingCurrent *= 10;
+		rdoRegStatusData.ObjectPos = (buffer2[3] & 0b0110'0000) >> 5;
+
+		RF_END_RETURN(rdoRegStatusData);
+	}
+
+public:
+	modm::ResumableResult<bool>
+	readRegister(Register reg, uint8_t* output, size_t length)
+	{
+		RF_BEGIN();
+		buffer[0] = uint8_t(reg);
+		this->transaction.configureWriteRead(buffer.data(), 1, output, length);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+	modm::ResumableResult<bool>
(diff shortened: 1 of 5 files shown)


hshose:
Just tested the module in the [Otter Iron](https://github.com/rleh/Otter-Iron-PRO/commit/d9141d640ab11d4f4c81a2a534ad875b0c0c94ae) with STM32F072 - works as expected.

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,160 @@
...
+		buffer.at(0) = uint8_t(reg);
+		std::memcpy(&buffer.at(1), data, length);
+		this->transaction.configureWriteRead(buffer.data(), length + 1, nullptr, 0);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+private:
+	static constexpr std::array<Register, 3> DpmSnkPdos {Register::DpmSnkPdo1, Register::DpmSnkPdo2, Register::DpmSnkPdo3};
+	std::array<uint8_t, 5> buffer;
+	std::array<uint8_t, 4> buffer2;
```
salkinium, line 151: Why are we using `std::array` for such small buffers? The [`.at` method](https://en.cppreference.com/w/cpp/container/array/at) does bounds checking (and throws an exception), but we only ever call it with fixed numbers. It'll probably be optimized away, but it's still kinda weird?

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,160 @@
...
+private:
+	template<uint32_t length>
+	modm::ResumableResult<bool>
+	readRegister(Register reg, uint8_t* output)
+	{
+		RF_BEGIN();
+		buffer[0] = uint8_t(reg);
+		this->transaction.configureWriteRead(buffer.data(), 1, output, length);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
```
salkinium, line 134: Let's pass length as an argument, the RF functions do generate a lot of code.

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,160 @@
...
+		rdoRegStatusData.MaxCurrent = buffer2.at(0) | ((buffer2.at(1) & 0b11) << 8);
+		rdoRegStatusData.MaxCurrent *= 10;
+		rdoRegStatusData.OperatingCurrent = ((buffer2.at(1) & 0b1111'1100) >> 2) | ((buffer2.at(2) & 0b1111) << 6);
+		rdoRegStatusData.OperatingCurrent *= 10;
+		rdoRegStatusData.ObjectPos = (buffer2.at(3) & 0b0110'0000) >> 5;
+
+		RF_END_RETURN(rdoRegStatusData);
+	}
+
+private:
```
salkinium: Make it public, so that devs can implement more functionality externally.

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,160 @@
...
+	 * Configures a PDO.
+	 *
+	 * Note: According to the USB standard the first PDO (1) always has to set
+	 *  voltage to 5000 [mV].
+	 *
+	 * @param	pdoNumber Which PDO to configure, range 1 to 3.
+	 * @param	voltage	  in mV
+	 * @param	current	  in mA
+	 */
+	inline modm::ResumableResult<bool>
```
salkinium: No need for inline in a templated class/struct.

src/modm/driver/usb/stusb4500.hpp
salkinium: Also: buffer2 is unnecessary?

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,160 @@
...
+		buffer[0] = uint8_t(reg);
+		this->transaction.configureWriteRead(buffer.data(), 1, output, length);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+	template<uint32_t length>
+	modm::ResumableResult<bool>
+	updateRegister(Register reg, uint8_t* data)
+	{
+		static_assert(length <= (std::tuple_size<decltype(buffer)>::value - 1), "buffer too small");
```
salkinium: Pass length as an argument and use `length = std::min(length, 5);`

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include <modm/architecture/interface/i2c_device.hpp>
+
+namespace modm
+{
+
+/// @ingroup modm_driver_stusb4500
+struct stusb4500
+{
+protected:
```
salkinium: public

src/modm/driver/usb/stusb4500.lb
```diff
@@ -0,0 +1,35 @@
...
+# STUSB4500
+
+## Standalone USB PD controller for power sinking devices
+
+The STUSB4500 is a USB power delivery controller for USB power sink devices.
+See https://www.st.com/en/interfaces-and-transceivers/stusb4500.html
+
+The STUSB4500 is connected via I2C to the microcontroller.
+
+This driver only implements basic functionality by now.
```
salkinium:
```suggestion
This driver only implements basic functionality for now.
```

src/modm/driver/usb/stusb4500.lb
```diff
@@ -0,0 +1,35 @@
...
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = ":driver:stusb4500"
+    module.description = """\
+# STUSB4500
```
salkinium:
```suggestion
# STUSB4500 USB PD controller
```

Crams in more info in one line on `lbuild discover` ;-P

salkinium (changes requested):
Thanks! I'm not convinced about the `std::array` usage.

src/modm/driver/usb/stusb4500.hpp
ghost: A span would be appropriate. And why use tuple_size. std::array is constexpr.

`static_assert(length <= (buffer.size() - 1), "buffer too small");`

src/modm/driver/usb/stusb4500.hpp
ghost: Everyone is taught not to use raw arrays in modern C++. In this case [] operators will be ok. no need for bounds check. Use iterators, algorithms (std::copy) then possible, they are fast and safer.

Auto deduction can be used here
`static constexpr std::array DpmSnkPdos {Register::DpmSnkPdo1, Register::DpmSnkPdo2, Register::DpmSnkPdo3};`

src/modm/driver/usb/stusb4500.hpp
rleh: Nope. `array.size()` does not work in static_assert() because `buffer` is not constexpr. Try it out yourself...

rleh: > use `length = std::min(length, 5);`

No, because I don't like silent failures. Run-time checking the `length` parameter and abort with `return false` now.

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,157 @@
...
+	readRegister(Register reg, uint8_t* output, size_t length)
+	{
+		RF_BEGIN();
+		buffer[0] = uint8_t(reg);
+		this->transaction.configureWriteRead(buffer.data(), 1, output, length);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+	modm::ResumableResult<bool>
+	updateRegister(Register reg, uint8_t* data, size_t length)
```
chris-durand: `data` should be a pointer to const here.

salkinium (approved):
Nice!

src/modm/driver/usb/stusb4500.hpp
```diff
@@ -0,0 +1,157 @@
...
+		buffer[0] = uint8_t(reg);
+		this->transaction.configureWriteRead(buffer.data(), 1, output, length);
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+	modm::ResumableResult<bool>
+	updateRegister(Register reg, const uint8_t* data, size_t length)
+	{
+		RF_BEGIN();
+		if(length > (std::tuple_size<decltype(buffer)>::value - 1))
```
ghost, line 140: You wrong. [Explain this](https://godbolt.org/z/Wzh178) and read the specifications.

rleh:
MacOS CI is drunk. Why don't we use a Docker image in MacOS?

ghost (commented):
You wrong about static asserts and arrays . [Explain this](https://godbolt.org/z/Wzh178) and read the [specifications](https://en.cppreference.com/w/cpp/container/array/size).

src/modm/driver/usb/stusb4500.hpp
rleh:
```
<source>: In function 'int main()':
<source>:9:37: error: passing 'volatile std::array<unsigned char, 5>' as 'this' argument discards qualifiers [-fpermissive]
    9 |     static_assert(4 <= (buffer.size() - 1), "buffer too small");
      |                                     ^
In file included from <source>:1:
/opt/compiler-explorer/gcc-10.2.0/include/c++/10.2.0/array:175:7: note:   in call to 'constexpr std::array<_Tp, _Nm>::size_type std::array<_Tp, _Nm>::size() const [with _Tp = unsigned char; long unsigned int _Nm = 5; std::array<_Tp, _Nm>::size_type = long unsigned int]'
  175 |       size() const noexcept { return _Nm; }
      |       ^~~~
Compiler returned: 1
```

rleh: See https://godbolt.org/z/58MTr7

chris-durand: @jasa It does not work here because `buffer` is a non-static class member. The `this` pointer is not a constant expression here. Therefore, calling a method on a non-static member never can be: https://godbolt.org/z/bfnh1f

If std::array had a static size method it would work.

salkinium: LOL, this is like the Uno reverse card of reviews: No, @jasa, how 'bout **YOU** WRONG!1!! 😂🤣 C++ is weird…

rleh: ~But here we could use `buffer.size()` since this check now is 100% runtime.~

Using `std::tuple_size<decltype(buffer)>::value` generates faster and smaller code: https://godbolt.org/z/GKMMY8

grafik

ghost: I'm blind, sorry.

chris-durand: We don't compile with `-O0` here :wink:

rleh: Yeah, compiled with `-Os` or `-O3` for ARM there is no difference: https://godbolt.org/z/3G4Tr8

grafik

rleh: Should I change it to `buffer.size()` (and thus trigger the CI again)?

chris-durand: I can't tell you what the CI can take, but `buffer.size()` is by far more readable and idiomatic code.

PR #575, opened 2021-03-11
Added support for os << color::RgbT<>

TomSaw:
Hi,
Thanks for developing and maintaining this incredible framework! I'm studying and learning from the code back to the xpcc days.
This is my first contribution. Please let me know, if i've missed a thing.

### DIFF
--- a/src/modm/driver/color/tcs3414.hpp
+++ b/src/modm/driver/color/tcs3414.hpp
@@ -22,8 +22,8 @@
 
 #include <stdint.h>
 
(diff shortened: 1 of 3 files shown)


salkinium:
Thanks, I'm glad you enjoy it, even though we went through so many changes since xpcc times!

TomSaw:
And with the improvements since then, the philosophy has been inscribed further!
I love to learn and apply the power of C++. Takes one some time to get started but the effort pays out a thousand times.

PR #577, opened 2021-03-11
[ui] Support to stream color::HsvT like color::RgbT

TomSaw:
Added support for streaming HSL-Colors:
- Implementation is copy&pasted from existing modm::color::RgbT<>


salkinium (approved):
Thanks!

rleh (approved):

PR #19, opened 2018-05-19
Add GpioGroup for efficient access to GPIO registers

salkinium:
This provides a base class for providing arbitrary sets of pins with an efficient access method to the GPIO port registers. All accesses are bundled, so ideally only one RMW access per port is required instead of per pin. A quick review of the assembly listing shows near optimal access for `Group::setOutput()`.
Whenever the implementation is as efficient as the single Gpio class implementation, it is used instead to limit code duplication.

`GpioGroup` does not understand pin order, it is just performing the *same* action for all pins in a more efficient way.
`SoftwareGpioPort` will extend `GpioGroup` to provide methods for efficient reading and writing of *ordered* pin groups.

This requires C++17 due to `constexpr if` and fold expressions used, see #18.

- [x] Implement and test for STM32
- [x] Implement for AVR
- [x] Test for AVR
- [x] Document classes
- [x] Expose traits of GpioPort
- [x] Update CMSIS and enable C11, C++17
- [x] Enable AvrDude Tool again
- [x] Move CMSIS & C++17 commit to separate PR
- [x] Move AvrDude tool out of this PR
- [x] Unset mask in ASCR on disconnect
- [x] Unittests for STM32
- [x] Unittests for STM32F1
- [x] Unittests for AVR

PR #580, opened 2021-03-17
[driver] Add best practice driver for soft graycode / quadrature decoding

TomSaw:
Hi ✋, missed the gentle communication - here's my next 💢 

Many modern MCUs have builtin quadrature-decoding support. However encoders alias rotary knobs kept being so usefull in small projects with small chips. Raphael commited the complement in 2019: modm/driver/encoder/encoder_output.hpp. So, let's don't discriminate the roots:

It's a port of well known, best practice solution: [mikrocontroller.net/articles/Drehgeber](https://www.mikrocontroller.net/articles/Drehgeber).

### Things to discus

1. I've redundantly added static_assert checking < UnderlyingType> in two methods: update() and getValue(). I found, this makes the compiler printing the assert-failure-string always on top. There may be a better technique.

2. getValue() could have been specialized for each value in < DIVISION >. Because of the difference of just 3 tiny lines, i found the it better using a switch-statement.

3. Alternative implemention of getValue() as functor: _operator()()_. To integrate the latest changes to a local variable v, one would type: _v  += encoder();_ instead of _v += encoder.getValue();_

4. Following an example i've seen in another modm-module, i've put the credits to Peter Dannegger's ANSI-C foundation into *encoder.lb* ... but this may not be sufficient.


TomSaw:
Example follows as soon as a future in the codebase is approved.

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+template<typename PH_A, typename PH_B, modm::encoder::Division_t DIVISION, typename UnderlyingType>
+UnderlyingType
+modm::Encoder<PH_A, PH_B, DIVISION, UnderlyingType>::getValue()
+{
+	static_assert(std::is_signed<UnderlyingType>::value and std::is_integral<UnderlyingType>::value,
+				  "UnderlyingType must be signed integral");
+	cli();
```
salkinium: There's no cli/sei on Cortex-M. Use `modm::atomic::Lock _;` as a properly scoped atomic section.

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,49 @@
+#ifndef MODM_ENCODER_HPP
+#error "Don't include this file directly, use 'encoder.hpp' instead!"
+#endif
+
+template<typename PH_A, typename PH_B, modm::encoder::Division_t DIVISION, typename UnderlyingType>
+inline void
+modm::Encoder<PH_A, PH_B, DIVISION, UnderlyingType>::update()
+{
+	static_assert(std::is_signed<UnderlyingType>::value and std::is_integral<UnderlyingType>::value,
+				  "UnderlyingType must be signed integral");
```
salkinium: > I've redundantly added static_assert checking < UnderlyingType> in two methods: update() and getValue(). I found, this makes the compiler printing the assert-failure-string always on top. There may be a better technique.

I would favor this in the class definition. I'm not sure optimizing for compiler specific error formatting is a good idea for C++. All hope is lost there…

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,49 @@
...
+	// Tare on power-on
+	UnderlyingType raw = 0;
+	UnderlyingType diff;
+
+	if (PH_A::read()) raw = 3;
+	if (PH_B::read()) raw ^= 1;  // convert gray to binary
```
salkinium: Separate reads may tear.
If you turn them into a `modm::platform::SoftwareGpioPort<PH_A, PH_B>`, then you get atomic reads for free at least for pins on the same port. You can assert on the number of ports `modm::platform::SoftwareGpioPort<PH_A, PH_B>::number_of_ports()` to force this atomicity.

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,49 @@
...
+			enc_delta = 0;
+			break;
+		case encoder::Division::TWO:
+			enc_delta = val & 1;
+			val >>= 1;
+			break;
+		case encoder::Division::FOUR:
+			enc_delta = val & 3;
+			val >>= 2;
+	}
```
salkinium: > getValue() could have been specialized for each value in < DIVISION >. Because of the difference of just 3 tiny lines, i found the it better using a switch-statement.

This looks like it could be generalized by passing an integer instead of an enum?

src/modm/driver/encoder/encoder.hpp
```diff
@@ -0,0 +1,72 @@
...
+struct encoder
+{
+	typedef enum Division
+	{
+		ONE,
+		TWO,
+		FOUR
+	} Division_t;
+};
+
```
salkinium: `/// @ingroup modm_driver_encoder` to make the Doxygen work right.

src/modm/driver/encoder/encoder.lb
```diff
@@ -0,0 +1,27 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = ":driver:encoder"
+    module.description = """
+Software graycode ported from C-Code of Peter Dannegger:
+https://www.mikrocontroller.net/articles/Drehgeber.
```
salkinium: > Following an example i've seen in another modm-module, i've put the credits to Peter Dannegger's ANSI-C foundation into encoder.lb ... but this may not be sufficient.

I think it's fine, we've haven't copied the code literally and we're even adapting/improving it.

src/modm/driver/encoder/encoder.lb
```diff
@@ -0,0 +1,27 @@
...
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = ":driver:encoder"
+    module.description = """
+Software graycode ported from C-Code of Peter Dannegger:
```
salkinium: As a note: This is interpreted as Markdown and the first line should be the heading, which is displayed in `lbuild discover` and used in all docs. Something like:

```
    module.description = """
# Graycode Encoder Input

{Description} blabla, use pins on the same port, blabla, Inspired by Peter: See url.
    """
```

salkinium (changes requested):
> Alternative implemention of getValue() as functor: operator()(). To integrate the latest changes to a local variable v, one would type: v += encoder(); instead of v += encoder.getValue();

Meh, we don't use operator() for anything else, so it would not be consistent with the rest of modm. I would actually prefer the getter, since it's explicit, otherwise it looks like a function call.

src/modm/driver/encoder/encoder.lb
TomSaw: What's wrong here? Not the proper doxygen outlining or what?

salkinium: It's not wrong, it's just that the modm.io module reference (and doxygen) take the first line of this description as the heading, and you don't want the first sentence to be used as the heading:
Example: https://modm.io/reference/module/modm-driver-encoder-output/ where the first line of the module description is `# Encoder Output`.

src/modm/driver/encoder/encoder.hpp
salkinium: Just FYI: The doxygen groups are automatically generated by the `modm:docs` module and they replicate the module structure, so `modm:driver:encoder` becomes `@defgroup modm_driver_encoder` `@addtogroup modm_driver` automatically.
Example: `modm:driver:encoder.output` https://docs.modm.io/develop/api/stm32f103zgt7/group__modm__driver__encoder__output.html

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,45 @@
...
+	modm::atomic::Unlock _;
```
salkinium: No, modm::atomic::Lock does the unlock automatically via RAII, hence creating the class on the stack as a variable here called `_`.

TomSaw:
Easy fixes done... let's generalize the Dividor

src/modm/driver/encoder/encoder.lb
TomSaw: Got this :)

src/modm/driver/encoder/encoder_impl.hpp
TomSaw: Very nice. That's my next and afterwards generalization for Divisor.

TomSaw: This works, but is it sexy??

auto read = modm::platform::SoftwareGpioPort<PH_A, PH_B>::read();
if (read & 0b10) raw = 3;
if (read & 0b01) raw ^= 1;  // convert gray to binary

salkinium: > but is it sexy??

Yes, it's an atomic read and I wanna make love to it!1!!

Btw, you can do `using Phases = modm::platform::SoftwareGpioPort<PH_A, PH_B>;` in the class definition for added sexyness.

TomSaw: Sweet. This beautyfied also connect()

Phases::setInput(::Gpio::InputType::PullUp);

rleh:
What is the use-case for this driver?
Typically I would wire an the A and B signals of an encoder to Ch1 and Ch2 of any STM32 "Advanced" or "General purpose" timer and let the timer hardware handle it.
Like here: https://github.com/roboterclubaachen/micro-motor/blob/master/src/hardware_rev2.hpp#L368-L390

With this driver you have to call `update()` at a very high frequency (>100kHz with some high-resolution encoders), otherwise some encoder ticks will be lost unnoticed.

TomSaw:
#

> What is the use-case for this driver?
> Typically I would wire an the A and B signals of an encoder to Ch1 and Ch2 of any STM32 "Advanced" or "General purpose" timer and let the timer hardware handle it.
> Like here: https://github.com/roboterclubaachen/micro-motor/blob/master/src/hardware_rev2.hpp#L368-L390
> 
> With this driver you have to call `update()` at a very high frequency (>100kHz with some high-resolution encoders), otherwise some encoder ticks will be lost unnoticed.

Hi Raphael. On small MCUs there's no such support. I've noted this in the opening message: "Many modern MCUs have builtin quadrature-decoding support. However encoders alias rotary knobs kept being so usefull in small projects with small chips."

Another case is, when you want an encoder on the fly, just for a quick debugging and no sufficient timer-pins are free.
Also there could be many encoders.. build some music-equipment f.e. where the hardware-decoders may not be sufficient.

rleh:
I don't know how this works on other microcontrollers (AVR, SAM, ....).
Maybe some of them don't have hardware features like STM32, so this "bitbang" driver is useful there.

In my opinion this driver belongs to the "bitbang" category (like e.g. the [`spi.bitbang` driver](https://github.com/modm-io/modm/blob/develop/src/modm/platform/spi/bitbang/module.lb)) and should be named accordingly.

rleh:
> On small MCUs there's no such support.

Which MCU are you using?

TomSaw:
> > On small MCUs there's no such support.
> 
> Which MCU are you using?

Various ATmega... still. I like them somehow

TomSaw:
> I don't know how this works on other microcontrollers (AVR, SAM, ....).
> Maybe some of them don't have hardware features like STM32, so this "bitbang" driver is useful there.
> 
> In my opinion this driver belongs to the "bitbang" category (like e.g. the [`spi.bitbang` driver](https://github.com/modm-io/modm/blob/develop/src/modm/platform/spi/bitbang/module.lb)) and should be named accordingly.

That's a thing i would give in your hands!

Edit: Keeping it in :driver,  encoder and encoder_output hang up together so the familiarity is exposed. encoders docs will tell you, that it's a **Soft** version and anyone who cares for the performance-drawbacks knows, there's hardware support.

rleh:
I propose to call this driver `modm:driver:encoder_input.bitbang`.
Someday someone will take the effort to redesign the (horrible) STM32 timer HAL (see #275, #337) and will add a STM32 specific `modm:driver:encoder_input` driver...

Unsure if this driver belong to `src/modm/driver/` or `src/modm/platform/`. What do you think @salkinium?

And maybe we should rename `modm:driver:encoder.output` to `modm:driver:encoder_output.bitbang` for consistency (but this would break backward compatibility)

src/modm/driver/encoder/encoder.hpp
```diff
@@ -0,0 +1,79 @@
...
+template<typename PH_A, typename PH_B, encoder::Division_t DIVISION,
+		 typename UnderlyingType = int8_t>
```
chris-durand: Since we have C++20 concepts available you could use `std::signed_integral` to constrain the type instead of the static assert:
```suggestion
		 std::signed_integral UnderlyingType = int8_t>
```

src/modm/driver/encoder/encoder.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+	inline void
+	connect()
+	{
+		Phases::setInput(::Gpio::InputType::PullUp);
+
+		// Tare power-on state
+		modm::delay(10us);
+		UnderlyingType raw = 0;
+		auto read = Phases::read();
```
chris-durand: This variable could be const
```suggestion
		const auto read = Phases::read();
```

src/modm/driver/encoder/encoder.hpp
```diff
@@ -0,0 +1,79 @@
...
+		if (read & 0b01) raw ^= 1;
+		raw_last = raw;
+	}
+
+	// Call @1kHz for manual movement
+	inline void
+	update();
+
+	UnderlyingType
+	getValue();
```
chris-durand:
```suggestion
	getValue() const;
```

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+template<typename PH_A, typename PH_B, modm::encoder::Division_t DIVISION, typename UnderlyingType>
+inline void
+modm::Encoder<PH_A, PH_B, DIVISION, UnderlyingType>::update()
+{
+	// Tare on power-on
+	UnderlyingType raw = 0;
+	UnderlyingType diff;
+
+	auto read = Phases::read();
```
chris-durand:
```suggestion
	const auto read = Phases::read();
```

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,46 @@
...
+	if (read & 0b10) raw = 3;
+	if (read & 0b01) raw ^= 1;  // convert gray to binary
+
+	diff = raw_last - raw;
```
chris-durand: This can be declared and initialized on one line instead of declaring it uninitialized:
```suggestion
	const auto diff = raw_last - raw;
```

chris-durand (changes requested):

src/modm/driver/encoder/encoder.hpp
salkinium: Doesn't work, because `enc_delta` is written back.

TomSaw: exactly

src/modm/driver/encoder/encoder.hpp
TomSaw: Very nice. Didn't know that.

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -2,27 +2,25 @@
...
 {
-	// Tare on power-on
+	const auto read = Phases::read();
+	
 	UnderlyingType raw = 0;
-	UnderlyingType diff;
-
-	auto read = Phases::read();
 	if (read & 0b10) raw = 3;
 	if (read & 0b01) raw ^= 1;  // convert gray to binary
```
TomSaw: I'm not sure, if this isn't already optimized by the compiler. Using a lookup-table instead of bit-ops?
```
	uint8_t raw_lookup[] = {0,1,3,2};
	uint8_t raw = raw_lookup[Phases::read()];
```

salkinium: Hard to say, I generally check with `scons listing` then searching for the function in the .lss file and comparing the generated assembly.

TomSaw: Ok, let me go for this.

src/modm/driver/encoder/encoder.hpp
chris-durand: Sorry, didn't notice. I didn't look carefully enough.

src/modm/driver/encoder/encoder_impl.hpp
```diff
@@ -0,0 +1,44 @@
...
+			break;
+		case encoder::Division::TWO:
+			enc_delta = val & 1;
+			val >>= 1;
+			break;
+		case encoder::Division::FOUR:
+			enc_delta = val & 3;
+			val >>= 2;
+	}
+	::modm::atomic::Unlock unlock;
```
salkinium: No, just remove this line, you don't need it. 

The `modm::atomic::Lock` is a class with a constructor, which disables IRQs on construction, ie. line 27, and restores IRQs on destruction when the object leaves scope, here on line 43. [Here's the implementation for AVR](https://github.com/modm-io/modm/blob/develop/src/modm/platform/core/avr/atomic_lock_impl.hpp#L30-L48). Instead of calling `sei()` in the destructor, it restores SREG, so that it doesn't re-enable interrupts that have previously been (intentionally) disabled for example in nested atomic locks.

This also works if you have multiple returns or throw exceptions in your function. [This idiom is known as RAII in C++](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization).

salkinium:
> Unsure if this driver belong to src/modm/driver/ or src/modm/platform/. 

In `modm/driver`. I would even argue that the bitbang I2C and SPI classes belong in modm/driver, since they only depend on the GPIO *interface*, not even the implementation.

> `modm:driver:encoder_input.bitbang` and `odm:driver:encoder_output.bitbang`

Yes, let's rename them.

src/modm/driver/encoder/encoder_impl.hpp
TomSaw: Okay forget about it: Initializing the list comsumes lot's more ticks than the bit-ops.

I've measured the period of update() without signal changes using simple Pin::set() / pin::reset():

ATmega328p@16Mhz:
With BitOps: 1.5us
With Lookup: 2.312us

src/modm/driver/encoder/encoder_impl.hpp
TomSaw: Now i understood the mechanism.

src/modm/driver/encoder/encoder_impl.hpp
salkinium: Oh wow, I would have guessed the exact opposite!

TomSaw: Double checked again... and jepp it's true.

TomSaw:
> > Unsure if this driver belong to src/modm/driver/ or src/modm/platform/.
> 
> In `modm/driver`. I would even argue that the bitbang I2C and SPI classes belong in modm/driver, since they only depend on the GPIO _interface_, not even the implementation.
> 
> > `modm:driver:encoder_input.bitbang` and `odm:driver:encoder_output.bitbang`
> 
> Yes, let's rename them.

modules are renamed.

How will the classes be named?
BitBangEncoderInput
BitBangEncoderOutput

Same for the preamble structs ... this should become encoder_[input/output] too, right?

salkinium:
> BitBangEncoder[In,Out]put, encoder_[input/output]

Yeah… kind of an eyesore, but it's consistent.

TomSaw:
> > BitBangEncoder[In,Out]put, encoder_[input/output]
> 
> Yeah… kind of an eyesore, but it's consistent.

It entlightens one fair enough about the hidden gems from software-encoding.

src/modm/driver/encoder/encoder_impl.hpp
TomSaw: > > getValue() could have been specialized for each value in < DIVISION >. Because of the difference of just 3 tiny lines, i found the it better using a switch-statement.
> 
> This looks like it could be generalized by passing an integer instead of an enum?

In my opinion that's the last thing left. Some for tomorrow or someone else. I want to progress the rich input from todays session first. Thanks again, such a nice trip!

TomSaw:
Finished a generic implementation for DIVISOR. It's light and clean but feel free to correct me.

I also wanted to trait out UnderlyingType from DIVISOR, but found that it's size also depends on how often getIncrement() is called / how much increments needs to be stocked.

src/modm/driver/encoder/encoder_input.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * @tparam PH_B          	Second modm::platform::Gpio pin to input the encoder signal.
+ * @tparam DIVISION		 	Resolution: count of phasechange for one increment.
+ * @tparam UnderlyingType	Must be signed integer
+ */
+template<
+	typename PH_A,
+	typename PH_B,
+	uint8_t DIVISION = 4,
+	std::signed_integral UnderlyingType = int8_t
+	>
```
TomSaw: For consistency: In bitbang_encoder_output the Gpio parameters are named <PinA, PinB, ...>

What's better or more modm'ish?

TomSaw (commented):
Consistence of Gpio template parameters

src/modm/driver/encoder/encoder_input.hpp
salkinium: I thought PH stood for Phase? Alternatively you can just call it SignalA and SignalB. We don't have a good naming convention for that…

TomSaw: Yes it stands for Phase and i like it. But in bitbang_encoder_**output** its called PinA, PinB

I want to unify them.

TomSaw: Ok i simple name them all Signalx.

src/modm/driver/encoder/encoder_input.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * @tparam SignalB          	Second modm::platform::Gpio pin to input the encoder signal.
+ * @tparam DIVISION		 	Resolution: count of phasechange for one increment.
+ * @tparam UnderlyingType	Must be signed integer
+ */
+template<
+	typename SignalA,
+	typename SignalB,
+	uint8_t DIVISION = 4,
+	std::signed_integral UnderlyingType = int8_t
+	>
```
TomSaw: Im busy with another thing right now.
Later, i definitely add a static_cast and check, if the size of UnderlyingType can at least handle DIVISION

src/modm/driver/encoder/encoder_input_impl.hpp
```diff
@@ -16,28 +18,22 @@ modm::BitBangEncoderInput<PH_A, PH_B, DIVISION, UnderlyingType>::update()
...
-			break;
-		case bitbang_encoder_input::Division::FOUR:
-			enc_delta = val & 3;
-			val >>= 2;
-	}
+	UnderlyingType val = delta;  // counts since last call
+
+	delta = val & (DIVISION - 1);
+	constexpr uint8_t shift = std::log2(DIVISION);
+	val >>= shift;
```
salkinium: Isn't this just `val /= DIVISION`?
Alternatively you can interpret DIVISION as a power of two and generate the mask via `(1ul << DIVISION) - 1` and trivially shift.

TomSaw: Alternatively you can interpret DIVISION as a power of two...
I think the **interface generalizes better**, when DIVIDER is "number of signal-flips to  count one increment", When DIVIDER is linear related to rotation of the encoders shaft  instead of Power of two thing ;)

TomSaw: Isn't this just val /= DIVISION?

My understanding confirms:
- `val >>= log2(DIVISOR)` _equals_ `val /=DIVISION`
- Shifting by log2 results in the division

But `val />> DIVISOR` does not work properly:

- When counting up, it behaves the same like `val >>= log2(DIVISOR)`
- ❗ But when counting down, the algorithm swallows decrements❗

I'm digging for the explanation. A good thing to learn more about twos complement.

TomSaw: Found the leak in `var /= DIVISOR` when counting down that was not happening with `val >>= log2(DIVISOR)` when counting down. FYI:

### Let's have

int8_t val = -4;
const uint8_t DIVISOR = 8;

### The difference is the rounding behaviour:

**`var /= DIVISOR` always rounds up.**
```
MODM_LOG_INFO << "val / DIVISOR = " << val / DIVISOR << modm::endl;; // outputs 0
```

**`val >>= log2(DIVISOR)` rounds _"smart"_,  away from 0!**
```
constexpr uint8_t shift = std::log2(DIVISOR);
MODM_LOG_INFO << "val >> log2(DIVISOR) = " << (val >> shift) << modm::endl; // outputs -1
```

salkinium: Ah, of course… signed division is not the same as shifting! Ok, let's leave it like this.

TomSaw: Again FYI, Here's why:
signed int has two representations of 0:
0000 -> +0
1111 -> -0

We can say negatives have an offset of -1 against positives.
The compiler and/or ALU knows this and handles the offset when dividing signed ints.
**But shift divisions are dump and therefore round down for negatives.**

TomSaw:
The module-names are **prefixed with the bitbang-classifier**: `encoder_[in/out]pu.bitbang`
I found that filenames on the other hand are **postfixed with the classifier**: f.e. `bitbang_i2c_master.hpp`

- For the encoder-files, i **followed the postfix** example

That's it from my side 😏 All implemented ✅ and tested ✅

src/modm/driver/encoder/bitbang_encoder_input.hpp
```diff
@@ -0,0 +1,69 @@
...
+		 std::signed_integral DeltaType = int8_t>
+class BitBangEncoderInput
+{
+	static_assert(std::popcount(POSTSCALER) == 1,
+				  "POSTSCALER must be an integer to basis 2 and not 0: 1, 2, 4, 8, 16, ...");
+	static_assert(POSTSCALER <= std::numeric_limits<DeltaType>::max(),
+				  "DeltaType is to small for POSTSCALER.");
+
+	using Phases = modm::platform::SoftwareGpioPort<SignalA, SignalB>;
+
```
salkinium, line 46:
```suggestion
	static_assert(Phases::number_of_ports() == 1, "Signal A/B must be on the same GPIO port to prevent signal tearing!");
```
Otherwise the atomic read properties may not be true.

src/modm/driver/encoder/bitbang_encoder_output_impl.hpp
```diff
@@ -10,79 +10,72 @@
  */
 // ----------------------------------------------------------------------------
 
-#ifndef	MODM_ENCODER_OUTPUT_HPP
-#	error	"Don't include this file directly, use 'encoder_output.hpp' instead!"
+#ifndef MODM_BITBANG_ENCODER_OUTPUT_IMPL_HPP
```
salkinium:
```suggestion
#ifndef MODM_BITBANG_ENCODER_OUTPUT_HPP
```

salkinium (approved):
Very nice! Let's merge once the CI passes.

src/modm/driver/encoder/bitbang_encoder_output_impl.hpp
TomSaw: oops... did not check building the encoder_output example.
Next time i'll do 😅

src/modm/driver/encoder/bitbang_encoder_input.hpp
TomSaw: Very nice mechanism.. Damn i ♥ love ♥ the way you've implemented the low level stuff

salkinium: I actually think it's just `Phases::number_of_ports` without a function call… I don't even remember the code I wrote.

salkinium: > Damn i ♥ love ♥ the way you've implemented the low level stuff

Thanks, half of the credit goes to C++17 with the relaxed constexpr rules:
https://github.com/modm-io/modm/pull/19#issuecomment-390767555

src/modm/driver/encoder/bitbang_encoder_output_impl.hpp
TomSaw: 😫 Just discovered this **Commit suggestion** button below reviews. It could have been so easy, haha.

chris-durand (approved):

src/modm/driver/encoder/bitbang_encoder_input.hpp
TomSaw: > I actually think it's just `Phases::number_of_ports` without a function call… I don't even remember the code I wrote.

It is and has already been concidered in the last commits ;)

TomSaw: > Thanks, half of the credit goes to C++17 with the relaxed constexpr rules:

Thanks for the interresting background!

TomSaw: What's the magic spell to [get these shift maps](https://github.com/modm-io/modm/pull/19#issuecomment-390767555)?

TomSaw:
If and how can i prevent in [unittests-linux-generic](https://app.circleci.com/pipelines/github/modm-io/modm/1170/workflows/0980f573-d906-47e1-a8f5-a14659491e2c/jobs/28342) ?

> 
> Please synchronize the modm documentation:
>     $ python3 tools/scripts/synchronize_docs.py
> and then commit the results!
> Exited with code exit status 1

rleh:
```bash
python3 tools/scripts/synchronize_docs.py
```

TomSaw:
> ```shell
> python3 tools/scripts/synchronize_docs.py
> ```

Shame on me...

TomSaw:
CI stucks with...

`ERROR: In 'modm:platform:uart:2':
OSError: [Errno 28] No space left on device: '/root/project/build/test_all_5ilvyiug/modm/src/modm/platform/uart/uart_hal_2_impl.hpp'`.

... and i wonder how - if any - i could have messed this up 🙄

salkinium:
Ugh, CircleCI has made some recent changes in their free plan and now it's making us so much trouble. I'll try to rerun it, or we'll simply merge it without the F7 job. You're contributions didn't touch the HAL anyways.

src/modm/driver/encoder/bitbang_encoder_input.hpp
salkinium: Here's the full example: https://github.com/modm-io/modm/blob/develop/examples/stm32f469_discovery/ports/main.cpp

TomSaw:
You're #588 hunting for an alternative...  No worries on my side :)

rleh (approved):

rleh:
Why did you close this pull request? @TomSaw

salkinium:
Thanks @TomSaw for sticking around for our review, it's a very beatiful driver now!

TomSaw:
The pleasure is on my side. I've learned a lot and feels good to dive deeper into the catacombs of contribution.

TomSaw:
Excuse the little disasters due to my github-newbieness @rleh , do you mean, it was a bad thing to delete the encoder branch after it has been merged into develop?

salkinium:
> The pleasure is on my side. I've learned a lot and feels good to go deeper into the catacombs of contribution.

That's good to hear, because I was worried we scared you off by descending on you like a plague of judgemental locusts…

> after it has been merged into develop?

Emphasis on **AFTER** 😜 I think you mistook my force push to your encoder branch with a merge into develop, but I wanted to check the CI *after* the rebase, which was good, because I messed it up and had to fix it.

TomSaw:
> I was worried we scared you off by descending on you like a plague of judgemental locusts

I think any generic concept benefits from perfection. Forget the Pareto-principle here: The expensive fine-works pay out squared^squared^squared when you later build application**s!** out of the generic puzzle-pieces.

I also find lot's of satisfaction, making one thing perfect rather than going for quantity. You have my fullest respect when judging my lines cause it's for the good!   - Amen

PR #582, opened 2021-03-17
[docs] Enable dark-mode on modm.io and docs.modm.io

salkinium:
- [x] Enable darkmode for mkdocs based on user preference.

- [x] Enable darkmode for Doxypress based on user preference.

cc @rleh


salkinium:
Neither Doxypress nor Doxygen have built-in dark themes, so I'm using my extremely advanced CSS skillz to hack something together based on what other people have done… 😒

salkinium:
I've copied over the MIT licensed CSS from https://github.com/maJerle/doxygen-dark-theme, which already gets us 95% there for the Doxypress theme, and it is somewhat beautiful even.

~However, there are some differences that I'm too dumb to fix: the header tab bar and the alternating list css. Nothing I tried changes these colors (even in the original css), I must be doing something very wrong. I'll continue trying, but this is really not my jam.~

salkinium:
Never mind, I discovered `!important`.

rleh:
One question: Is the dark theme the only theme available or is there a switch to enable/disable it?

salkinium:
Mkdocs changes based on your web browser preference.
But the Doxypress and docs homepage CSS settings are definitely hardcoded.

salkinium:
Lemme check how the user preference CSS thingy works, maybe it's easy to adapt.

salkinium:
> Is the dark theme the only theme available or is there a switch to enable/disable it?

It now supports both by just wrapping it into: `@media (prefers-color-scheme: dark) { ... }`.
Let me check if I can add an explicit switch somewhere.

salkinium:
> Let me check if I can add an explicit switch somewhere.

Apparently there's a toggle for MkDocs, but it's only available for 10€/month sponsorships. Meh.
https://squidfunk.github.io/mkdocs-material-insiders/setup/changing-the-colors/#color-palette-toggle

For Doxypress CSS we have to implement a toggle ourselves… but I'm not motivated to do it. I would assume that people using dark mode in their OS setting are fine with getting dark mode websites.
https://jouni.kantola.se/blog/2020-02-29/dark-mode-toggle/

salkinium:
I've tested this on Safari with both prefers-color-scheme light/dark.

rleh (approved):

salkinium:
Thanks, I'll test it first with a develop-dark-theme branch before I mess something this big up.

rleh (changes requested):
grafik

Black text on dark-grey background. Should be changed...

salkinium:
Hm, not digging the dark blue or whatever that is either.

rleh:
The SVG graphics are also an issue...
grafik

salkinium:
I fixed the PNG (yes!) graphics by inverting them. It looks ok, but most importantly it doesn't require two sets of graphics (which I have no clue how to get).



The search field is in a iframe, and Doxypress did not add the custom CSS to that iframe. 🤦‍♂️



So I had to hack it with some Javascript. It's not elegant but it works. I'll open an issue with Doxypress later.

salkinium:
Found another issue with the dependency graphs in the module docs:



Good thing we have the `develop-*` build rule!

salkinium:
This is fixed now too:

salkinium:
Also inverted these and removed the inaccessible CircleCI reference.

rleh:
> PNG (yes!) graphics

Wait, all this graphics are PNG? Oh no...

rleh (approved):

PR #584, opened 2021-03-18
[bugfix] Add a short delay to loop in HD77480::writeCRGAM()

TomSaw:
Sometimes modm is to fast for old Hardware.
As soon as i add this little delay, the custom chars look like intended instead of random.

- I think the hack is acceptable cause CGRAM is typically written only once on initialisation.

### DIFF
--- a/src/modm/driver/display/hd44780_base_impl.hpp
+++ b/src/modm/driver/display/hd44780_base_impl.hpp
@@ -177,6 +177,7 @@ modm::Hd44780Base<DATA, RW, RS, E>::writeCGRAM(uint8_t character, uint8_t *cg)
 	while(not writeCommand(SetCGRAM_Address | (character << 3)))
 		;
 	for (std::size_t ii = 0; ii < 8; ++ii) {
+		modm::delay(50us);
 		writeRAM(cg[ii]);
 	}
 	return true;


src/modm/driver/display/hd44780_base_impl.hpp
```diff
@@ -177,6 +177,7 @@ modm::Hd44780Base<DATA, RW, RS, E>::writeCGRAM(uint8_t character, uint8_t *cg)
 	while(not writeCommand(SetCGRAM_Address | (character << 3)))
 		;
 	for (std::size_t ii = 0; ii < 8; ++ii) {
+        modm::delay_ms(10);
```
salkinium: taaaaabs ;-P

TomSaw: This was older edit without clang... forgive me ohhhh uC-whisperer

salkinium:
Where is this ext/dlr/scons-build-tools coming from all the time?

TomSaw:
Edit: I don't really understand 😩
Edit2: On observation

salkinium (approved):
Thanks!

TomSaw:
Don't merge... i'll test check again, how many ms or even us are really required...

TomSaw:
Try & Error turned out, 30us are enough, so i've adjusted to 50us.

PR #163, opened 2019-02-12
WIP: Better AVR support

asmfreak:
This PR is work in progress for better AVR support in modm.
It will, eventually, fix the following problems:

1. [ ] Timer/Counters support (#157 )
   * [X] module implemented
   * [X] tc8 implemented
       * [ ] with docs
       * [ ] with examples
   * [ ] tc8-async implemented
       * [ ] with docs
       * [ ] with examples
   * [ ] tc16 implemented
       * [ ] with docs
       * [ ] with examples
   * [ ] tc10 implemented
       * [ ] with docs
       * [ ] with examples
2. [ ] More and better boards support
    * [X] *duino Uno (atmega328p)
    * [ ] *duino Leonardo (atmega32u4)
    * [ ] *duino Mega2560 (atmega2560)
    * [ ] Digispark with attiny85

PR #586, opened 2021-03-19
[avr] Support avrdude fusebits via options

TomSaw:
**This has no priority!**
But when it flows, it flows. 🏄

I'm not 100% intimate with the inner workings of lbuild. My implementation of fusebit-options for AVRs is rather copy-catted than wisely coded. It works pretty good but again let me know, if i missed a thing, or two. 

- validate_fuse may be over-engineered
- Is :build:avrdude the right scope for the fusebits? They do not depend on the programmer itself!

Edit: in src/modm/board/arduino_nano/board.xml i've set the default fusebits for arduino-nano, so you can lbuild one of the nano-examples and quickly play around.


tools/build_script_generator/common.py
```diff
@@ -154,11 +160,20 @@ def common_avrdude_options(env):
         option_baudrate = env.collector_values(":build:default.avrdude.baudrate", option_baudrate)[0]
     if not option_options:
         option_options = env.collector_values(":build:default.avrdude.options", option_options)[0]
+    if not option_hfuse:
+        option_options = env.collector_values(":build:default.avrdude.hfuse", option_hfuse)[0]
+    if not option_lfuse:
+        option_options = env.collector_values(":build:default.avrdude.lfuse", option_lfuse)[0]
+    if not option_efuse:
+        option_options = env.collector_values(":build:default.avrdude.efuse", option_efuse)[0]
```
salkinium: This only makes sense if there are some default collectors for these too.

salkinium:
Ah, wait a second. The idea was to program the fuses via the ELF file mechanism, see this: https://www.nongnu.org/avr-libc/user-manual/group__avr__fuse.html
That's why this isn't implemented yet.

I guess we should document this somewhere… 😬😳

TomSaw:
Ohhh No 😓

salkinium:
The issue is that avrdude doesn't support a config file, like `openocd.cfg`, so I didn't want to pipe all options through lbuild, since that just bloats it all up, only the absolutely necessary options.

The question is, whether you can create a `modm_weak FUSES = { ...defaults... };` in the `board.cpp` file, which can be overwritten by an application?

TomSaw:
There are various fractals named 'fuses' in the python-files.  I thought fusebits have not been re-implemented after a bigger refactorisation.

Do you need thes fractals for the new implementation?

salkinium:
> Do you need thes fractals for the new implementation?

"fractals for the new implementation"? 🤨

The FUSES struct should already work with the code today. I remember testing it in hardware.

TomSaw:
Ok, this goes beyond my skillz - would have to learn more background.

- I can put the fuses into a struct in my local main.cpp?
- How does 'scons program-fuses' recognize them?

salkinium:
> I can put the fuses into a struct in my local main.cpp?

Yes, like this:
```
FUSES = 
{
    0xFF, // .low
    0xDA, // .high
    0xFD, // .extended
};
```

> How does 'scons program-fuses' recognize them?

The `FUSES` is a macro expanding to `__fuse_t __fuse __attribute__((section (".fuse"))`, which puts the `__fuse` object into the linkersection `.fuse`, which is not part of the application binary, it is only for avrdude, which can find and read the values from the ELF file. We just need to tell avrdude, which fuses we want programmed (all by default for scons).

TomSaw:
Yea got this  :).. you have even the #Makros:

FUSES = 
{
    .low = LFUSE_DEFAULT,
    .high = HFUSE_DEFAULT,
    .extended = EFUSE_DEFAULT,
};


but 'scons program-fuses' still produces:
`avrdude -V -p m328p -c avrisp2 -P usb -b 115200 -U flash:w:build/release/Sorter.elf -U hfuse:w:build/release/Sorter.elf -U lfuse:w:build/release/Sorter.elf -U efuse:w:build/release/Sorter.elf`

salkinium:
> but 'scons program-fuses' still produces:

Yes, that's correct. The format is `section name : operation : source` with the source being the ELF file.

TomSaw:
Cool this sound interesting. I'll check it out.

salkinium:
See https://www.nongnu.org/avrdude/user-manual/avrdude_4.html#Option-Descriptions `-U memtype:op:filename[:format]`

salkinium:
The interesting question is, if you can define a weak default struct in the BSP `board.cpp` and *overwrite* it in the application:

in `modm/src/board/board.cpp` (`modm_weak` = weak linkage):
```c
modm_weak FUSES = 
{
    LFUSE_DEFAULT, // .low
    HFUSE_DEFAULT, // .high
    EFUSE_DEFAULT, // .extended
};
```
(or non-default values if required for the specific board).

Then in `main.cpp` to overwrite these (no `modm_weak`, ie. strong linkage):
```c
FUSES = 
{
    0xFF, // .low
    0xDA, // .high
    0xFD, // .extended
};
```

TomSaw:
Ok, there's plenty of stuff i need to read + understand before digging deeper! Gimme some ⏰ and i'll be back in the conversation.

salkinium:
> if you can define a weak default struct in the BSP board.cpp and overwrite it in the application

Yes, just tested that and it works as expected. However, this is completely undocumented right now, I'm adding docs for `scons program-fuses`. I guess that would be the right place, since people will hopefully check the build system for the right commands.

> Gimme some ⏰ and i'll be back in the conversation.

Sure, no stress, happy weekend!

TomSaw:
I'm longing for your docs.
Happy weekend back!

TomSaw:
Ahhh 😏 Weekend. Project-Time!

salkinium:
I'm closing this in favor of the ELF mechanism. Sorry for the lost work!

TomSaw:
No prob. Learned some about lbuild. And FUSES and ELF.

For me is newland, so honestly, thanks for the clarification and good luck with the ELF.

I'll throw an eye on #163 as a compensation.

PR #587, opened 2021-03-19
[avr] Add default fuses to BSPs and documentation for scons program-fuses

salkinium:
AVR fuse support is done via ELF section: https://www.nongnu.org/avr-libc/user-manual/group__avr__fuse.html
Alternative to #586.

- [x] Tested in hardware using the Arduino Nano.

cc @TomSaw

### DIFF
--- a/src/modm/board/al_avreb_can/board_fuses.cpp
+++ b/src/modm/board/al_avreb_can/board_fuses.cpp
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2021, Niklas Hauser
+ *
+ * This file is part of the modm project.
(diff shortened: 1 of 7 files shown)


rleh (approved):
I would prefer this over #586. (sorry @TomSaw)

Code looks fine to me, but I have not tested it on hardware.

TomSaw:
I would prefer this too! Was just my humble try to implement a feature, that has low prio to me.

salkinium:
I've checked the default fuse values with the AVR fuse calculator and they all seem sensible, so I'm going to merge this when the CI passes.

PR #589, opened 2021-03-21
Replace CircleCI with Github Actions (self-hosted runners)

rleh:


rleh:
Duplicate of #588, but I want to test self-hosted runners in parallel...

salkinium:
dafuq why is this so fast?!?

rleh:
grafik

rleh:
grafik

The last 15 minutes there are not enough jobs to fully load the server:
grafik

salkinium:
I'm not sure if there was a miscommunication, but I'm still very interested in the speedup of self-hosting, but it's perhaps too much work?

rleh:
Using own runners means a lot of effort for us/me and brings only small performance improvements.
Therefore I will not continue this for now, I think there are currently more interesting features (e.g. hwut... 😆 ).

salkinium:
> Using own runners means a lot of effort for us/me and brings only small performance improvements.

Yes, that's also the impression I got. Thanks for trying this out though.

> I think there are currently more interesting features (e.g. hwut... 😆 ).

HECK YEAH!!!

PR #596, opened 2021-03-27
[example] Fixed 2 bugs in example/arduino_nano/color

TomSaw:
Fixed 2 bugs in example/arduino_nano/color
+
Took the chance to beautify other color-examples by adding Descriptor "HSV: " to output-stream.

### DIFF
--- a/examples/arduino_nano/color/main.cpp
+++ b/examples/arduino_nano/color/main.cpp
@@ -51,6 +51,8 @@ public:
 		// Setup WaitTime to further slow down samplerate
 		PT_CALL(sensor.setWaitTime(modm::tcs3472::WaitTime::MSEC_2_4));
 
+		// Dummy read required
+		PT_CALL(sensor.readColor());
 		// Fetch one sample ...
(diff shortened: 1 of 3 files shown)


salkinium:
Ok, but let's squash them all into one commit please.

TomSaw:
Squashing done.

salkinium (approved):

salkinium:
Just as additional info: the CI has changed and will not auto-cancel builds anymore (we didn't find an easy way to fix this). This means for every commit you push the CI runs for ~12mins @ 13 jobs. Since we're running on the free plan, we're limited to 20 parallel jobs, so if you push a lot of commits you'll have to wait for a long time until the jobs become available.

salkinium:
You can hopefully see the jobs queue here: https://github.com/modm-io/modm/actions

TomSaw:
`You can hopefully see the jobs queue`
I see them and got this.
From now on i'll push only when there's a real benefit from it.
Being aware of squashing commits will also reduce clutter.

PR #599, opened 2021-03-30
[ci] Use workflow_dispatch on label creation

salkinium:
Tested here: https://github.com/salkinium/modm/pull/1 

Unfortunately we'll have to remove and add the `ci:hal` label to restart the CI, however, it's only a small line in the version history and I don't expect we need to do it too often (considering the quick hal jobs).

cc @rleh

### DIFF
--- a/.github/workflows/compile-all.yml
+++ b/.github/workflows/compile-all.yml
@@ -1,13 +1,12 @@
 name: Compile HAL for all devices
 
 on:
-  pull_request_review:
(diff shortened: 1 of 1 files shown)


salkinium:
Oh, nice it already works from inside this PR!

salkinium:
Removing and adding the label also restarts the tests, which is great!

rleh (approved):

PR #600, opened 2021-03-30
Nucleo-L452RE BSP & 2046 Touch driver & minor fixes

rleh:
- Nucleo-L452RE BSP
  - incl. Blink and Graphics-Touch (Ili9341 + TSC2046) examples
- 2046 Touch driver
- Make `uint8_t*` `const` in  `SpiMaster*::transfer(rx*, tx*, len)`
- Ili9341: Fix width/height swap in clear() function
- Example blue-pill/graphics: enabe backlight, so that the display actually displays something

cc @hshose

---

While creating this PR I noticed the [`ads7843`-driver](https://modm.io/reference/module/modm-driver-ads7843/) which should be compatible with the `*2046` touchscreen drivers...

### DIFF
--- a/examples/nucleo_l452re/graphics_touch/main.cpp
+++ b/examples/nucleo_l452re/graphics_touch/main.cpp
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2021, Henrik Hose
+ * Copyright (c) 2021, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/driver/display/ili9341_spi.hpp>
+
+#include <modm/driver/touch/touch2046.hpp>
+
+using namespace Board;
+
+namespace tft
+{
+	using DmaRx = Dma1::Channel2;
+	using DmaTx = Dma1::Channel3;
+	using Spi = SpiMaster1_Dma<DmaRx, DmaTx>;
+	//using Spi = SpiMaster1;
+	using Cs = modm::platform::GpioC8;
+	using Sck = modm::platform::GpioA5;
+	using Miso = modm::platform::GpioA6;
+	using Mosi = modm::platform::GpioA7;
+	using DataCommands = modm::platform::GpioC5;
+	using Reset = modm::platform::GpioC6;
+	using Backlight = modm::platform::GpioC9;
+}
+
+modm::Ili9341Spi<
+	tft::Spi,
+	tft::Cs,
+	tft::DataCommands,
+	tft::Reset,
+	tft::Backlight
+> tftController;
+
+namespace touch
+{
+	using Spi = SpiMaster2;
+	using Cs = modm::platform::GpioB3;
+	using Sck = modm::platform::GpioB13;
+	using Miso = modm::platform::GpioB14;
+	using Mosi = modm::platform::GpioB15;
+	//using Interrupt = modm::platform::GpioA10;
+}
+
+modm::Touch2046<touch::Spi, touch::Cs> touchController;
+
+
+int
+main()
+{
+	using namespace modm::literals;
+	Board::initialize();
+	Dma1::enable();
+
+	tft::Spi::connect<
+		tft::Sck::Sck,
+		tft::Miso::Miso,
+		tft::Mosi::Mosi>();
+	tft::Spi::initialize<SystemClock, 40_MHz>();
+	tftController.initialize();
+	tftController.enableBacklight(true);
+
+	touch::Spi::connect<
+		touch::Sck::Sck,
+		touch::Miso::Miso,
+		touch::Mosi::Mosi>();
+	touch::Spi::initialize<SystemClock, 2500_kHz>();
+	modm::touch2046::Calibration cal{
+		.OffsetX = -11,
+		.OffsetY = 335,
+		.FactorX = 22018,
+		.FactorY = -29358,
+		.MaxX = 240,
+		.MaxY = 320,
+		.ThresholdZ = 500,
+	};
+	touchController.setCalibration(cal);
+
+	LedGreen::set();
+
+	tftController.setColor(modm::glcd::Color::black());
+	tftController.setBackgroundColor(modm::glcd::Color::red());
+	tftController.clear();
+
+	tftController.setFont(modm::font::ArcadeClassic);
+	tftController.setColor(modm::glcd::Color::white());
+	tftController.setCursor(10,12);
+	tftController << "(10,10)";
+	tftController.setCursor(150,202);
+	tftController << "(150,200)";
+	tftController.drawLine(5, 10, 15, 10);
+	tftController.drawLine(10, 5, 10, 15);
+	tftController.drawLine(145, 200, 155, 200);
+	tftController.drawLine(150, 195, 150, 205);
+
+	tftController.setColor(modm::glcd::Color::black());
+	tftController.setFont(modm::font::Ubuntu_36);
+
+	int16_t X = 0;
+	int16_t Y = 0;
+	int16_t Z = 0;
+
+	while (true)
+	{
+		LedGreen::set();
+
+		std::tie(X, Y, Z) = RF_CALL_BLOCKING(touchController.getRawValues());
+		tftController.setColor(modm::glcd::Color::red());
+		tftController.fillRectangle({30, 50}, 90, 115);
+		tftController.setColor(modm::glcd::Color::black());
+		tftController.setCursor(0, 50);
+		tftController << "X=" << X;
+		tftController.setCursor(0, 90);
+		tftController << "Y=" << Y;
+		tftController.setCursor(0, 130);
+		tftController << "Z=" << Z;
+
+		tftController.setColor(modm::glcd::Color::red());
+		tftController.fillRectangle({30, 220}, 120, 35);
+		tftController.setColor(modm::glcd::Color::black());
+		if(RF_CALL_BLOCKING(touchController.isTouched())) {
+			std::tie(X, Y) = RF_CALL_BLOCKING(touchController.getTouchPosition());
+			tftController.setCursor(5, 220);
+			tftController << "> (" << X << "," << Y << ")";
+		}
+
+		LedGreen::reset();
+		modm::delay(300ms);
+	}
+
+	return 0;
+}
--- a/src/modm/architecture/interface/spi_master.hpp
+++ b/src/modm/architecture/interface/spi_master.hpp
@@ -117,7 +117,7 @@ public:
 	 *      number of bytes to be shifted out
 	 */
 	static void
-	transferBlocking(uint8_t *tx, uint8_t *rx, std::size_t length);
+	transferBlocking(const uint8_t *tx, uint8_t *rx, std::size_t length);
 
 	/**
 	 * Swap a single byte and wait for completion non-blocking!.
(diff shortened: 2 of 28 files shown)


salkinium:
> While creating this PR I noticed the ads7843-driver which should be compatible with the *2046 touchscreen drivers...

But is it? https://github.com/modm-io/modm/pull/503

src/modm/architecture/interface/spi_master.hpp
```diff
@@ -103,7 +103,7 @@ class SpiMaster : public ::modm::PeripheralDriver, public Spi
 	 * @return	received data
 	 */
 	static uint8_t
-	transferBlocking(uint8_t data);
+	transferBlocking(const uint8_t data);
```
salkinium: This is passed by value, so it's irrelevant if it's const. It also will cause issues, since some of the drivers modify the copy of the input data.

chris-durand: Both function declarations are exactly equivalent, it only matters in the definition. `const` on values is entirely ignored by the compiler in declarations. const by-value parameters are not generally recommended. Not so much here, but in other cases with non-trivial data types `const` inhibits move.  For example https://quuxplusone.github.io/blog/2019/01/03/const-is-a-contract/ has more discussions on that subject.

examples/nucleo_l452re/graphics_touch/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+	using Miso = modm::platform::GpioB14;
+	using Mosi = modm::platform::GpioB15;
+	//using Interrupt = modm::platform::GpioA10;
+}
+
+modm::Touch2046<touch::Spi, touch::Cs> touchController;
+
+
+/*
+ * Blinks the green user LED with 1 Hz.
```
chris-durand: Description needs update, seems to be a copy-pase mistake.

src/modm/architecture/interface/spi_master.hpp
salkinium: Yeah, I remember falling into this trap when trying to have an overloaded function with and without const. Doesn't work, since it's ignored in the signature…

src/modm/driver/touch/touch2046.hpp
```diff
@@ -0,0 +1,134 @@
...
+	 */
+	struct Calibration
+	{
+		int16_t OffsetX = 0;
+		int16_t OffsetY = 0;
+		int32_t FactorX = 1'000'000;
+		int32_t FactorY = 1'000'000;
+		uint16_t MaxX = 240;
+		uint16_t MaxY = 320;
+		uint16_t TresholdZ = 1500;
```
chris-durand: Spelling mistake
```suggestion
		uint16_t ThresholdZ = 1500;
```

src/modm/driver/touch/touch2046.hpp
```diff
@@ -0,0 +1,134 @@
...
+	modm::ResumableResult<std::tuple<uint16_t,uint16_t>>
+	getTouchPosition();
+
+private:
+	static constexpr uint8_t MeasureZ1 = 0xB1;
+	static constexpr uint8_t MeasureZ2 = 0xC1;
+	static constexpr uint8_t MeasureX = 0xD1;
+	static constexpr uint8_t MeasureY = 0x91;
+	static constexpr uint8_t Powerdown = 0b1111'1100;
+	const std::array<uint8_t, 17> bufferWrite = {
```
chris-durand: All array values are compile-time constant expressions. `bufferWrite` can be `static` and `constexpr`. Otherwise the compiler will place it in RAM for each instance.

src/modm/driver/touch/touch2046.hpp
```diff
@@ -0,0 +1,134 @@
...
+		MeasureY, 0x00,
+		MeasureX, 0x00,
+		MeasureY, 0x00,
+		MeasureX, 0x00,
+		MeasureY, 0x00,
+		(MeasureX & Powerdown), 0x00,
+		0x00};
+	std::array<uint16_t, 9> bufferRead = {};
+
+    uint16_t x = 0;
```
chris-durand: Indentation is not consistent with tabs and spaces, also in other places.

src/modm/architecture/interface/spi_master.hpp
rleh: The point is: I would like to pass `const uint8_t*` ~(and `const uint8_t`)~ to the `transfer*()` functions.
With the `const` in the function signature we are able to pass `const` and non-`const` data, otherwise only non-`const`.

src/modm/driver/touch/touch2046.hpp
```diff
@@ -0,0 +1,134 @@
...
+	setCalibration(Calibration calibration) {
+		cal = calibration;
+	}
+
+	/**
+	 * \brief	Get raw X, Y and Z values
+	 *
+	 * \return	Position and intensity of touch point. Full int16_t range.
+	 */
+	modm::ResumableResult<std::tuple<uint16_t,uint16_t,uint16_t>>
```
chris-durand, line 86: In other places `modm::Vector<T, 2>` and `modm::Vector<T, 3>` are used for points. What do you think about using some named type instead of tuple?

EDIT: Oh, they don't work for unsigned 16bit ints. I see what you are doing there with `std::tie`. I still have the feeling an explicit point type would be a cleaner interface.

rleh:
> But is it? #503

Interesting...

src/modm/driver/touch/touch2046.hpp
rleh: Oh sorry.
I somehow switch my IDE to spaces and didn't noticed...

src/modm/architecture/interface/spi_master.hpp
chris-durand: I totally agree for `const uint8_t*`, but `const uint8_t data` is exactly the same as `uint8_t data` for a parameter in a **function declaration**, because it's a value. The value is always copied, so it is completely transparent for the caller if it is modified or not. You can always pass any const and rvalue thing to a function `foo(uint8_t)`. Think of `foo(5)`. It's semantically the same. `const uint8_t*` has const on the object it points to.  The analogy for `const uint8_t data` would be a const of the pointer itself like: `const uint8_t* const data`. That would be equivalent to `const uint8_t* data` in the context here.

src/modm/driver/touch/touch2046.hpp
rleh: There is also `modm::glcd::Point` (which is a typedef of `Vector<int16_t, 2>`), but since modm has no real graphics subsystem (yet?) the user write something himself or integrate third party library like [lvgl](https://lvgl.io/).
For that use-cases I consider `std::tuple` the most useful option, especially thanks to the existence of `std::tie()`.

My impression is also that the modm containers are used less and less, especially since we have [avr-libstdcpp](https://github.com/modm-io/avr-libstdcpp).
I think we could deprecate the modm containers at some point in the future and use STL (where possible) or [ETL](https://github.com/ETLCPP/etl/) instead.

salkinium: +1 for deprecation in favor of ETL.

src/modm/driver/touch/touch2046.hpp
rleh: 🙈

src/modm/architecture/interface/spi_master.hpp
rleh: You are right, I was focused on the pointers and didn't thought about the values.

src/modm/driver/touch/touch2046.hpp
rleh: But ETL will not happen in this PR and probably not in the next week.

So is `std::tuple` ok here or should I change it?

chris-durand: I'd say leave it for now. Here it is quite clear that the values are coordinates and ordered x,y,z. I've seen real bugs caused by questionable APIs returning tuples with two `int`s and three `bool`s challenging users to get the order right. That's why I got a bit triggered by the extensive use of `std::tuple` when I saw it first but in this case there is no actual risk.

rleh:
Proof photo from the hardware test:
grafik

rleh:
> But [ETL](https://github.com/ETLCPP/etl/) will not happen in this PR and probably not in the next week.

Might happen faster than expected... https://github.com/modm-ext/etl-partial

src/modm/driver/touch/touch2046.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+/// \ingroup modm_driver_touch2046
+struct touch2046 {
+	enum class Orientation {
+		Normal,
+		//...
+	};
+
+	/**
+	 * \brief	Calibration values are used to calculate touch point
```
salkinium: We don't need the `\brief`/`@brief` keywords, since we have autobrief enabled. Just FYI before you copy-paste that everywhere ;-P

src/modm/driver/touch/touch2046_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_TOUCH2046_HPP
+	#error	"Don't include this file directly, use 'ltc2984.hpp' instead!"
+#endif
```
salkinium, line 15: Maybe we should just get rid of this and use `#pragma once` everywhere. This is just an additional source of error, nobody will include the `_impl.hpp` file anyways.

src/modm/driver/touch/touch2046_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+	RF_END_RETURN(z > cal.ThresholdZ);
+}
+
+template < class SpiMaster, class Cs >
+modm::ResumableResult<std::tuple<uint16_t,uint16_t>>
+modm::Touch2046<SpiMaster, Cs>::getTouchPosition()
+{
+	RF_BEGIN();
+
+	std::tie(x, y, std::ignore) = RF_CALL(getRawValues());
```
salkinium, line 65: Amazing that this works considering the UB hell that is RF_CALL 🤣

examples/nucleo_l452re/graphics_touch/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+modm::Touch2046<touch::Spi, touch::Cs> touchController;
+
+
+int
+main()
+{
+	using namespace modm::literals;
+	Board::initialize();
+    Dma1::enable();
```
salkinium: 🚨🚨🚨🚨 indentation police 🚨🚨🚨🚨

src/modm/driver/touch/touch2046.lb
```diff
@@ -0,0 +1,38 @@
...
+# TSC2046/x2046 Resistive Touch Controller
+
+The TSC2046 by Texas Instruments is a popular touchscreen controller
+on cheap displays.
+
+https://www.ti.com/lit/ds/symlink/tsc2046.pdf
+
+There are many compatible devices from other manufacturers available,
+such as the XPT2046 by XPTEK or "2046" labeled chips from unknown manufacturers.
+All of these are 100%% compatible with the TSC2046.
```
salkinium: Is this %% for escaping?

rleh: It was copied from `ads7843.lb` and my guess was some python and/or Doxygen escaping, but that does not seem to be the case.

src/modm/driver/touch/touch2046.hpp
rleh: Nice to known.

src/modm/driver/touch/touch2046_impl.hpp
rleh: I would like to keep the `_impl.hpp`-include-guards; from my experience a lot of people just try to include any file they see (in their IDE auto-completion).

salkinium (approved):

PR #601, opened 2021-03-31
[display] Fix color paradox of monochrome displays 

TomSaw:
In `MonochromeGraphicDisplayBufferedVertical<Width, Height>::drawHorizontalLine` no pixels are set, because `this->foregroundColor != glcd::Color::black()`. The default foregroundColor fot graphic displays is Color::white()!

So i have to `myMonochromeDisplay.setColor(Color::black())` after instantiation of my display to have horizontal lines renders.

**-> That's confusing, cause anything else - like text, diagonal lines, circles - renders (in black) out of the box!**

### DIFF
--- a/src/modm/driver/display/ili9341.hpp
+++ b/src/modm/driver/display/ili9341.hpp
@@ -15,7 +15,7 @@
 #include <modm/architecture/interface/register.hpp>
 #include <modm/math/utils/endianness.hpp>
 #include <modm/platform/gpio/base.hpp>
-#include <modm/ui/display/graphic_display.hpp>
+#include <modm/ui/display/colored_graphic_display.hpp>
 
 namespace modm
 {
--- a/src/modm/driver/display/parallel_tft.hpp
+++ b/src/modm/driver/display/parallel_tft.hpp
@@ -16,7 +16,7 @@
 #define MODM_PARALLEL_TFT_HPP
 
 #include <modm/architecture/interface/delay.hpp>
-#include <modm/ui/display/graphic_display.hpp>
+#include <modm/ui/display/colored_graphic_display.hpp>
 
 namespace modm
 {
--- a/src/modm/ui/display/colored_graphic_display.cpp
+++ b/src/modm/ui/display/colored_graphic_display.cpp
@@ -1,0 +1,26 @@
+#include "colored_graphic_display.hpp"
+
+// ----------------------------------------------------------------------------
+void
+modm::ColoredGraphicDisplay::setColor(const glcd::Color& newColor)
+{
+//	if (newColor == glcd::Color::black()) {
+//		draw = &modm::GraphicDisplay::clearPixel;
+//	}
+//	else {
+//		draw = &modm::GraphicDisplay::setPixel;
+//	}
+
+	/* When using a multicolor display we don't need clearPixel(), or at least
+	 * not the way it was implemented above. Maybe check if newColor equals
+	 * backgroundColor.
+	 * */
+	draw = &modm::ColoredGraphicDisplay::setPixel;
+	this->foregroundColor = newColor;
+}
+
+void
+modm::ColoredGraphicDisplay::setBackgroundColor(const glcd::Color& newColor)
+{
+	this->backgroundColor = newColor;
+}
--- a/src/modm/ui/display/colored_graphic_display.hpp
+++ b/src/modm/ui/display/colored_graphic_display.hpp
@@ -1,0 +1,108 @@
+#ifndef MODM_COLORED_GRAPHIC_DISPLAY_HPP
+#define MODM_COLORED_GRAPHIC_DISPLAY_HPP
+
+#include <modm/platform.hpp>
+
+#include "graphic_display.hpp"
+
+using namespace modm::platform;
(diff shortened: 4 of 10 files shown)


salkinium:
> The default foregroundColor fot graphic displays is Color::white()!

Isn't this the issue? Shouldn't this be initialized to black?

salkinium:
Hm, so apparently by default we have dark mode UI…

```cpp
foregroundColor(glcd::Color::white()),
backgroundColor(glcd::Color::black()),
```

The issue I see is that the mapping from color to bit operation is undefined. The display could be black-on-white LCD or a white-on-black OLED for example. Maybe a general `setInverted()` function that reverses the bit operations set <=> clear would be better?

TomSaw:
I wanted to quickly patch this but - why not - let's touch it once!

**Color** has nothing to do in any of the `MonochromeGraphicDisplay`. For MonochromeGraphicDisplay it's not colors but pixel **Light / Dark** (Don't forget some mono-displays support grayscales):

- LCD pixels will be blackish on light background. Background **color is a hardware property.**
- OLED pixels will be light on black background. Pixel **color is a hardware property.**

**The GraphicDisplay inheritance should be refactored:**

GraphicDisplay <-- MonochromeGraphicDisplay(**Light / dark** stuff in here) <-- SSD1306, KS0108, ...
GraphicDisplay <-- ColoredGraphicDisplay(**Color** stuff in here) <-- ILI9341, ,,,

> The issue I see is that the mapping from color to bit operation is undefined

... no more mapping required, and we also spare some RAM when using mono-displays.

salkinium:
I mean, the UI stuff in modm is a little very broken. It started with binary displays and then color was grafted onto it, then the `:ui:gui` modules were added. But the whole thing was never properly refactored. I started something here as a research project, but never finished it… https://github.com/salkinium/upainter

Do you want to spend time on fixing this? It seems like a lot of work tbh. I'd be ok with adding more hot fixes onto this mess and then burning it with fire some time later… or integrate lvgl or similiar.

TomSaw:
Looks like lot's of fun. If i can lean my head on your shoulder (collecting feedback and input) from time to time, i can imagine this!

Let's clone and fly over your repo first.

TomSaw:
> then the :ui:gui modules were added

I've developed an interesting approach for gui modeling. Want to publish it when i'm ready. This all makes more&more sense 😏

TomSaw:
Ok salkinium, got a first impress. Thats a very ambitious attempt and indeed looks like a lot of work 😂
I would have to learn a ton of stuff to master it but i think there's already too much on my list. (f.e. the AVR Timers PR + discovered various traps in the l3gd20 driver)

Pixels are beautiful, so i put this to my bed-lecture and let it flourish. Also i will not spend to much time in improving the current graphics-API.

Maybeee, it's ok for now, if we have aliased, slow graphics on our MCU-projects. I don't want to render the next Toystory. I'm happy with some text and horizontal lines.

**Let's go for some refactoring?**

salkinium:
> Let's go for some refactoring?

Sure, go ahead. I don't have much of an opinion on the GUI stuff though, so whatever you come up with that fixes your problem without breaking everything is fine with me ;-P

TomSaw:
What's VirtualGraphicDisplay for? Last update was 2013. Do we need it?

salkinium:
Good question, ~~I think it was related to using a SDL-backed display on the PC? It's unused in the repo, I think we can remove it.~~

salkinium:
Oh wait, it could be for using a subdisplay in a larger display. Like side-by-side splitting?

Edit: Yes, that's exactly what it's used for inside the unpublished @roboterclubaachen code. Don't remove it!

TomSaw:
Okee

rleh:
I also looked at upainter (after giving up on LittlevGL (Info: LittlevGL was renamed to [LVGL](https://lvgl.io/) last year.) in #362);
Since then I consider creating and maintaining a graphics framework in modm to be a very big effort, which I would not want to spend.

That's why I ended up with [LVGL](https://lvgl.io/) again yesterday, although I'd much rather have a C++ API.
I got [LVGL](https://lvgl.io/) running on the STM32F469NI-Disco board running today, will add examples for Ili9341 and/or SSD1306 displays soon and open a [PR](https://github.com/rleh/modm/tree/lvgl)...

grafik

TomSaw:
Looks promising.. can't wait for your examples @rleh 
And if it's just to study another working thing.

TomSaw:
Slept over it and obtained a clear opinion:

Let's keep modm's graphic capabilities rudimental but maintained.
It's kinda RISC vs. CISC architecture thinking ;)
If modm's graphic APIs stay lightweight **and!!!** C++ with all it's advantages, it fulfills the purpose of a tool as it's best!

_I often prefer monitoring data on a small OLED instead of the terminal. This invites me for playing around, even when the laptop is shutdown. For this case, i prefer lightweight code and don't want the pling pling and rainbow-colors from advanced renderings._

Whenever one feels the urge to build a smartphone there are plenty of powerfull 3rd-party libraries - like LVGL - out there, more coming for sure.

Herefore followup PR has been created: #604

PR #609, opened 2021-04-08
Fix STM32F3/G4/L4 ADC initialization

chris-durand:
Two fixes, tested with Nucleo G474RE and a custom board with an STM32G473RB:

- ADC initialization hangs for larger clock prescalers
Setting ADC_CR_ADEN after calibration is only possible after 4 ADC clock
cycles have elapsed. In case the ADC clock is much slower than the CPU
clock that flag was never set and the initialization blocked infinitely
waiting for the ready flag to get set.

- The ADC clock in the Nucleo G474RE example was set to 170 MHz which exceeds the 60 MHz maximum clock.

### DIFF
--- a/examples/nucleo_g474re/adc_basic/main.cpp
+++ b/examples/nucleo_g474re/adc_basic/main.cpp
@@ -22,8 +22,10 @@ main()
 	MODM_LOG_INFO << " running on Nucleo-G474RE" << modm::endl << modm::endl;
 
 	MODM_LOG_INFO << "Configuring ADC ...";
+	// max. ADC clock for STM32G474: 60 MHz
+	// 170 MHz AHB clock / 4 = 42.5 MHz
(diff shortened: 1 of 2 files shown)


chris-durand:
The driver needs some more fixes, e.g. the public `calibrate()` function can't be called after initialize without manually setting some registers . I will do it during the weekend.

salkinium:
Friendly ping?

rleh:
Do we want to merge this PR already as it is now? Additional fixes and improvements can be done later anyway...

rleh (approved):

chris-durand:
I am quite busy right now and don't have time for modm, sorry. This PR contains important bug fixes and can be merged as is. Any further refactoring can be done in another PR.

salkinium (approved):

PR #604, opened 2021-04-02
Refactor modm:ui:display

TomSaw:
Followup from #601 

Color has nothing to do in any of the `MonochromeGraphicDisplay` but it inherits Color from `GraphicDisplay`. For `MonochromeGraphicDisplay` it shouldn't be color but pixel On / Off.

- LCD pixels will be blackish on light background. Background color is a hardware property.
- OLED pixels will be light on black background. Pixel color is a hardware property.

#### New inheritances

GraphicDisplay <-- ColoredGraphicDisplay _(Color stuff in here)_ <-- ILI9341, ,,,
GraphicDisplay <-- MonochromeGraphicDisplay _(Light / dark stuff in here)_ <-- SSD1306, KS0108, ...

- [x] New specialised interface GraphicDisplay <- ColoredGraphicDisplay; move color-stuff in there
  - [x] Test Displays inheriting `ColoredGraphicDisplay`
- [x] New specialised interface GraphicDisplay <- MonochromeGraphicDisplay
  - [x] Test Displays inheriting `MonochromeGraphicDisplay`
  - [x] Implement optimized drawVerticalLine for MonochromeVerticalDisplay
- [x] Modernize code
- [x] Update docs

PR #616, opened 2021-04-17
Refactor modm:ui:color

TomSaw:
While working on #604 the fragmentation of Colors (modm::ui::color and modm::glcd::Color) forced me to this PR.

### Key changes
- Integrated modm::glcd::Color (Rgb565) with modm::ui::color
  - Backwards compatible: named constructors still work `glcd::Color::black()`, `glcd::Color::white()` a.s.o.
- New "Color" Brightness to be used with Grayscale and Monochrome Displays, Simple dimmed LEDs
- New Color-Factory color::RgbHtml returning color::Rgb by by HTML-Colornames
- Moved colors into own dir ../modm/ui/color and seperate *.hpp
- Replaced ::toHsv(...), ::toRgb(...) with **constexpr conversion constructors**.
  - You can now simply type `color_display.setColor(color::Rgb(255, 0, 0))` or `color_display.setColor(color::Hsv(0, 100, 100))` to submit pure Red in Rgb565 format.
  - Also check /src/modm/ui/led/rgb.hpp where `setColor(Hsv color)` and `fadeTo(Hsv color, uint16_t time)` has become rudimentary.

### Performance
Using **conversion constructors** is defacto more elegant. I've compared the compilates: **No differences ...**

- ... no matter if you compare old vs new implementation
- ... no matter if you compare various constructors for one and the same color:
  - `color_display.setColor(color::Rgb565(0xF800))`
  - vs `color_display.setColor(color::Rgb(255, 0, 0))`
  - vs `color_display.setColor(color::Hsv(0, 100, 100))`

### DIFF
--- a/examples/arduino_nano/color/main.cpp
+++ b/examples/arduino_nano/color/main.cpp
@@ -64,11 +64,8 @@ public:
 			PT_WAIT_UNTIL(TCS3472_INT::read() == false);
 			if (PT_CALL(sensor.readColor()))
 			{
-				const auto color = data.getColor();
-				MODM_LOG_INFO << "RGB: " << color;
-				modm::color::HsvT<uint16_t> hsv;
-				color.toHsv(&hsv);
-				MODM_LOG_INFO << "\tHSV: " << hsv << modm::endl;
+				const auto rgb = data.getColor();
+				MODM_LOG_INFO << "RGB: " << rgb << "\tHSV: " << modm::color::Hsv(rgb) << modm::endl;
 			}
 		}
 
--- a/examples/avr/display/dogm128/benchmark/main.cpp
+++ b/examples/avr/display/dogm128/benchmark/main.cpp
@@ -21,6 +21,7 @@
 using namespace modm::platform;
 using namespace modm::literals;
 using namespace std::chrono_literals;
+using namespace modm::color::html;
 
 
 namespace led
@@ -161,7 +162,7 @@ main()
 		while (!timer.execute())
 		{
 			// rectangle in left side of screen
-			// TODO display.setColor(Color::black());
+			// display.setColor(Black);
 			display.drawRectangle(Point(0, 0), 62, 62);
 
 			// rounded rectangle around text area
@@ -176,14 +177,14 @@ main()
 			display.drawCircle(Point(31, 31), 31);
 
 			// clear previous spinner position
-			// TODO display.setColor(Color::white());
+			// TODO display.setColor(White);
 			display.fillRectangle(Point(87, 40), 16, 16);
 
 			static uint8_t loops = 0;
-			// TODO display.setColor(Color::black());
+			// TODO display.setColor(Black);
 			drawSpinner(Point(95, 48), loops++);
 
-			// TODO display.setColor(Color::white());
+			// TODO display.setColor(White);
 			display.setCursor(Point(25, 40));
 			display << ++iter;
 			display.update();
@@ -192,7 +193,7 @@ main()
 		// print number of iterations in one second
 		display.clear();
 		display.setCursor(Point(80, 10));
-		// TODO display.setColor(Color::white());
+		// display.setColor(White);
 		display << "FPS=" << iter;
 	}
 }
--- a/src/modm/driver/display/ili9341.hpp
+++ b/src/modm/driver/display/ili9341.hpp
@@ -236,12 +236,12 @@ public:
 	{ setColoredPixel(x, y, backgroundColor); }
 
     // TODO implement getPixel for ili9341
-	glcd::Color
+	color::Rgb565
 	getPixel(int16_t x, int16_t y) const final
 	{
 		(void) x;
 		(void) y;
-		return glcd::Color::white();
+		return modm::color::html::White;
 	}
 
 	void
@@ -270,7 +270,7 @@ public:
 				 modm::accessor::Flash<uint8_t> data) final;
 
 	void
-	drawRaw(glcd::Point upperLeft, uint16_t width, uint16_t height, glcd::Color* data);
+	drawRaw(glcd::Point upperLeft, uint16_t width, uint16_t height, color::Rgb565* data);
 
 	void
 	setScrollArea(uint16_t topFixedRows, uint16_t bottomFixedRows, uint16_t firstRow);
@@ -290,7 +290,7 @@ protected:
 
 private:
 	void
-	setColoredPixel(int16_t x, int16_t y, glcd::Color const &color);
+	setColoredPixel(int16_t x, int16_t y, color::Rgb565 const &color);
 
 	void
 	setClipping(uint16_t x, uint16_t y, uint16_t width, uint16_t height);
--- a/src/modm/ui/color/brightness.hpp
+++ b/src/modm/ui/color/brightness.hpp
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2021, Thomas Sommer
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include <stdint.h>
+
+#include <concepts>
+
+#include "hsv.hpp"
+#include "rgb.hpp"
+#include "rgb565.hpp"
+
+namespace modm::color
+{
+
+// forward declarations for convertion constructors
+template<std::unsigned_integral T>
+class RgbT;
+
+template<std::unsigned_integral T>
+class HsvT;
+
+class Rgb565;
+
+/**
+ * @brief		Brightness as unsigned integral. Add's the math for conversion to and from
+ *				Color-Types. Use with: Grayscale Buffers, Dimmed LEDs, Brightness sensors
+ *
+ * @author		Thomas Sommer
+ *
+ * @tparam	T	Unsigned integral for the brightness-value
+ * @ingroup		modm_ui_color
+ */
+template<std::unsigned_integral T>
+class BrightnessT
+{
+public:
+	T value{0};
+
+	constexpr BrightnessT() = default;
+
+	constexpr BrightnessT(T value) : value(value) {}
+
+	/**
+	 * Copy Constructor 8bit->16bit
+	 */
+	template<std::unsigned_integral U>
+		requires std::is_same_v<T, uint16_t> && std::is_same_v<U, uint8_t>
+	constexpr BrightnessT(const BrightnessT<U> &brightness_other)
+		: value(brightness_other.value << 8)
+	{}
+
+	/**
+	 * Copy Constructor 16bit->8bit
+	 */
+	template<std::unsigned_integral U>
+		requires std::is_same_v<T, uint8_t> && std::is_same_v<U, uint16_t>
+	constexpr BrightnessT(const BrightnessT<U> &brightness_other)
+		: value(brightness_other.value >> 8)
+	{}
+
+	/**
+	 * Convertion Constructor for RGB Color
+	 *
+	 * @param rgb	RGB Color
+	 */
+	template<std::unsigned_integral U>
+	constexpr BrightnessT(RgbT<U> rgb)
+		: value((0.2125 * float(rgb.red)) + (0.7154 * float(rgb.green)) +
+				(0.0721 * float(rgb.blue)))
+	{}
+
+	/**
+	 * Convertion Constructor for HSV Color
+	 *
+	 * @param hsv	HSV Color
+	 */
+	template<std::unsigned_integral U>
+	constexpr BrightnessT(HsvT<U> hsv) : value(hsv.value)
+	{}
+
+	/**
+	 * Convertion Constructor for RGB565 Color
+	 *
+	 * @param rgb565	RGB565 Color
+	 */
+	constexpr BrightnessT(Rgb565 rgb565) : BrightnessT<T>(RgbT<uint8_t>(rgb565)) {}
+
+	constexpr bool
+	operator==(const BrightnessT<T> &other) const = default;
+};
+
+/// @ingroup modm_ui_color
+using Brightness = BrightnessT<uint8_t>;
+
+}  // namespace modm::color
--- a/src/modm/ui/color/hsv.hpp
+++ b/src/modm/ui/color/hsv.hpp
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2009, Martin Rosekeit
+ * Copyright (c) 2009-2013, Fabian Greif
+ * Copyright (c) 2012-2013, 2015, Niklas Hauser
+ * Copyright (c) 2013, David Hebbeker
+ * Copyright (c) 2021, Thomas Sommer
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_COLOR_HSV_HPP
+#define MODM_COLOR_HSV_HPP
+
+#include <stdint.h>
+#include <concepts>
+
+#include "brightness.hpp"
+#include "rgb.hpp"
+#include "rgb565.hpp"
+
+namespace modm::color
+{
+
+// forward declarations for convertion constructors
+template<std::unsigned_integral T>
+class RgbT;
+
+template<std::unsigned_integral T>
+class BrightnessT;
+
+class Rgb565;
+
+/**
+ * @brief		Color in HSV Colorspace
+ *
+ * @author		Martin Rosekeit, Fabian Greif, Niklas Hauser, David Hebbeker, Thomas Sommer
+ * @ingroup		modm_ui_color
+ */
+template<std::unsigned_integral T>
+class HsvT
+{
+public:
+	T hue{0};
+	T saturation{0};
+	T value{0};
+
+	constexpr HsvT() = default;
+
+	constexpr HsvT(T hue, T saturation, T value) : hue(hue), saturation(saturation), value(value) {}
+
+	/**
+	 * Copy Constructor 8bit->16bit
+	 */
+	template<std::unsigned_integral U>
+		requires std::is_same_v<T, uint16_t> && std::is_same_v<U, uint8_t>
+	constexpr HsvT(const HsvT<U> &hsv_other)
+		: hue(hsv_other.hue << 8), saturation(hsv_other.saturation << 8), value(hsv_other.value << 8)
+	{}
+
+	/**
+	 * Copy Constructor 16bit->8bit
+	 */
+	template<std::unsigned_integral U>
+		requires std::is_same_v<T, uint8_t> && std::is_same_v<U, uint16_t>
+	constexpr HsvT(const HsvT<U> &hsv_other)
+		: hue(hsv_other.hue >> 8), saturation(hsv_other.saturation >> 8), value(hsv_other.value >> 8)
+	{}
+
+	/**
+	 * Convertion Constructor for RGB Color
+	 *
+	 * @param rgb	RGB Color
+	 */
+	template<std::unsigned_integral U>
+	constexpr HsvT(const RgbT<U>& rgb);
+
+	/**
+	 * Convertion Constructor for Brightness
+	 *
+	 * @param brightness	Brightness 'Color'-object
+	 */
+	template<std::unsigned_integral U>
+	constexpr HsvT(const BrightnessT<U> gray) : hue(0), saturation(0), value(gray.value)
+	{}
+
+	/**
+	 * Convertion Constructor for RGB565 Color
+	 *
+	 * @param rgb565	RGB565 Color
+	 */
+	constexpr HsvT(const Rgb565& rgb565) : HsvT(RgbT<uint8_t>(rgb565)) {}
+
+	constexpr bool
+	operator==(const HsvT<T>& other) const = default;
+
+private:
+	template<std::unsigned_integral U>
+	friend IOStream&
+	operator<<(IOStream&, const HsvT<U>&);
+};
+
+/// @ingroup modm_ui_color
+using Hsv = HsvT<uint8_t>;
+
+
+#if __has_include(<modm/io/iostream.hpp>)
+#include <modm/io/iostream.hpp>
+
+template<std::unsigned_integral U>
+IOStream&
+operator<<(IOStream& os, const color::HsvT<U>& color)
+{
+	os << color.hue << "\t" << color.saturation << "\t" << color.value;
+	return os;
+}
+#endif
+
+}  // namespace modm::color
+
+#include "hsv_impl.hpp"
+
+#endif  // MODM_COLOR_HSV_HPP
--- a/src/modm/ui/color/rgb.hpp
+++ b/src/modm/ui/color/rgb.hpp
@@ -1,0 +1,164 @@
+/*
+ * Copyright (c) 2009, Martin Rosekeit
+ * Copyright (c) 2009-2013, Fabian Greif
+ * Copyright (c) 2012-2013, 2015, Niklas Hauser
+ * Copyright (c) 2013, David Hebbeker
+ * Copyright (c) 2021, Thomas Sommer
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_COLOR_RGB_HPP
+#define MODM_COLOR_RGB_HPP
+
+#include <stdint.h>
+#include <bit>
+
+#include <concepts>
+#include <modm/io/iostream.hpp>
+#include <modm/math/utils/arithmetic_traits.hpp>
+#include <type_traits>
+
+#include "brightness.hpp"
+#include "hsv.hpp"
+#include "rgb565.hpp"
+
+namespace modm::color
+{
+
+// forward declarations for convertion constructors
+template<std::unsigned_integral T>
+class HsvT;
+
+template<std::unsigned_integral T>
+class BrightnessT;
+
+class Rgb565;
+
+/**
+ * Color in HSV Colorspace
+ *
+ * @author		Martin Rosekeit, Fabian Greif, Niklas Hauser, David Hebbeker, Thomas Sommer
+ * @ingroup		modm_ui_color
+ */
+template<std::unsigned_integral T>
+class RgbT
+{
+public:
+	T red{0};
+	T green{0};
+	T blue{0};
+
+	using TSum = modm::WideType<T>;
+
+	constexpr RgbT() = default;
+
+	constexpr RgbT(T red, T green, T blue) : red(red), green(green), blue(blue) {}
+
+	/**
+	 * Copy Constructor 8bit->16bit
+	 */
+	template<std::unsigned_integral U>
+		requires std::is_same_v<T, uint16_t> && std::is_same_v<U, uint8_t>
+	constexpr RgbT(const RgbT<U> &rgb_other)
+		: red(rgb_other.red << 8), green(rgb_other.green << 8), blue(rgb_other.blue << 8)
+	{}
+
+	/**
+	 * Copy Constructor 16bit->8bit
+	 */
+	template<std::unsigned_integral U>
+		requires std::is_same_v<T, uint8_t> && std::is_same_v<U, uint16_t>
+	constexpr RgbT(const RgbT<U> &rgb_other)
+		: red(rgb_other.red >> 8), green(rgb_other.green >> 8), blue(rgb_other.blue >> 8)
+	{}
+
+	/**
+	 * Convertion Constructor for HSV Color
+	 *
+	 * @param hsv	HSV Color
+	 */
+	template<std::unsigned_integral U>
+	constexpr RgbT(const HsvT<U>& hsv);
+
+	/**
+	 * Convertion Constructor for Brightness
+	 *
+	 * @param brightness	Brightness 'Color'-object
+	 */
+	// TODO Plump conversion, implement the right way
+	template<std::unsigned_integral U>
+	constexpr RgbT(const BrightnessT<U> brightness)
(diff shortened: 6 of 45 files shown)


examples/arduino_nano/color/main.cpp
```diff
@@ -64,11 +64,8 @@ class Sensorthread : public modm::pt::Protothread
 			PT_WAIT_UNTIL(TCS3472_INT::read() == false);
 			if (PT_CALL(sensor.readColor()))
 			{
-				const auto color = data.getColor();
-				MODM_LOG_INFO << "RGB: " << color;
-				modm::color::HsvT<uint16_t> hsv;
-				color.toHsv(&hsv);
-				MODM_LOG_INFO << "\tHSV: " << hsv << modm::endl;
+				const auto rgb = data.getColor();
+				MODM_LOG_INFO << "RGB: " << rgb << "\tHSV: " << modm::color::Hsv(rgb) << modm::endl;
```
salkinium, line 68: Yup, so much better with the conversion constructor!

src/modm/ui/color/hsv.hpp
```diff
@@ -0,0 +1,80 @@
...
+{
+public:
+	UnderlyingType hue{0};
+	UnderlyingType saturation{0};
+	UnderlyingType value{0};
+
+	constexpr HsvT() = default;
+
+	constexpr HsvT(const UnderlyingType hue, const UnderlyingType saturation,
+				   const UnderlyingType value)
```
salkinium: no need for const when pass-by-copy.

src/modm/ui/color/hsv.hpp
```diff
@@ -0,0 +1,80 @@
...
+{
+namespace color
+{
+
+template<typename UnderlyingType>
+class RgbT;
+
+class Rgb565;
+/// @ingroup modm_ui_color
+template<typename UnderlyingType = uint8_t>
```
salkinium: Do we want to use `T` instead of `UnderlyingType`? It would be more consistent with the rest of modm.

src/modm/ui/color/rgb_impl.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+	switch (i)
+	{
+		case 0: red = hsv.value; green = u; blue = p; break;
+		case 1: red = u; green = hsv.value; blue = p; break;
+		case 2: red = p; green = hsv.value; blue = u; break;
+		case 3: red = p; green = u; blue = hsv.value; break;
+		case 4: red = u; green = p; blue = hsv.value; break;
+		case 5: red = hsv.value; green = p; blue = u; break;
+	}
```
salkinium, line 47: Do you want to add a simple unittest for the color conversion stuff? Can be a simple hardcoded color comparison HSV vs RGB.

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,62 @@
...
+class Rgb565
+{
+public:
+	uint16_t color{0x0000};
+
+	constexpr Rgb565() = default;
+
+	constexpr Rgb565(const uint16_t color) : color(color) {}
+
+	constexpr Rgb565(const uint8_t red, const uint8_t green, const uint8_t blue)
```
salkinium:
```suggestion
	constexpr Rgb565(uint8_t red, uint8_t green, uint8_t blue)
```

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,62 @@
...
+// Rgb565 (565) Format
+/// @ingroup modm_ui_color
+class Rgb565
+{
+public:
+	uint16_t color{0x0000};
+
+	constexpr Rgb565() = default;
+
+	constexpr Rgb565(const uint16_t color) : color(color) {}
```
salkinium:
```suggestion
	constexpr Rgb565(uint16_t color) : color(color) {}
```

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,128 @@
...
+class RgbT
+{
+public:
+	UnderlyingType red{0};
+	UnderlyingType green{0};
+	UnderlyingType blue{0};
+
+	constexpr RgbT() = default;
+
+	constexpr RgbT(const UnderlyingType red, const UnderlyingType green, const UnderlyingType blue)
```
salkinium:
```suggestion
	constexpr RgbT(UnderlyingType red, UnderlyingType green, UnderlyingType blue)
```

salkinium (commented):
So far so good!

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,62 @@
...
+// Rgb565 (565) Format
+/// @ingroup modm_ui_color
+class Rgb565
+{
+public:
+	uint16_t color{0x0000};
+
+	constexpr Rgb565() = default;
+
+	constexpr Rgb565(uint16_t color) : color(color) {}
```
chris-durand, line 53: In my opinion all one-argument conversion constructors in the different color types should be marked `explicit`. Otherwise very obscure conversions can happen and the resulting code gets hard to read, especially with ambiguities when you overload on different color types. Type safety is the main reason why we define these types.

src/modm/driver/display/ili9341.hpp
```diff
@@ -280,7 +280,7 @@ class Ili9341 : public Interface, public modm::GraphicDisplay
 
 private:
 	void
-	setColoredPixel(int16_t x, int16_t y, glcd::Color const &color);
+	setColoredPixel(int16_t x, int16_t y, color::Rgb565 const &color);
```
chris-durand, line 293: For small types (size <= 2..3 ints) pass by value is generally recommended: [F.16: For “in” parameters, pass cheaply-copied types by value and others by reference to const](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in):

> When copying is cheap, nothing beats the simplicity and safety of copying, and for small objects (up to two or three words) it is also faster than passing by reference because it does not require an extra indirection to access from the function.

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+	template<typename _T>
+	constexpr Rgb565(const HsvT<_T> &hsv) : Rgb565(RgbT<uint8_t>(hsv))
+	{}
+
+	inline bool
+	operator==(const Rgb565 &other) const
+	{
+		return (color == other.color);
+	}
```
chris-durand: `constexpr` missing and the compiler can generate that one for you since C++20:
```suggestion
	constexpr bool
  	operator==(Rgb565 other) const = default;
```

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,128 @@
...
+		return RgbT<ReturnType>(getRelativeRed<IntermediateType, multiplier, ReturnType>(),
+								getRelativeGreen<IntermediateType, multiplier, ReturnType>(),
+								getRelativeBlue<IntermediateType, multiplier, ReturnType>());
+	}
+
+	constexpr bool
+	operator==(const RgbT<T>& other) const
+	{
+		return (red == other.red and green == other.green and blue == other.blue);
+	}
```
chris-durand: Same as below for the comparison in `Rgb565`.

src/modm/ui/color/hsv.hpp
```diff
@@ -0,0 +1,77 @@
...
+	template<typename _T>
+	constexpr HsvT(const RgbT<_T>& rgb);
+
+	constexpr HsvT(const Rgb565& rgb565) : HsvT(RgbT<uint8_t>(rgb565)) {}
+
+	constexpr bool
+	operator==(const HsvT<T>& other) const
+	{
+		return (hue == other.hue and saturation == other.saturation and value == other.value);
+	}
```
chris-durand: Same as below for the comparison in `Rgb565`.

src/modm/ui/color/rgb565.hpp
TomSaw: Sweeet

TomSaw:
<strike>Hey @salkinium please check color::Brightness. I've quickly sketched the RgbT(BrightnessT<_T> brightness) conversion constructor. What's the right reverse of the formula???

> 
> 8bpp Grayscale
> 8bits per pixel, but grayscale! This uses the following formula to map colors onto their perceived brightness:
> Luminence: (0.2125 * red) + (0.7154 * green) + (0.0721 * blue)</strike>

`/test/modm/ui/color/color_test.hpp` unveils asymetries on forward-backward conversion (Rgb -> Hsv -> Rgb). I think it's normal cause information is lost on conversion but really that much?

Edit: Got some new ideas and  investigate this in another session... **color_test.hpp is work in progress** Stay tuned...

src/modm/ui/color/rgb565.hpp
TomSaw: I see the point but also like the comfort of `implicit` conversion.

### 95% of applications look like
`ili9341.setColor(modm::color::Rgb(255, 0, 0));`
`ili9341.setColor(modm::color::Hsv(60 + hue_delta, 100, 100));`
`led.setBrightness(some_data.getRgb());`

**With `explicit` these instructions blowup.** Whilst updating the examples and display-drivers due to color-refactor, i had no issues with ambigouos conversions, so is kind of proofed, the situation can be handled.

**Let's try the `implicit` comfortable way, and when it shows that it's annoying, we add `explicit`!?**

chris-durand: Ok, if the implicit conversions are actually needed here for a usable interface leave them.

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,124 @@
...
+	/**
+	 * Convertion function to plain color
+	 */
+	inline operator uint16_t() const {
+		return color;
+	};
+
+	constexpr bool operator==(const Rgb565 &other) const = default;
+
+	Rgb565 operator+(const Rgb565 other) const {
```
TomSaw, line 105: I don't get this 😵 Very incomprehensible "incomplete type"

> error: variable ‘modm::color::Rgb565::RgbCalcType other_rgb’ has initializer but incomplete type RgbCalcType other_rgb(12, 13, 14);

Could you please give me a hint @chris-durand

TomSaw (commented):
Complicated "incomplete type" trouble

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,124 @@
...
+		// Compiler error: incomplete type
+		// But i have fwd declaration
+		// And i have already used RgbCalcType in constructors above!
+		// RgbCalcType other_rgb(12, 13, 14);
```
TomSaw, line 109: I don't get this 😵 Very incomprehensible "incomplete type"

> error: variable ‘modm::color::Rgb565::RgbCalcType other_rgb’ has initializer but incomplete type RgbCalcType other_rgb(12, 13, 14);

Could you please give me a hint @chris-durand

TomSaw: I don't get this 😵 Very incomprehensible "incomplete type"

> error: variable ‘modm::color::Rgb565::RgbCalcType other_rgb’ has initializer but incomplete type RgbCalcType other_rgb(12, 13, 14);

Could you please give me a hint @chris-durand

chris-durand: You have a cyclic dependency between Rgb565 and RgbT. Both instantiate values of each other inside their definitions in the headers.

chris-durand: If you only had references or pointers, the forward declarations would be enough. You can't create a value of an incomplete type because you don't know yet what it contains.

TomSaw: Okay thanks. I still don't understand, why i could use RgbCalcType above in the constructor but i'll implement it differently.

chris-durand: Because the particular constructors are template functions instantiated later.

src/modm/ui/color/rgbhtml.hpp
```diff
@@ -0,0 +1,190 @@
...
+	static constexpr RgbHtml Gainsboro()		{ return RgbHtml(220, 220, 220); };
+	static constexpr RgbHtml LightGray()		{ return RgbHtml(211, 211, 211); };
+	static constexpr RgbHtml Silver()			{ return RgbHtml(192, 192, 192); };
+	static constexpr RgbHtml DarkGray()			{ return RgbHtml(169, 169, 169); };
+	static constexpr RgbHtml Gray()				{ return RgbHtml(128, 128, 128); };
+	static constexpr RgbHtml DimGray()			{ return RgbHtml(105, 105, 105); };
+	static constexpr RgbHtml LightSlateGray()	{ return RgbHtml(119, 136, 153); };
+	static constexpr RgbHtml SlateGray()		{ return RgbHtml(112, 128, 144); };
+	static constexpr RgbHtml DarkSlateGray()	{ return RgbHtml(47, 79, 79); };
+	static constexpr RgbHtml Black()			{ return RgbHtml(0, 0, 0); };
```
salkinium: I would prefer a simple static constexpr variable, instead of function. There's no benefit to the function and it's weird to have them like this.

I also don't see a benefit of scoping them in a class. Perhaps something as simple as a `modm::color::html` namespace with these constexpr Rgb8 colors would suffice?

examples/avr/display/dogm128/benchmark/main.cpp
```diff
@@ -161,7 +161,7 @@ main()
 		while (!timer.execute())
 		{
 			// rectangle in left side of screen
-			display.setColor(Color::black());
+			display.setColor(modm::color::RgbHtml::Black());
```
salkinium: Maybe we can add a `using namespace modm::color` to these examples?

src/modm/math/utils/clip.hpp
```diff
@@ -0,0 +1,33 @@
...
+		 */
+		template <typename T>
+		inline T clip(const T min, const T value, const T max) {
+			return std::max( std::min(value, max), min);
+		}
+	}
+
+}
+
+#endif	// MODM_MATH_CLIP_HPP
```
salkinium: Why not `std::clamp`?
https://en.cppreference.com/w/cpp/algorithm/clamp

There may be some performance improvement for defining a pass-by-copy version of std::clamp (`modm::clamp`)? See issues here, but keep in mind this is LLVM on x86, may not apply to us https://secret.club/2021/04/09/std-clamp.html


But in any case that would be implemented here https://github.com/modm-io/modm/blob/develop/src/modm/math/utils/misc.hpp

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,274 @@
...
+#include <modm/io/iostream.hpp>
+#include <modm/math/utils/arithmetic_traits.hpp>
+#include <modm/math/saturated/saturated.hpp>
+
+#include "hsv.hpp"
+#include "brightness.hpp"
+#include "rgb565.hpp"
+
+template<typename _Tp>
+    concept fundamental = std::is_fundamental_v<_Tp>;
```
salkinium: Don't leak into top-level namespace. At least put it into modm.

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,274 @@
...
+		blue_sat *= factor;
+		RETURN_FROM_SATURATED_RGB;
+	}
+
+	template <fundamental _F>
+	RgbT<T> operator/(const _F factor) {
+		DECLARE_SATURATED_RGB;
+		red_sat *= factor;
+		green_sat *= factor;
+		blue_sat *= factor;
```
salkinium: /=

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,274 @@
...
+									modm::Saturated<T> blue_sat(blue)
+
+#define RETURN_FROM_SATURATED_RGB	return RgbT<T>(\
+									red_sat.getValue(),\
+									green_sat.getValue(),\
+									blue_sat.getValue())
+
+#define REASIGN_FROM_SATURATED_RGB	red = red_sat.getValue();\
+									green = red_sat.getValue();\
+									blue = red_sat.getValue()
```
salkinium: Prefix all these with `MODM_` and `#undef` them at the end of the file. (also it's `REASSIGN_`)

I forgot about `modm::Saturated`, nice use for them! However be aware that it is quite old and I'm not sure how well it's been tested outside of AVRs. It's also not optimized for ARMv7-M with DSP instructions: https://www.keil.com/pack/doc/CMSIS/Core/html/group__intrinsic__SIMD__gr.html

src/modm/ui/display/graphic_display.hpp
```diff
@@ -34,75 +35,19 @@ namespace modm
...
-
-			bool
-			operator == (const Color& other) const {
-				return (color == other.color);
-			}
-
-		private:
-			uint16_t color;
-		};
+		using Color = color::RgbHtml;
```
salkinium: Ah, I see… but using constants instead of functions will break backwards compatibility anyways. I'm ok with that.
The fix for users will be a relatively simple regex find/replace, I think that's fair.

salkinium (commented):
Nice! 🔥

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,138 @@
...
+	 * \param brightness	Brightness 'Color'-object
+	 */
+	template<typename _T>
+	constexpr Rgb565(const BrightnessT<_T> brightness) : Rgb565(RgbCalcType(brightness))
+	{}
+
+	/**
+	 * Convertion function to plain color
+	 */
+	inline operator uint16_t() const {
```
chris-durand: Do we really need an implicit conversion here? This would make the following code legal:
```
modm::color::Rgb565 rgb = ...;
rgb = 2.3f * rgb;
```
It wouldn't do the right thing but compiles. What happens is that at first `rgb` is converted to `uint16_t`, that is converted to `float`, multiplied, narrowed to `uint16_t` and implicitly converted to `Rgb565` again.

src/modm/ui/color/rgb.hpp
TomSaw: I made a similar thing few years ago. I see some improvement to be done on Saturated later.

src/modm/ui/display/graphic_display.hpp
TomSaw: This was the reason for the RgbHtml class. Much better without and if you're OK with breaking (more) stuff, i'm in!

src/modm/ui/color/rgb.hpp
TomSaw: Maybe this is the beginning of a wonderful header with more concepts!?

chris-durand: Isn't the property required here rather `std::is_arithmetic_v`? Fundamental types include `void` and `nullptr_t`.

Also, an identifier with an underscore following a capital letter like `_Tp` is [reserved to the standard library implementation](https://en.cppreference.com/w/c/language/identifier). Wouldn't ever cause practical problems used like here, but it's easily avoided by removing the `_`.

src/modm/math/utils/clip.hpp
TomSaw: Looks like this is more lightweight. Need to check later:

```
template <typename T>
inline T clamp(const T min, const T value, const T max) {
	if(value < min)
		return min;
	if(value > max)
		return max;
	return value;
}
```

src/modm/ui/color/rgb565.hpp
TomSaw: Maybe this was a drip too much of implicit comfort 🙄?

Edit: I find it natural, when a class that holds only one parameter can be handed like this plain parameter.
But in this case... the uint16_t in rgb565 does not represent a true int and should therefore not be handed out implicitly.
So dropped.

TomSaw:
Because of modm::Saturated @salkinium : [gcc Integer-Overflow-Builtins.html](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) may be the tools for a reimplementation.

Using these builtin functions, one gets overflow / underflow flags as bool in C:

> The compiler will attempt to use hardware instructions to implement these built-in functions where possible, like conditional jump on overflow after addition, conditional jump on carry etc.

src/modm/ui/color/rgb.hpp
TomSaw: Created a `concept real_number`, hope you like it.

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,283 @@
...
+	{
+		RgbT<T> result;
+		if (__builtin_add_overflow(red, other.red, &result.red))
+			result.red = numeric_limits<T>::max();
+		if (__builtin_add_overflow(green, other.green, &result.green))
+			result.green = numeric_limits<T>::max();
+		if (__builtin_add_overflow(red, other.blue, &result.blue))
+			result.blue = numeric_limits<T>::max();
+		return result;
+	}
```
TomSaw: One touch: I also reimplement modm::Saturated. Here's a quick test of a modern solution using [Integer Overflow Builtins](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) for RgbT<T>::operator+.

> The compiler will attempt to use hardware instructions to implement these built-in functions where possible, like conditional jump on overflow after addition, conditional jump on carry etc. 

**Works great + fast on all platforms (except for ARM with DSP)**

src/modm/math/utils/concepts.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <concepts>
+
+namespace modm
+{
+
+template<typename T>
+concept fundamental = std::is_fundamental_v<T>;
+
+template<typename T>
+concept arithmetic = std::is_arithmetic_v<T>;
```
salkinium: I'm not convinced that this warrants a separate header file tbh. I also don't like the added indirection, I would consider it simpler to have it like this:

```
template<typename T> requires std::is_arithmetic_v<T>
class Foo {};
```

src/modm/ui/color/brightness.hpp
```diff
@@ -0,0 +1,94 @@
...
+	constexpr BrightnessT(RgbT<_T> rgb)
+		: value((0.2125 * float(rgb.red)) + (0.7154 * float(rgb.green)) + (0.0721 * float(rgb.blue)))
+	{}
+
+	/**
+	 * Convertion Constructor for HSV Color
+	 *
+	 * \param hsv	HSV Color
+	 */
+	template<std::unsigned_integral _T>
```
salkinium:
```suggestion
	template<std::unsigned_integral U>
```

We usually use U, because it comes after T. `_` is reserved for system libs.

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+	template<modm::fundamental IntermediateType = float, unsigned int multiplier = 100,
+			 std::unsigned_integral ReturnType = T>
+	ReturnType
+	getRelative(const T color) const
+	{
+		return IntermediateType(color) * IntermediateType(multiplier) /
+			   IntermediateType(getColorSum());
+	}
```
salkinium: I feel like the multiplier should be a runtime argument. And not sure about default 100, seems arbitrary?

salkinium:
Have you [looked into the upainter colors](https://github.com/salkinium/upainter/tree/master/ges/pixel_color)? They are already implemented for a lot of formats and tested and fairly fast with some basic bit optmizations and include implementations [for all of the Porter&Duff composition operations](https://en.wikipedia.org/wiki/Alpha_compositing) ([original paper](https://keithp.com/~keithp/porterduff/p253-porter.pdf)). Some of these operations are also supported in hardware by the STM32 DMA2D, which is why I initially implemented them.

src/modm/math/utils/concepts.hpp
TomSaw: Good point too. I've proposed creating a concepts-header earlier in this PR and got a 👍 from @chris-durand ( Where is this conversation!?!? 🙄 ) I see a big future for (more complicated) widely used concepts witch deserve this header.

**Would be very clarifying if you can let your IDE list all templates using a specific concept, even the simple ones like `arithmetic`.**

src/modm/ui/color/rgb.hpp
TomSaw: Can you please give me an application example for relative colors, so i can add more head in future considerations.

TomSaw:
> Have you [looked into the upainter colors](https://github.com/salkinium/upainter/tree/master/ges/pixel_color)? They are already implemented for a lot of formats and tested and fairly fast with some basic bit optmizations and include implementations [for all of the Porter&Duff composition operations](https://en.wikipedia.org/wiki/Alpha_compositing) ([original paper](https://keithp.com/~keithp/porterduff/p253-porter.pdf)). Some of these operations are also supported in hardware by the STM32 DMA2D, which is why I initially implemented them.

I'm busy with work so this slowed down. **Let me finish up the color-tests + CI-claims so you can merge the state**. The impros fit the goal of "refactoring color" and are worth being shared with the folks! @salkinium 

These Rgb-Calculations, Saturated-stuff and upainter-integrations then receive their personal silent hours in a fresh PR.

src/modm/ui/color/rgb565.hpp
```diff
@@ -0,0 +1,145 @@
...
+		return Rgb565(0);
+	}
+
+	/// Saturated division ⊘
+	Rgb565
+	operator/(const Rgb565 other) const
+	{
+		(void)other;
+		return Rgb565(0);
+	}
```
salkinium: Not sure this makes sense? When would you multiply/divide pixels (with *integer* resolution)?
This should just be a multiplication/division with a floating point type. That would be useful for gamma correction etc.

src/modm/ui/color/rgbhtml.hpp
```diff
@@ -0,0 +1,191 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include "rgb.hpp"
+
+namespace modm::color::html
```
salkinium, line 16: You can do this since C++17 instead of nesting namespaces.

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,190 @@
...
+#include <modm/io/iostream.hpp>
+
+template<std::unsigned_integral U>
+IOStream&
+operator<<(IOStream& os, const color::RgbT<U>& color)
+{
+	os << color.red << "\t" << color.green << "\t" << color.blue;
+	return os;
+}
+#endif
```
salkinium, line 158: We typically don't depend on the `:io` module and instead guard the IO serialization implementation this way.
This keeps the `:io` module optional, since it pulls in *lot* code…

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,190 @@
...
+	{}
+
+	TSum
+	getColorSum() const
+	{
+		return red + green + blue;
+	}
+
+	template<typename IntermediateType = float, std::unsigned_integral ReturnType = T>
+			requires std::is_fundamental_v<IntermediateType>
```
salkinium: I've made this explicit, since I think it's clearer.

src/modm/math/utils/concepts.hpp
salkinium: I would wait a little until we understand concepts a little better. We typically copy paste some stuff around until someone™ gets annoyed and then refactors it into a generic library (`modm::Prescaler` for example).

src/modm/ui/color/rgb.hpp
salkinium: Good question… 

This was implemented for the Eurobot competition by @dhebbeker to distinguish between red and blue in bad lighting conditions. I'm not sure if it is particularly useful and would perhaps even just remove it from the class…
I feel like this would be better as an algorithm function that takes a RGB value and outputs a different RGB.

salkinium:
I refactored your branch a little myself instead of giving you a bazillion comments. I mostly changed style and removed some top-level `using namespace foo;` in headers which would leak into the whole code.

Apart from the relative/normalize color functions I'm happy with this now!

TomSaw:
I'll check it firstly in the morning + and gonna add these tests so we can clean the table:

Got pretty good drive today optimizing the dusty graphic algorithms witch is lots more fun.

TomSaw:
Didn't want to touch the relative/normalize stuff in Rgb since it doesn't hurt me (to much), I actually have no usage for it and especially didn't want to break someone's algos.

TomSaw:
Few more sessions general improvements on the graphics API and my head is ready to study upainter and see what we can make with all of it.

src/modm/ui/color/rgbhtml.hpp
TomSaw: Very nice.. also #pragma once

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,216 @@
...
+	T green{0};
+	T blue{0};
+
+	using TSum = modm::WideType<T>;
+
+	constexpr RgbT() = default;
+
+	constexpr RgbT(T red, T green, T blue) : red(red), green(green), blue(blue) {}
+
+	/**
```
TomSaw, line 65: Added copy constructors  for 8bit -> 16bit and vice verse for all template color types. I've played around with alternative, more generic implementations but found this simple and readable ones using _concept overloading_ the best.

8 and 16 bit colors is feels like good enough for MCUs (until the day someone builds high-dynamic cameras with 24bits and more, haha)

Test added. If you like it ~~~> Please 📀 Merge

src/modm/ui/color/rgb.hpp
```diff
@@ -0,0 +1,216 @@
...
+	 * Copy Constructor 16bit->8bit
+	 */
+	template<std::unsigned_integral U>
+	requires std::is_same_v<T, uint8_t> && std::is_same_v<U, uint16_t>
+    constexpr RgbT(const RgbT<U> &rgb_other)
+		: red(rgb_other.red >> 8),
+		green(rgb_other.green >> 8),
+		blue(rgb_other.blue >> 8)
+	{}
```
TomSaw: Added copy constructors  for 8bit -> 16bit and vice verse for all template color types. I've played around with alternative, more generic implementations but found those - using _concept overloading_ - most simple and readable.

8 and 16 bit colors feels like good enough for MCUs (until someone builds high-dynamic cameras with 24bits and more, haha)

Tests added. If you like it ~~~> Please 📀 Merge

salkinium (approved):
I extracted the weird RGB normalize into their own standalone function.

TomSaw:
> I extracted the weird RGB normalize into their own standalone function.

Wunderbar. I've been very dismotivated to do that... thanks @salkinium 🍬

TomSaw:
... instead, i've made great strides with the graphics functions. Anything has become so badly fast.. coming soon. need to cleanu and prepare a good presentation cause i wanna change the APIs.

PR #624, opened 2021-04-27
[display] Support SH1106 Display, Improve SSD1306 Display

TomSaw:
# Support SH1106

Various SSD1306 Displays sold today do actually integrate the cheaper SH1106. This thing is said to be compatible with SSD1306 what's not 100% the case: **[SH1106 does only support Page MemoryMode](https://community.arduboy.com/t/fire-panic-on-sh1106/8047/10).** thus requires a different `writeDisplay()` routine than actually used for modm::Ssd1306.

Current result using modm::Ssd1306 with SH1106

When sending a frame, [one has to alternate between setting the incremented page-address and sending page-data](https://forum.arduino.cc/t/solved-ssd1306-can-only-write-to-part-of-screen-u8glib-works-adafruit-not/256421/6).

However, SSD1306 can also be driven in `MemoryMode::PAGE` but would result in worse performance vs. transmitting the whole frame aSh1106t once like it is now, thanks to `MemoryMode::VERTICAL` / `MemoryMode::HORIZONTAL`.

### Solutions

1. Create a class `Sh1106` inheriting `Ssd1306` and override for the differences.
2. Pass `MemoryMode` as template parameter to `Ssd1306` and specialize for the differences.
3. Pass `Ssd1306_I2cWriteTransaction` as template parameter and handle the differences there.

### I've choosen 1. because:

- SH1106 is a different IC and deserves it's own class. The compatibility to SSD1306 is expressed by the strong inheritance on `Ssd1306
- Solution 3. sounds elegant but results in more complex code cause way more binding is required between `SSD1306` and `Ssd1306_I2cWriteTransaction`.
- One does not want to drive SSD1306 in `MemoryMode::PAGE` due to worse performance. This excludes 2. and 3.

# Improved SSD1306
### Structure
- Detailed Registers in own header ′ssd1306_register.hpp′
- Renamed `DataTranmissionAdapter` -> `Ssd1306_I2cWriteTransaction` for clarification
- removed numerous cumbersome `WriteData()` methods
### Performance
- Bulk data transmissions
  - Much faster Initialize() and slightly faster `writeDisplay()`
  - Size of `commandBuffer` could be halfed
- Simplified `Ssd1306_I2cWriteTransaction`

# Flipped buffer[x][y] to buffer[y][x]
The current layout of `MonochromeGraphicDisplay::buffer` is not directly compatible with the memory-order requested by SH1106. **For SH1106 we need the rows ( x-dimension) being ordered behind each other rather than the columns**.
SH1106 unfortunately offers no other options for MemoryMode to adjust for the current buffer-layout.
**Kindly Ssd1306 comes to meet the constraints of SH106.** We can switch from `MemoryMode::VERTICAL` -> `MemoryMode::HORIZONTAL` in Ssd1306::initialisation(). Now SSD1306 also receives the rows in a line so the buffer dimensions could be flipped: `buffer[x][y]` -> `buffer[y][x]`.

### DIFF
--- a/examples/arduino_nano/color/main.cpp
+++ b/examples/arduino_nano/color/main.cpp
@@ -45,7 +45,7 @@ public:
 
 		MODM_LOG_INFO << "TCS34725 responded" << modm::endl;
 
-		PT_CALL(sensor.initialize(modm::tcs3472::Enable::POWER_ON_INTERRUPT_AND_WAITTIME));
+		PT_CALL(sensor.initialize(sensor.Enable_InterruptMode_Waittime));
 		PT_CALL(sensor.configure(modm::tcs3472::Gain::X16, modm::tcs3472::IntegrationTime::MSEC_2_4));
 		PT_CALL(sensor.setInterruptPersistenceFilter(modm::tcs3472::InterruptPersistence::CNT_20));
 		// Setup WaitTime to further slow down samplerate
--- a/src/modm/driver/color/tcs3414.hpp
+++ b/src/modm/driver/color/tcs3414.hpp
@@ -54,13 +54,13 @@ struct tcs3414
 	enum class
 	Prescaler : uint8_t
 	{
-		D1	= 0b000,	///< divide by 1
-		D2	= 0b001,	///< divide by 2
-		D4	= 0b010,	///< divide by 4
-		D8	= 0b011,	///< divide by 8
-		D16	= 0b100,	///< divide by 16
-		D32	= 0b101,	///< divide by 32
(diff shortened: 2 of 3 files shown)


salkinium:
Did you mean to close this PR?

TomSaw:
I broke my branch due to bad stashing-strategy. This closed the PR and for the ease, i've just made a new one #627

salkinium:
Ok, no problem. 
Just FYI: the semester has begun and we all busy with exercises and lectures, so expect some delay in reviews…

PR #627, opened 2021-05-04
[display] Improved SSD1306, Support for SH1106

TomSaw:
# Support SH1106

Various SSD1306 Displays sold today do actually integrate the cheaper SH1106. This thing is said to be compatible with SSD1306 what's not 100% the case: **[SH1106 does only support Page MemoryMode](https://community.arduboy.com/t/fire-panic-on-sh1106/8047/10).** thus requires a different `writeDisplay()` routine than actually used for modm::Ssd1306.

Current result using modm::Ssd1306 with SH1106 [looks lke this](https://i.imgur.com/9JdtUGI.png)

When sending a frame, [one has to alternate between setting the incremented page-address and sending page-data](https://forum.arduino.cc/t/solved-ssd1306-can-only-write-to-part-of-screen-u8glib-works-adafruit-not/256421/6).

However, SSD1306 can also be driven in `MemoryMode::PAGE` but would result in worse performance vs. transmitting the whole frame at once.

### Solutions

1. Create a class `Sh1106` inheriting `Ssd1306` and override for the differences.
2. Pass `MemoryMode` as template parameter to `Ssd1306` and specialize for the differences.
3. Pass `Ssd1306_I2cWriteTransaction` as template parameter and handle the differences there.

**I've choosen 1. because:**

- SH1106 is a different IC and deserves it's own class. The compatibility to SSD1306 is expressed by the strong inheritance on `Ssd1306`
- Solution 3. sounds elegant but results in more complex code cause way more binding is required between `SSD1306` and `Ssd1306_I2cWriteTransaction`.
- One does not want to drive SSD1306 in `MemoryMode::PAGE` due to worse performance. This excludes 2. and 3 additionally

# Improved SSD1306
### Structure
- Detailed Registers in own header ′ssd1306_register.hpp′
- Renamed `DataTranmissionAdapter` -> `Ssd1306_I2cWriteTransaction` for clarification
- removed numerous cumbersome `WriteData()` methods
### Performance
- Bulk data transmissions
  - Much faster Initialize() and slightly faster `writeDisplay()`
  - Size of `commandBuffer` could be halfed
- Simplified `Ssd1306_I2cWriteTransaction`

# Changes on MonochromeGraphicDisplay
- Flipped buffer[x][y] to buffer[y][x] to be compatible with SH1106
Current layout of the `buffer` is not compatible with the memory of SH1106. **For SH1106 we need the rows ( x-dimension) being ordered behind each other rather than the columns**.
SH1106 unfortunately offers no other options to adjust the mapping. **Kindly Ssd1306 comes to meet these constraints and archieve max performance for both displays.** We can switch from `MemoryMode::VERTICAL` -> `MemoryMode::HORIZONTAL` so both displays expect the rows in a line and the common ′buffer′s dimensions could be flipped.


TomSaw:
Cool. This think looks fine to me. Could you please review @salkinium !?

src/modm/driver/display/sh1106.hpp
```diff
@@ -0,0 +1,75 @@
...
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_SH1106_HPP
+#define MODM_SH1106_HPP
+
+#include "ssd1306.hpp"
+namespace modm
+{
+/**
+ * \brief	SH1106 is said to be 'compatible' with SSD1306. However
```
salkinium: There's no need for `\brief` and the indentation.

src/modm/driver/display/sh1106.hpp
```diff
@@ -0,0 +1,75 @@
...
+ */
+template<class I2cMaster, uint8_t Height = 64>
+class Sh1106 : public Ssd1306<I2cMaster, Height>
+{
+public:
+	Sh1106(uint8_t address = 0x3C) : Ssd1306<I2cMaster, Height>(address) {}
+
+protected:
+	// TODO Generalize in Ssd1306 so no more override is required
+	inline modm::ResumableResult<void>
```
salkinium: inline not needed for template classes.

src/modm/driver/display/sh1106.hpp
```diff
@@ -0,0 +1,75 @@
...
+namespace modm
+{
+/**
+ * \brief	SH1106 is said to be 'compatible' with SSD1306. However
+ * 			there's one relevant difference: SH1106 does only support
+ * 			MemoryMode::PAGE. This requires a more extensive writeDisplay()
+ * 			routine. We have to alternate between setting Page-address and
+ * 			sending page-data instead of sending the whole 	buffer at once
+ * 			like is possible for MemoryMode::VERTICAL or MemoryMode::HORIZONTAL.
+ */
```
salkinium, line 27: `@ingroup modm_driver_sh1106`?

src/modm/driver/display/ssd1306.hpp
```diff
@@ -193,66 +126,67 @@ class Ssd1306 : public ssd1306, public MonochromeGraphicDisplayVertical<128, Hei
...
+	inline modm::ResumableResult<bool>
 	setDisplayMode(DisplayMode mode = DisplayMode::Normal)
-	{ return writeCommand(static_cast<Command>(mode)); }
+	{
+		commandBuffer[0] = mode;
+		return writeCommands(1);
+	}
 
-	modm::ResumableResult<bool> modm_always_inline
+	inline modm::ResumableResult<bool>
```
salkinium: no inline, no `modm_always_inline`.

src/modm/driver/display/ssd1306.hpp
```diff
@@ -193,66 +126,67 @@ class Ssd1306 : public ssd1306, public MonochromeGraphicDisplayVertical<128, Hei
...
 protected:
-	/// Write a command without data
 	modm::ResumableResult<bool>
-	writeCommand(uint8_t command);
+	writeCommands(std::size_t length);
 
-	/// Write a command with one byte data
-	modm::ResumableResult<bool>
-	writeCommand(uint8_t command, uint8_t data);
+	virtual inline modm::ResumableResult<void>
```
salkinium: nope

src/modm/driver/display/ssd1306.hpp
```diff
@@ -15,74 +16,20 @@
 
 #include <modm/architecture/interface/i2c_device.hpp>
 #include <modm/architecture/utils.hpp>
+#include <modm/math/utils/bit_constants.hpp>
```
salkinium: Is this even used in this file?

src/modm/driver/display/ssd1306.hpp
```diff
@@ -94,40 +41,28 @@ struct ssd1306
...
-		DataTransmissionAdapter(uint8_t address);
-
-		void modm_always_inline
-		setCommandBuffer(uint8_t *buffer)
-		{ commands = buffer; }
+		Ssd1306_I2cWriteTransaction(uint8_t address);
 
 		bool
-		configureDisplayWrite(uint8_t (*buffer)[Height / 8], std::size_t size);
+		configureDisplayWrite(const uint8_t *buffer, const std::size_t size);
```
salkinium:
```suggestion
		configureDisplayWrite(const uint8_t *buffer, std::size_t size);
```
Pass-by-value is always a copy, hence const is redundant.

src/modm/driver/display/ssd1306_impl.hpp
```diff
@@ -10,64 +10,103 @@
...
+	TransactionSuccess &= RF_CALL(writeCommands(7));
+
+	RF_CALL(initializeMemoryMode());
+
+	commandBuffer[0] = TimingAndDrivingCommands::ChargePump;
+	commandBuffer[1] = ChargePump::V7_5;
+	commandBuffer[2] = HardwareConfigCommands::SegmentRemap127;
+	commandBuffer[3] = HardwareConfigCommands::ComOutputScanDirectionDecrement;
+	commandBuffer[4] = HardwareConfigCommands::DisplayStartLine;
+	commandBuffer[4] |= 0;
```
salkinium: ?

src/modm/driver/display/ssd1306_impl.hpp
```diff
@@ -10,64 +10,103 @@
...
 modm::ResumableResult<bool>
 modm::Ssd1306<I2cMaster, Height>::initialize()
 {
 	RF_BEGIN();
+	TransactionSuccess = true;
+
+	commandBuffer[0] = FundamentalCommands::DisplayOff;
+	commandBuffer[1] = TimingAndDrivingCommands::DisplayClockDivideRatio;
+	commandBuffer[2] = 8 << 4;  // Frequency (influences scrolling speed too)
+	commandBuffer[2] |= 0;      // Prescaler
```
salkinium: ?

src/modm/driver/display/ssd1306_impl.hpp
```diff
@@ -80,105 +119,61 @@ modm::Ssd1306<I2cMaster, Height>::writeDisplay()
...
+		commandBuffer[5] = 0x00;
+		commandBuffer[6] = 0xFF;
 	}
 
-	RF_WAIT_UNTIL( startTransactionWithLength(14) );
-
-	RF_WAIT_WHILE(this->isTransactionRunning());
-
-	RF_END_RETURN(this->wasTransactionSuccessful());
+	return writeCommands(7);
```
salkinium: Without RF_BEGIN/END, the whole function is executed again for every polling call!

src/modm/driver/display/ssd1306_impl.hpp
```diff
@@ -80,105 +119,61 @@ modm::Ssd1306<I2cMaster, Height>::writeDisplay()
...
+	} else if (orientation == glcd::Orientation::Landscape180)
 	{
-		RF_RETURN_CALL(writeCommand((orientation == glcd::Orientation::Landscape0) ?
-				Command::SetComOutputScanDirectionDecrement : Command::SetComOutputScanDirectionIncrement));
+		commandBuffer[0] = HardwareConfigCommands::SegmentRemap0;
+		commandBuffer[1] = HardwareConfigCommands::ComOutputScanDirectionIncrement;
 	}
 
-	RF_END_RETURN(false);
+	return writeCommands(2);
```
salkinium: Same here: RFs without the macros should only be used for forward convenience functions without logic.

src/modm/driver/display/ssd1306_impl.hpp
```diff
@@ -80,105 +119,61 @@ modm::Ssd1306<I2cMaster, Height>::writeDisplay()
 	RF_END_RETURN(this->wasTransactionSuccessful());
 }
 
-template < class I2cMaster, uint8_t Height >
-modm::ResumableResult<bool>
+template<class I2cMaster, uint8_t Height>
+inline modm::ResumableResult<bool>
```
salkinium:
```suggestion
modm::ResumableResult<bool>
```

salkinium: inline is only necessary for non-templated classes with method implementations in a header. Otherwise the method implementation is generated in multiple compile units and leads to linker errors. Aka. one definition rule.

src/modm/driver/display/ssd1306_impl.hpp
TomSaw: Ides is to communicate in code, that an integer could be set here. In this case is 0.

src/modm/driver/display/ssd1306_impl.hpp
TomSaw: Same 🙄

src/modm/driver/display/ssd1306_impl.hpp
TomSaw: Check 🙌  also found more on this online...

src/modm/ui/display/monochrome_graphic_display_vertical_impl.hpp
```diff
@@ -16,121 +16,88 @@
...
 modm::MonochromeGraphicDisplayVertical<Width, Height>::drawVerticalLine(glcd::Point start,
 																		uint16_t length)
 {
-	if (start.x >= 0 and start.x < Width)
+	if (this->xOnScreen(start.x))
 	{
-		const int8_t end_y = start.y + length;
-		const uint8_t y_last = end_y / 8;
+		size_t yb = start.y / 8;
+		const int16_t y_max = std::min<size_t>((start.y + length), Height);
```
salkinium:
```suggestion
		const size_t y_max = std::min<size_t>((start.y + length), Height);
```

src/modm/ui/display/monochrome_graphic_display.hpp
```diff
@@ -76,7 +76,28 @@ class MonochromeGraphicDisplay : public GraphicDisplay
...
+	inline bool
+	xValidBuffer(const std::size_t x) const
+	{
+		return x >= 0 and x < BufferWidth;
+	}
+	inline bool
+	yValidBuffer(const std::size_t y) const
+	{
+		return y >= 0 and y < BufferHeight;
+	}
```
salkinium: Inline police ;-p

salkinium (approved):
Very nice! Perhaps split into two commits, one for the display changes and one for the driver code.

src/modm/ui/display/monochrome_graphic_display.hpp
TomSaw: Honestly??? But is not a template methode and should really be inlined! (Is still don't get it, propably)

salkinium: Yes, but the class of the method is a template, therefore everything in here is implicitly marked `inline`. This isn't really important in this case, the compiler just ignores the duplicate `inline`, you can leave it like that.

TomSaw:
Clean. if you're happy, merge pls!

TomSaw:
Ready to merge, sir!

PR #636, opened 2021-05-27
[examples] blue-pill ITM logging

strongly-typed:


examples/blue_pill_f103/itm/main.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+	LedGreen::set();
+
+	while (true)
+	{
+		static modm::PeriodicTimer tmr{500ms};
+
+		if (tmr.execute())
+		{
+			tmr.restart(500ms);
```
salkinium: Not necessary for a periodic timer?

rleh: Restarting the periodic timer with every `tmr.execute()` event will create al lot jitter. (This is not critical here, but from my experience example code is copied and extended very often...)

strongly-typed: You are right, this was copied from the F303 example. But this example uses
```C++
tmr.restart(Button::read() ? 100ms : 500ms);
```
Fixed.

rleh (approved):

salkinium (approved):

strongly-typed:
Does the CI hang again? The `*-compile-all` jobs did not report their status ...

LGTM!

salkinium:
No, I just went to bed before merging.
The compile all is triggered manually now, but not necessary here.

strongly-typed:
Thanks! I was just confused by the checks that were not triggered.

PR #637, opened 2021-05-28
[tests] Use MODM_ARRAY_SIZE()

strongly-typed:
Code actually was incorrect and triggered the following warning:
```
modm-test/src/modm-test/math/filter/moving_average_test.cpp: In member function 'void MovingAverageTest::testFloatAverage()':
modm-test/src/modm-test/math/filter/moving_average_test.cpp:114:52: warning: expression does not compute the number of elements in this array; element type is 'const {anonymous}::TestData', not '{anonymous}::TestDataFloat' [-Wsizeof-array-div]
  114 |         for (uint_fast8_t i = 0; i < (sizeof(data) / sizeof(TestDataFloat)); ++i)
      |                                       ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
```

This bug was buried for some years now I would guess ;-)

### DIFF
--- a/test/modm/math/filter/moving_average_test.cpp
+++ b/test/modm/math/filter/moving_average_test.cpp
@@ -100,7 +100,7 @@ MovingAverageTest::testAverage()
 {
 	modm::filter::MovingAverage<TestData::Type, 4> filter;
 
-	for (uint_fast8_t i = 0; i < (sizeof(data) / sizeof(TestData)); ++i)
+	for (uint_fast8_t i = 0; i < MODM_ARRAY_SIZE(data); ++i)
 	{
 		filter.update(data[i].input);
 		TEST_ASSERT_EQUALS(filter.getValue(), data[i].output);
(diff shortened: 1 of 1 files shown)


rleh:
To make modm code as reusable and less special as possible I would be in favor of using `std::array<>::size()` in the long run, buit that's not the scope of this PR.
(And also nobody likes macros 😆)

rleh (approved):

rleh:
I don't think we need to run the full HAL tests for this PR, or do you disagree @salkinium?
🐧🐧

strongly-typed:
Agreed.

There are some more occurrences of the `sizeof / sizeof` usage.
```
communication/sab/slave.hpp:168:		 *             sizeof(actionList) / sizeof(modm::sab::Action));

platform/clock/stm32/rcc_impl.hpp.in:
34:	uint8_t lut_size(sizeof(flash_latency_{{ (table|sort)[0] }}) / sizeof(uint32_t));
40:		lut_size = sizeof(flash_latency_{{ mV }}) / sizeof(uint32_t);

```

That could be a nice refactoring.

chris-durand:
Just in case someone attempts to do the refactoring, the preferred way of determining the element count of a C array in C++ is using a template function taking a reference to an array and deducing the size:
```
template <class T, std::size_t N>
constexpr std::size_t size(const T (&array)[N]) noexcept
{
    return N;
}
```
Since C++20, this is available as an overload of `std::size` as defined above. This is preferable to `MODM_ARRAY_SIZE` because the latter will happily compile when passed a pointer type but will report a wrong size. `(sizeof(x) / sizeof(x[0]))` would always evaluate to 4 on a 32 bit machine when passed a `char*` instead of a char array.

PR #639, opened 2021-06-03
[amnb] Use modm::inplace_function for handlers

salkinium:
How much UB do I want? Yes.

There are two things I want:

- An fixed-size, non-allocating [`std::inplace_function`](https://github.com/WG21-SG14/SG14/blob/master/SG14/inplace_function.h). I also [checked `etl::function`](https://www.etlcpp.com/function.html), but it's not clear to me if they can store a closure.
- Copy the function to a type-erased storage, and then cast it back to the specific type. I tried a union, but it doesn't work with `std::function` because of the destructor.

cc @chris-durand whelp, how do I do this correctly?


chris-durand:
> I also checked etl::function, but it's not clear to me if they can store a closure.

Looked at the code, it is non-owning. It can be constructed with a member function and stores a pointer to the object.
The same is true for its non-deprecated successor `etl::delegate`.

So you want:
> - An fixed-size, non-allocating std::inplace_function
> - Copy the function to a type-erased storage

and you want completely type-erased storage for function objects with arbitrary signatures.

Are you aware of [`std::any`](https://en.cppreference.com/w/cpp/utility/any)? It can store any copyable object.  For small objects it won't allocate and stores the object in an internal buffer. We could make an inplace_any that would never allocate nor throw exceptions but assert at compile time if the object doesn't fit. The implementation is [a bit nifty](https://github.com/llvm-mirror/libcxx/blob/master/include/any) but it's totally doable. I might try tomorrow. Put an inplace_function inside of it and you should have what you need. I presume performance wise it could be good enough.

chris-durand:
I have implemented this inplace_any type: https://github.com/chris-durand/modm/tree/feature/inplace_any
Have a look at the [unit test](https://github.com/chris-durand/modm/blob/feature/inplace_any/test/modm/utils/inplace_any_test.cpp#L62) on how to use it. The default of maximum size 16 was chosen quite randomly, but it can be changed with a template parameter. Tell me if you find it useful, then I'll do a merge request and add doxypress documentation.

It doesn't have the throwing overloads of `any_cast` like `std::any` that return a reference to the contained object. Wasn't sure if it is a good idea to have them with modm_assert on error or not.

salkinium:
Your `modm::inplace_any` works like a charm, thank you very, very much!

I'm struggling with `modm::inplace_function`, which I copied from SG14. My main concern is not being able to do just `modm::inplace_function(lambda)` and have it derive the signature automatically, I still need to manually define it. 

Is there a better way to do this in general? Perhaps an overload for lambdas (enabled with concepts)?

salkinium:
Ok, the issue seem to be the overloading of the argument type for callbacks, which it has trouble to deduct?

```cpp
template<typename T>
Listener(int command, function<void(uint8_t, const T&)> listener)
```

I played around with capturing lambda's directly via template type:
This works well in general, until I get to the handler's argument type again:

```cpp
template< typename F >
Listener(uint8_t command, F &&handler)
    requires std::convertible_to<decltype(handler), function<void(uint8_t)>>;

template< typename F, typename T >
Listener(uint8_t command, F &&handler)
    requires std::convertible_to<decltype(handler), function<void(uint8_t, const T&)>>;
```

I tried to find a concept for any callable type with a specific signature, which is really what I want here. Is this something we could build?

chris-durand:
> I tried to find a concept for any callable type with a specific signature, which is really what I want here. Is this something we could build?

A concept won't help in that case. The issue is deducing the template argument `T`. The compiler has no way know which `T` to instantiate the template class `function<void(uint8_t, const T&)>` with to check the concept. The only way is looking at the passed handler type and figuring it out from there manually.

A lambda with a capture is basically an anonymous struct with an `operator()` defined. You could get `decltype(&F::operator())` which is a member function pointer like `void (F::*)(uint8_t, const int&)` for a matching handler and go from there. Very ugly stuff... I will try to come up with something practicable later.

chris-durand:
I have something that could work, at least it compiles: https://github.com/chris-durand/modm/tree/feature/amnb_std_function
This was the least ugly solution I could come up with ...

I defined a helper type trait class to deduce the function signature from the callable, e.g. from `[&](int){}` to `void(int)` and defined a function overload set on the deduced signature to return the appropriate `redirect` function in the `Listener` class. With that solution the implicit conversion to `inplace_function<...>` for user handlers should work.

salkinium:
I've tested your changes in hardware and it works! Thanks a lot, this is such a usable solution!
I want to see how much we can use `modm::inplace_function` for things like universal callbacks in the HAL to for example remove the buffers from UART etc, so this is a good experience.

salkinium:
I found an issue with using `modm::inplace_any`/`std::any`: mutable lambdas which modify their closure state don't maintain that state, since the closure is copy constructed new on every single call of the redirect.

I've hacked around with `std::variant` now, which works fine, until it gets to the `typename T` issue again, since it won't let me match that type for obvious reasons. However, I've now `reinterpret_cast` dat shit, and that works very well with keeping closure state.

How acceptable is that in terms of UB? I looked at the inplace_function and as far as I understand `Args...` isn't a relevant part of any constructor call.

chris-durand:
> I found an issue with using modm::inplace_any/std::any: mutable lambdas which modify their closure state don't maintain that state, since the closure is copy constructed new on every single call of the redirect.

This is easily fixed. Use `any_cast` on `inplace_any` which does not copy anything. `get` returns a `std::optional` of the contained value which copies it. 

> However, I've now reinterpret_cast dat shit, and that works very well with keeping closure state.
> How acceptable is that in terms of UB? I looked at the inplace_function and as far as I understand Args... isn't a relevant part of any constructor call.

That is totally UB, variant is the wrong abstraction there. I modified the `inplace_any` solution to accomplish what you want: https://github.com/chris-durand/modm/commit/3a58c866293236241378ba6bfeb41a3eca096a79

salkinium:
> This is easily fixed.

You really weren't kidding, thanks!

It works now just as expected and it's absolutely fantastic! Definitely want to use more modern C++ in modm now. 😍
I'll add some more docs before merging, since it's a bit… meager.

chris-durand:
Please let me know when you intend to merge this. I'll have to add doxygen documentation to `inplace_any`.

salkinium:
Beginning next week probably. I also want to add more Doxygen to AMNB.

salkinium:
> I'll have to add doxygen documentation to inplace_any.

Kind reminder to do this before the 29th please. I personally just added the right doxypress tags, I'm leaving the proper module documentation up until I get a better understanding of `modm::inplace_function`.

salkinium:
Kind reminder @chris-durand to finish the remaining tasks today or tomorrow please.

chris-durand:
> Kind reminder @chris-durand to finish the remaining tasks today or tomorrow please.

I was very busy with work today and am on it. Expect it to be finished by tomorrow evening.

chris-durand:
I found some smaller issue with inplace_any. It also has to use the SG14 implementation of `std::aligned_storage` to avoid the gcc/libstdc++ bug that causes excessive sizes for types smaller than the maximum alignment.  So I will factor out the aligned storage class to a separate file.

chris-durand:
Done, the changes are there: https://github.com/chris-durand/modm/tree/feature/amnb_std_function

I adapted the interface of `inplace_any` to closer resemble `std::any`. We also have an asserting overload of `any_cast` now that returns a reference instead of a pointer. @salkinium You could use it in the amnb implementation in cases you don't check for nullptr anyway. It could be slightly less efficient though since we get one branch more for the assert.

salkinium:
Thank you, @chris-durand! (I've decided to keep using the pointer version, since there's no real chance of an empty storage.)

PR #645, opened 2021-06-24
scons output path extension

kikass13:
## related issue: 
 * https://github.com/modm-io/lbuild/issues/65

## what does it do
when invoking `lbuild --path build build` we cannot call `scons` directly afterwards, because the scons template generation is buggy (the new export path 'build' is not propagated throuh)

I added some changes inside the modm:scons module which can be activated via an option (`link_scons_to_pwd`). This option turns on this feature, which is deactivated by default. It will change some paths around inside the scons template generation, to fit with the new `--path build` export path, as well as creating the `SConstruct` file inside the current pwd, so that it can be called like the "normal" modm + lbuild workflow suggests:

 * lbuild --path <DIR> build
 * scons

I tested this feature with one of my workspaces and it seems to work, but I guess someone here can see the mistakes i made and give me some tips on what to do better :)


salkinium:
I'm a little skeptical: Alternatives include calling `scons -C build` or using a custom SConstruct with the paths changed.
Is there a specific reason why you are generating the modm sources inside a build folder? What's wrong with having the modm folder in your app folder?

kikass13:
We are planning on using modm repositories (+internal modules) extensively and these will clutter up the workspace (file tree). [N~15]

Fixing the scons file in place is also bad, because then we loose the nice "generate and forget" design of the current method.

Your command:
`scons -C build` will not find my application sources or includes .. except for when these are copied into the build dir as well(there is no toolchain in modm for that)  or if I add a fixed SConstruct file and add that to my `scons` command. This is not a viable option. `Build`, `Gen` and `SRC/include` dirs (as well as their separation) are nothing fancy .. in fact, that is the default use case in my opinion.

The fact that modm is not doing(not eben allowing) the  separation of generated code and application code without manual intervention is not really helpful :)

Just for clarification: what is wrong with extending the scons code generation to have more flexibility? 
The code generator is way too simple and should do/allow all kinds of things anyways.

 Everything that does not need manual intervention for edge cases is good. Manual intervention makes the modm:scons module useless, because it defeats the point of generating the scons file in the first place. It also kills the use case,  that allows me to change the underlying build system.

### TLDR
* Manual change if SConstruct files defeats the point of SConstruct generation
* the change I made is non intrusive (as should be)
* the scons generator is not a holy thing, it should support way more things in general, as it is the only build system wrapper/prepper modm has right now

salkinium:
> what is wrong with extending the scons code generation to have more flexibility?

We'd also need to adapt the `:build:make` and `:build:cmake` modules. Ideally this would need support from lbuild to introduce and document a "cwd path" to copy files to. The `--path` cli option would need to be added to the `project.xml` settings to make sense. Plus unit tests. So the whole thing is already quite a lot of unpaid work for me 🙃.

> Manual intervention makes the modm:scons module useless, because it defeats the point of generating the scons file in the first place. It also kills the use case, that allows me to change the underlying build system.

No, it doesn't make it useless. The top-level SConstruct is intended to be replaced with your own. [From the docs](https://modm.io/reference/module/modm-build-scons/):

>> We do not intend to serve every possible use-case with this module. If you need something special, write your own SConstruct file, maybe starting by modifying ours. It is intentionally BSD-licensed so that you do not have to publish your changes to it.

> The code generator is way too simple and should do/allow all kinds of things anyways.
> makes the modm:scons module useless
> Modules should be agnostic to other modules or their repositories by design.
> Querie designs are almost never needed
> Thats a little backwards don't you think?

You tend to make a lot of blanket statements about other people's work and it's not getting you any favors from over here.

The thing that annoys me is that the solution to this somewhat minor path inconvenience is to use a custom SConstruct and commit it. It'll cost you a few minutes to copy it and fix the paths and a few minutes a week to maintain it. You will most likely need a custom SConstruct *anyways* to add things like unit tests, adapt compile flags, integrate custom tools etc.

However, instead of doing that you're now asking me to implement a whole new feature in lbuild or to maintain this hack in modm. This solves your problem by making it mine.

kikass13:
> No, it doesn't make it useless. The top-level SConstruct is intended to be replaced with your own

Im fine with you saying that (it's your design choice after all), but having a 200 line module that generates 30 lines of SConstruct, just to be replaced and disregarded afterwards(and convert a build tool agnostic design into a fixed build tool in the process) is exactly the point for the argument I'm making.

I understand and accept you're choice of not wanting this 'hack'. Design choices are not a simple matter and my attacks on some of them shouldn't be taken personally (it's never a good idea to take self designed stuff too seriously against criticism, even though the criticism seems unjustified :) ). So I'm sorry if I sound too rude, the intention is to have a quick expression of personal taste without the need of beating around the bush (and waste time) :)

Thanks for the time anyways :)
I will try having a self made SConstruct script under version control :)

daixtrose:
> What's wrong with having the modm folder in your app folder?

IMHO everything. Mixing the output of the build system with the code under version control leads to many problems in large code bases with many dependencies. 

Also I think that the current PR does not really add to the complexity, so why not support that feature? For the CMake target you need it anyway.

salkinium:
A proper solution looks like this:

1. Make the `--path` option available inside lbuild and allow to specify it in the `project.xml` config (currently you have to manually call `lbuild --path build build` every single time, which is not great either…)
2. Add the concept of a `cwdpath("SConstruct")` output path into lbuild (analog to `localpath()/repopath()`) and a way to get the relative path to the `--path`.
3. Use these to generate the SConstruct/Makefile.
4. Niklas is happy.

I think your use-case is perfectly valid and this is a good solution to the problem, however, you're basically assigning technical debt to me to fix this properly in the future. So I'm swinging my uhm… maintainership around to force you to do this right (or to live with the SConstruct hack).

kikass13:
> 1. Make the `--path` option available inside lbuild and allow to specify it in the `project.xml` config (currently you have to manually call `lbuild --path build build` every single time, which is not great either…)

> 2. Add the concept of a cwdpath("SConstruct") output path into lbuild (analog to localpath()/repopath()) and a way to get the relative path to the --path.

 * https://github.com/modm-io/lbuild/pull/66

> 3. Use these to generate the SConstruct/Makefile.

 * See incoming commit 
     * https://github.com/modm-io/modm/commit/e769260a08fb4f030ec3dfdaa96cae8e7e496b4f
     
> 4. Niklas is happy.

 * well that's all we care about after all :)

kikass13:
I tested these changes with a local project and the following things:

 *  lbuild --path export build && scons
```xml
no special project.xml
```

lbuild and scons work fine, repository exports are in "export"

 *  lbuild --path export build && scons
```xml
 <options>
   <option name="lbuild:path"> </option>
  </options>
```
lbuild and scons work fine, repository exports are in "export"
modm:build.path has not changed, these compile outputs are in ./build/<myApplicationName>/

 *  lbuild --path export build && scons
```xml
 <options>
   <option name="lbuild:path"> build </option>
  </options>
```

lbuild and scons work fine, repository exports are in "build"
build option inside .xml has overwritten command line arg
modm:build.path has not changed, these compile outputs are in ./build/<myApplicationName>/ together with the exported repository contents

 *  lbuild build && scons
```xml
 <options>
    <option name="lbuild:path"> gen/stuff/xoxo </option>
  </options>
```
lbuild and scons work fine, repository exports are in "gen/stuff/xoxo"
modm:build.path has not changed, these compile outputs are in ./build/<myApplicationName>/

salkinium:
Much better, this I like!

salkinium:
I guess this is a lot more work than I thought…

kikass13:
Why that ?

salkinium:
Look at my changes… Lot's of paths need changing.

kikass13:
but these are now for make file generation and stuff ... should'nt these get their own PR at some point?
I mean, it is not necessary to add this feature to the other generators right now because it wasn't there in the first place anyways? :D

I mean i don't mind, but it's probably (as you put it) "a lot of extra work"

salkinium:
No, I want all build systems to have feature parity, otherwise I go crazy with remembering all the caveats. I think it's a good idea to be able to move the entire modm folder outside of the application folder to have a clear "license border". That's the argument that convinced me.

kikass13:
> No, I want all build systems to have feature parity, otherwise I go crazy with remembering all the caveats. I think it's a good idea to be able to move the entire modm folder outside of the application folder to have a clear "license border". That's the argument that convinced me.

can i help in any way? :) 

EDIT: or am I  just a burden when learning all the oder modules as well? :D

salkinium:
@kikass13 Please test if these changes work for you.

salkinium (approved):

salkinium:
@kikass13 is this change still relevant for you?

kikass13:
@salkinium oh sorry, did not notice the response.

Sure, I very much like to utilize this (as well as the repository glob) immediately for our firmware projects :) 

I'll test it tomorrow :)

kikass13:
@salkinium it seems to work fine :)

I like it very much ! :+1:

salkinium (approved):
Let's do this!

PR #652, opened 2021-07-06
Add STM32H7 support

salkinium:
Felt cute might merge later.
I'm testing in hardware on NUCLEO-H743ZI, which boots with a 64MHz HSI, so that's pretty fast.
Depends on https://github.com/modm-io/modm-devices/pull/63.

- [x] Several corrections to modm-devices data regarding peripheral types
- [x] Fixup linkerscript .heapN/.data issues with RAM vs D1_SRAM{1-3} vs D2_SRAM{1-3} vs D3_SRAM
- [x] NUCLEO-H743ZI BSP with clock config for 400MHz
- [x] NUCLEO-H723ZI BSP with 550MHz clock
- [x] CI integration
- [x] Testing
- [x] Ignore dual-core devices for now
- [x] Fix RCC prescaler definitions
- [x] Fix Flash Latency
- [x] Fix SysTick at 550MHz. Running at 5Hz instead of 4Hz?
- [x] SysTick not /8 on revY, see errata
- [x] RCC Overdrive mode
- [x] RCC Apb3/4 Prescaler
- [x] RCC Pll implementation
- [x] RCC power plus initialization before SRAM access
- [x] Linkerscript refactoring


salkinium:
We need to limit support to single core devices and then deal with dual-core devices later.

chris-durand:
I have tested this on a Nucleo H723ZG and am trying to get the PLL to work. The RCC PLL implementation is not correct, but I expect to have a fix soon. The LED is blinking already at the right rate and I have uart output. I am currently adding settings for the APB3/4 prescaler.

The flash latency data is also not correct yet. For higher CPU frequencies the assertion always triggers that the frequency was not supported. And I get an assertion from the systick driver: "HLCK is too fast for the SysTick to run at 4Hz"

salkinium:
Please wait, I've already fixed a bunch of stuff incl. inside modm-devices. The RCC implementation isn't from me and it's completely wrong for the PLL at least for my H743.

salkinium:
> And I get an assertion from the systick driver: "HLCK is too fast for the SysTick to run at 4Hz"

Yes, the fractional 32x32=64 multiplication overflows at ~536MHz and the H723 is running at 550MHz… When I implemented that the H7 only had the 480MHz devices… grrr

salkinium:
I've pushed my fixes, but haven't tested the PLL config yet. Just so you're synced up.

chris-durand:
> I've pushed my fixes, but haven't tested the PLL config yet. Just so you're synced up.

Thanks, I will look into it tomorrow. I noticed that in the H743 BSP the Apb* clocks are 240 MHz, the maximum is 120 MHz.

salkinium:
> the Apb* clocks are 240 MHz, the maximum is 120 MHz.

Yup, had to look into the datasheet for that. ST's docs are a little crappy…

I can get my H743 up to 240MHz, so the PLL implementation is correct, however the µC crashes when I try 480MHz. I suspect my `enableOverdriveMode()` implementation, see Power Control -> Voltage Scaling in the RefMan, perhaps you can try on your board.

chris-durand:
> I can get my H743 up to 240MHz, so the PLL implementation is correct, however the µC crashes when I try 480MHz. I suspect my `enableOverdriveMode()` implementation, see Power Control -> Voltage Scaling in the RefMan, perhaps you can try on your board.

Yes, same for me. Around 300 MHz was the maximum with default power settings.

I could get the H723 to run at 500 MHz with:
```
PWR->CR3 = PWR_CR3_LDOEN;
while(!(PWR->CSR1 & PWR_CSR1_ACTVOSRDY));

PWR->D3CR &= ~PWR_D3CR_VOS;
while(!(PWR->D3CR & PWR_D3CR_VOSRDY));
```
It still hangs at 550 MHz. The overdrive bit does not exist for the H723.

I found a mistake in the RCC implementation, the [Apb3 prescaler setting](https://github.com/modm-io/modm/blob/1df923add4fab55d8e510d85fefc1e3ef6ac8c89/src/modm/platform/clock/stm32/rcc.hpp.in#L213) in the RCC driver configures the Apb4 prescaler:
apb4

The setting for Apb3 is in the D1 clock domain configuration register:
```
enum class
Apb3Prescaler : uint32_t
{
    Div1   = RCC_{{cfgr1}}_D1PPRE_DIV1,
    Div2   = RCC_{{cfgr1}}_D1PPRE_DIV2,
    Div4   = RCC_{{cfgr1}}_D1PPRE_DIV4,
    Div8   = RCC_{{cfgr1}}_D1PPRE_DIV8,
    Div16  = RCC_{{cfgr1}}_D1PPRE_DIV16
};

static inline bool
setApb3Prescaler(Apb3Prescaler prescaler)
{
	RCC->{{cfgr1}} = (RCC->{{cfgr1}} & ~RCC_{{cfgr1}}_D1PPRE) | uint32_t(prescaler);
	return true;
}
```

chris-durand:
According to the H723 reference manual we have to make sure no RAM is written until some power settings are configured:
h7_pwr

salkinium:
> The setting for Apb3 is in the D1 clock domain configuration register:

Oops, thanks! I'll fix it.

> PWR->CR3 = PWR_CR3_LDOEN; while(!(PWR->CSR1 & PWR_CSR1_ACTVOSRDY));

Ah, the internet said something about running from LDO, I'll check that.

I found this horror story: [Migration from RevY to RevV for STM32H743/753 and STM32H750 Value line microcontrollers](https://www.st.com/content/ccc/resource/technical/document/application_note/group1/95/22/7c/0c/57/de/4b/f9/DM00609692/files/DM00609692.pdf/jcr:content/translations/en.DM00609692.pdf)



[Errata says the same thing](https://www.st.com/resource/en/errata_sheet/dm00368411-stm32h742xig-and-stm32h743xig-device-limitations-stmicroelectronics.pdf)

salkinium:
> According to the H723 reference manual we have to make sure no RAM is written until some power settings are configured:

Ok, then I'll put the LDO stuff into `platform_startup.c`.

salkinium:
We don't have revision encoded into the device id, so there's no way to query that sadly. ~We should probably think about adding revision in modm-devices.~ EDIT: eh, way too much work

chris-durand:
> Ok, then I'll put the LDO stuff into `platform_startup.c`.

Which is fine unless someone is using the switch mode supply configuration. There are different options how to use the hardware:
pwr

Shouldn't we make it an lbuild option?

chris-durand:
Only some H7 have the SMPS option (H7x5, some H730, some H7Ax/H7Bx).

salkinium:
I can't get it to work regardless of what I try. I'm now going to leave the BSP running at 240MHz and someone more motivated than me can fix it later. I've upgraded you to maintainer so you should be able to write to my fork's branch to add your BSP and test the CI etc.

chris-durand:
> I can't get it to work regardless of what I try. I'm now going to leave the BSP running at 240MHz and someone more motivated than me can fix it later.

One last easy thing we might want to try is to also add the Apb4 prescaler. It boots on 1 and without setting it to 2 you will be overclocking the peripherals by a factor of 2 at 480 MHz CPU / 240 MHz AHB clock. I have already added it and will push it later when I have time.

> I've upgraded you to maintainer so you should be able to write to my fork's branch to add your BSP and test the CI etc.

Thank you!

salkinium:
Your tip about the prescaler was correct! If I set the prescalers *before* switching the SystemClock, everything works just fine (@ 400MHz though). That's awesome, but we should probably fix this issue on all other BSPs to avoid copy-paste-errors in the future.

chris-durand:
I have added a function in the RCC driver to set the voltage scaling mode to be consistent across all H7. The overdrive mode bit is only available on H74* and H75*. The H74* reference manual also calls the highest voltage mode with overdrive "Scale 0" or "VOS0", consistent with the other manuals. It can be activated with `Rcc::setVoltageScaling(Rcc::VoltageScaling::Scale0);` for all H7 now.

The CI should hopefully pass.

chris-durand:
Oh ST, I guess some senior marketing specialist really had to rename the D3 control register to "SmartRun domain control register" for H7(A|B) and shuffle all the bits. Would have been way too consistent otherwise.

chris-durand:
> Ah yes, this completely pointless problem again. You can use `"d3"` instead, see below.

done

chris-durand:
CI passed, I'll later try to run the H723 at 550 MHz with CubeMX and if it works, look what they do differently .

salkinium:
I tried and failed to get TinyUSB working on this board. I leave it to someone else to deal with this.

salkinium:
Also really awesome bug but *only in revision Y*:


I couldn't find any revision data in the CubeMX DB so I'm thinking about adding a `modm:revision` repo option with just an (optional) string input like `"Y"` that the modules can query additionally. Technically the revision is not part of the identifier either, it's just a separate, floating character on the package, so it's perhaps the best option for now.

salkinium:
> adding a modm:revision repo option

Done this now. I think it's not terrible and probably the best we can do. What do you think? @chris-durand @rleh

salkinium:
> adding a modm:revision repo option

I've changed this to append the revision to the `modm:target` string and then splitting it in the custom `TargetOption` and adding it to the device identifier manually. So this: `stm32h743zit6/revY` can now simply be queried as part of the device id: `env[":target"].identifier.revision`. What do you think about the `/rev` separator? Do you know of a more common format?

chris-durand:
> What do you think about the /rev separator? Do you know of a more common format?

I like that, don't have a better idea right now. 

Finally I managed to get the H723 to run at full 550 MHz speed. The voltage scaling level had to be set before enabling the PLL. It seems it is powered from the core voltage as well and can't start up at 550 MHz with VOS3 voltage.

chris-durand:
We should configure the [power source](https://github.com/modm-io/modm/blob/62410e51c845053022b10cb3cbf2b7162aacd9fe/src/modm/platform/clock/stm32/rcc.hpp.in#L775) in the platform startup code. It can only be written once and requires a controller power cycle to be reset. It would be nice if users could set the mode matching their hardware.

The reference manual states no RAM should be written before setting the power source but the ST HAL code uses the stack  before configuring it. Is ST ignoring their own advice or am I missing something?

@salkinium Do you think an lbuild option could be the right way to select the mode? The possible options would depend on the target. I don't think any present option in modm requires that.

salkinium:
> Is ST ignoring their own advice or am I missing something?

I guess the risk is about the SRAM getting under-/overvoltage and loosing some of its data. It probably doesn't happen uniformly, so this would make for some great Heisenbugs tho…

> Do you think an lbuild option could be the right way to select the mode?

Perhaps a generic solution would be a `modm_weak void modm_initialize_platform(void)` that is called directly after `__modm_initialize_platform`? That could be overwritten by the user, or in our case by the BSP.

chris-durand:
It seems the wrong header is selected for the H7(A|B)*q variants. The CI fails for stm32h7a3zit6q with `PWR_CR3_SMPSEN` not being present but the symbol is defined inside the stm32h7a3xxq.h header file.

chris-durand:
> Perhaps a generic solution would be a `modm_weak void modm_initialize_platform(void)` that is called directly after `__modm_initialize_platform`? That could be overwritten by the user, or in our case by the BSP.

That is a lot cleaner than lbuild options :+1:

salkinium:
> but the symbol is defined inside the stm32h7a3xxq.h header file.

Ah, variant header files. I've fixed the same problem for L1 files, [adapt the code here](https://github.com/modm-io/modm/blob/develop/ext/st/module.lb#L38-L45).

chris-durand:
@salkinium Could you test whether [configuring the power source](https://github.com/modm-io/modm/blob/29110d2657eaff779cd468e34ac6e42acc9cc207/src/modm/board/nucleo_h723zg/startup.cpp) lets you run the H743 at 480 MHz? It made a difference for me on the H723, without it I also couldn't go beyond 400 MHz.

salkinium:
I added some docs to your changes and reworked the linkerscript templates to allow collapsing the heap sections to zero if the .data section spills beyond the end of SRAM1 into SRAM2 etc. It also adds `.data_{ram}` and `.noinit_{ram}` sections to each memory, but I'll have to test more and write more docs.

> Could you test whether configuring the power source lets you run the H743 at 480 MHz?

Sure, but officially revY does not have the ODEN bit or VOS0 scale.

chris-durand:
> I added some docs to your changes and reworked the linkerscript templates to allow collapsing the heap sections to zero if the .data section spills beyond the end of SRAM1 into SRAM2 etc. It also adds .data_{ram} and .noinit_{ram} sections to each memory, but I'll have to test more and write more docs.

Nice!

> Sure, but officially revY does not have the ODEN bit or VOS0 scale.

Oh, ok, I wasn't aware that the processor on the Nucleo is a revision Y.

chris-durand:
I have added some more device data fixes in modm-io/modm-devices#65.

salkinium:
OK, I'm doing spot checks for my linkerscript changes and then I'm going to merge this.

salkinium:
> I'm doing spot checks for my linkerscript changes

I accidentally rewrote some of this and made the heap section additions generic, so we don't need a new linkerscript template for the H7 and can reuse the `idtcm.ld.in` we also use for F7. The other scripts are also agnostic to the number and arrangement of SRAM. I'm very happy with these changes now.

salkinium:
I made the mistake of looking into the Cortex-M7 Technical Reference Manual to figure out how the memory mapping works and thus tuned the linkerscripts again. I also added extensive documentation for STM32 and SAM.

salkinium:
> Error generating documentation for device stm32h735zgt6: 'NoneType' object is not subscriptable

This again… LeSigh™

PR #654, opened 2021-07-07
mmc5603.hpp correction: swapDataFloat shouldn't call itself

### LINKED ISSUE #653: modm:driver:mmc5603 recursive function call
`		float inline
		swapDataFloat(uint8_t index) const
		{ return swapDataFloat(index) / 16384.f; }`

should be

`		float inline
		swapDataFloat(uint8_t index) const
		{ return swapData(index) / 16384.f; }`

FlashForger:


salkinium (approved):
Thanks!

PR #656, opened 2021-07-11
[animation] add missing dependency to :processing:timer

daniel-k:
I tried to add `<module>modm:ui:animation</module>` to my `project.xml` in order to use some animation code, but compiling failed with a missing defintion of `ShortTimestamp`. After some digging, I found it in `modm:processing:timer` and it looks like the animation module misses this as a dependency.

But maybe I just used it wrong? FWIW, I've added the dependency and the missing include and now I can just add `modm:ui:animation` and it works as expected.

### DIFF
--- a/src/modm/ui/animation/base.hpp
+++ b/src/modm/ui/animation/base.hpp
@@ -15,6 +15,7 @@
 #include <stdint.h>
 #include <modm/architecture/interface/clock.hpp>
 #include <modm/math/utils/arithmetic_traits.hpp>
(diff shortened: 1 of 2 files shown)


salkinium (approved):
Your reasoning is perfectly fine, I probably just missed this since all the examples included the timer themselves already. Good catch!

daniel-k:
:tada: 

BTW: fading an LED using pre-computed Gamma correction, hardware PWM and Protothreads feels soooo overengineered but is pure awesomeness :heart:

PR #657, opened 2021-07-11
[feature] Nanopb scons integration

lmoesch:
Here is the current draft of nanopb (protobuf) integration into modm. Currently there are some points that are still on my todo-list

- wait for new nanopb release tag that includes python path fix on windows
- rework description text for module and protofile path
- squash commits

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -185,7 +185,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Checkout code and update modm tools
         run: |
-          (git submodule sync && git submodule update --init --jobs 8) & pip3 install --upgrade --upgrade-strategy=eager modm & wait
+          (git submodule sync && git submodule update --init --jobs 8) & pip3 install --upgrade --upgrade-strategy=eager modm protobuf==3.20.1 grpcio-tools & wait
       - name: Examples STM32F4 Without Discovery Board
         if: always()
         run: |
--- a/examples/nucleo_f429zi/nanopb/main.cpp
+++ b/examples/nucleo_f429zi/nanopb/main.cpp
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2022, Lucas Moesch
+ * Copyright (c) 2022, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This file originated from the nanopb project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <modm/board.hpp>
+#include <pb_encode.h>
+#include <pb_decode.h>
+#include <simple.pb.hpp>
+#include <complex.pb.hpp>
+
+int main()
+{
+    Board::initialize();
+
+    /* This is the buffer where we will store our message. */
+    uint8_t buffer[128];
+    size_t message_length;
+    bool status;
+
+    {
+        /* Allocate space on the stack to store the message data.
+         *
+         * Nanopb generates simple struct definitions for all the messages.
+         * - check out the contents of simple.pb.h!
+         * It is a good idea to always initialize your structures
+         * so that you do not have garbage data from RAM in there.
+         */
+        SimpleMessage message = SimpleMessage_init_zero;
+
+        /* Create a stream that will write to our buffer. */
+        pb_ostream_t stream = pb_ostream_from_buffer(buffer, sizeof(buffer));
+
+        /* Fill in the lucky number */
+        message.lucky_number = 42;
+
+        /* Now we are ready to encode the message! */
+        status = pb_encode(&stream, SimpleMessage_fields, &message);
+        modm_assert(status, "pb.enc", "Encoding SimpleMessage failed!");
+        message_length = stream.bytes_written;
+        modm_assert(message_length, "pb.len", "Empty SimpleMessage buffer!");
+    }
+    {
+        /* Now we could transmit the message over network, store it in a file or
+         * wrap it to a pigeon's leg.
+         */
+
+        /* But because we are lazy, we will just decode it immediately. */
+
+        /* Allocate space for the decoded message. */
+        SimpleMessage message = SimpleMessage_init_zero;
+
+        /* Create a stream that reads from the buffer. */
+        pb_istream_t stream = pb_istream_from_buffer(buffer, message_length);
+
+        /* Now we are ready to decode the message. */
+        status = pb_decode(&stream, SimpleMessage_fields, &message);
+        modm_assert(status, "pb.dec", "Decoding SimpleMessage failed!");
+        modm_assert(message.lucky_number == 42, "lucky_number", "Incorrect SimpleMessage values!");
+    }
+
+    {
+        /* Allocate space on the stack to store the message data.
+         *
+         * Nanopb generates simple struct definitions for all the messages.
+         * - check out the contents of simple.pb.h!
+         * It is a good idea to always initialize your structures
+         * so that you do not have garbage data from RAM in there.
+         */
+        ComplexMessage message = ComplexMessage_init_zero;
+
+        /* Create a stream that will write to our buffer. */
+        pb_ostream_t stream = pb_ostream_from_buffer(buffer, sizeof(buffer));
+
+        /* Fill in the unlucky number */
+        std::strcpy(message.query, "Hello World");
+        message.unlucky_number = 13;
+        message.toggle = true;
+        message.value = 4.00012f;
+
+        /* Now we are ready to encode the message! */
+        status = pb_encode(&stream, ComplexMessage_fields, &message);
+        modm_assert(status, "pb.enc", "Encoding ComplexMessage failed!");
+        message_length = stream.bytes_written;
+        modm_assert(message_length, "pb.len", "Empty ComplexMessage buffer!");
+    }
+    {
+        /* Now we could transmit the message over network, store it in a file or
+         * wrap it to a pigeon's leg.
+         */
+
+        /* But because we are lazy, we will just decode it immediately. */
+
+        /* Allocate space for the decoded message. */
+        ComplexMessage message = ComplexMessage_init_zero;
+
+        /* Create a stream that reads from the buffer. */
+        pb_istream_t stream = pb_istream_from_buffer(buffer, message_length);
+
+        /* Now we are ready to decode the message. */
+        status = pb_decode(&stream, ComplexMessage_fields, &message);
+        modm_assert(status, "pb.dec", "Decoding ComplexMessage failed!");
+
+        modm_assert(!strcmp(message.query, "Hello World"), "query", "Incorrect ComplexMessage values!");
+        modm_assert(message.unlucky_number == 13, "unlucky_number", "Incorrect ComplexMessage values!");
+        modm_assert(message.toggle == true, "toggle", "Incorrect ComplexMessage values!");
+        modm_assert(int(message.value) == 4, "value", "Incorrect ComplexMessage values!");
+    }
+
+    while (true)
+    {
+        /* tumbleweed */
+    }
+    return 0;
+}
--- a/ext/nanopb/nanopb.lb
+++ b/ext/nanopb/nanopb.lb
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2021, Lucas Moesch
+# Copyright (c) 2022, Niklas Hauser
+#
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+import os
+import subprocess
+import tempfile
+
+from pathlib import Path
+
+def init(module):
+    module.name = ":nanopb"
+    module.description = """
+# Nanopb - Protocol Buffers for Embedded Systems
+
+Nanopb is a small code-size Protocol Buffers (protobuf) implementation in ansi
+C. It is especially suitable for use in microcontrollers, but fits any memory
+restricted system.
+
+See https://github.com/nanopb/nanopb.
+
+
+## Build System Integration
+
+You can optionally point the build system to multiple protofiles using a
+comma-separated list of paths in the `modm:nanopb:source` option.
+You can specify the output location using the `modm:nanopb:path` option.
+See the `modm:build` submodules for further details.
+
+!!! bug "Currently only with SCons support"
+    Only the `modm:build:scons` module currently supports this feature.
+"""
+
+def prepare(module, options):
+    module.add_set_option(
+        PathOption(name="sources", absolute=True, empty_ok=True,
+                   description="Comma-separated list of paths to protofiles."),
+        default="")
+    module.add_option(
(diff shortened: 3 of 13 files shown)


ext/nanopb/nanopb.lb
```diff
@@ -0,0 +1,74 @@
...
+    module.description = """
+# Nanopb - Protocol Buffers for Embedded Systems
+
+Nanopb is a small code-size Protocol Buffers implementation in ansi C. It is especially suitable 
+for use in microcontrollers, but fits any memory restricted system.
+"""
+
+def prepare(module, options):
+    module.add_option(
+    PathOption(name="protofile.path", empty_ok=False, absolute=True, description=""))
```
salkinium:
```suggestion
    PathOption(name="protofile.path", empty_ok=False, absolute=True, description="", default=""))
```
Otherwise lbuild requires an option input without default.

ext/nanopb/nanopb.lb
```diff
@@ -0,0 +1,74 @@
...
+
+    env.copy("nanopb/pb_common.c", dest="pb_common.c")
+    env.copy("nanopb/pb_common.h", dest="pb_common.h")
+    env.copy("nanopb/pb_decode.c", dest="pb_decode.c")
+    env.copy("nanopb/pb_decode.h", dest="pb_decode.h")
+    env.copy("nanopb/pb_encode.c", dest="pb_encode.c")
+    env.copy("nanopb/pb_encode.h", dest="pb_encode.h")
+    env.copy("nanopb/pb.h", dest="pb.h")
+
+    build_protofile(env)
```
salkinium: I'll move this wrapper into modm_tools and see if I can get it integrated into a build system.

lmoesch:
Hey everyone - I know it's been quite a long time 😅. Nanopb pushed a new release "recently" and since I'm developing with modm again I'd like to look again into this issue.

@salkinium Did you had some insight with the integration in modm_tools?

salkinium:
> Did you had some insight with the integration in modm_tools?

No, not yet. I just finished my master thesis, so a lot of todos accumulated over the last six months. I would recommend to just leave it as it is right now, and later integrate it into the build systems. It's all just Python anyways.

lmoesch:
> > Did you had some insight with the integration in modm_tools?
> 
> No, not yet. I just finished my master thesis, so a lot of todos accumulated over the last six months. I would recommend to just leave it as it is right now, and later integrate it into the build systems. It's all just Python anyways.

First things first: Congratulations on finishing your thesis! 

I nevertheless dug further into the inner workings of scons and modm and I think I can manage to integrate it atleast as part of the scons build. This would have mostly the benefit of having a ways of simply rebuild the protocol without rebuilding modm and it could automatically be updated on scons build. 

Let me hear your thoughts on the following setup:

- The nanopb modm-repo is responsible for providing the basic headers (maybe via modm-tools) and signals that nanopb is used.
- A nanopb compiler script is added to scons/site-tools which adds a new Builder for nanopb (e.g. simply scons nanopb) which could also be part of the general build (not that easy because of chain dependencys so some Emittermagic is required).

salkinium:
> Let me hear your thoughts on the following setup:

Yeah, that's pretty much how I would do it.

There is [some emitter magic in the XPCC generator](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/site_tools/xpcc_generator.py), which also generates C++ code from an abstract description via a Python script.
But I remember there's also a SCons builder already part of the nanopb project itself, perhaps copying that would be better?

lmoesch:
> > Let me hear your thoughts on the following setup:
> 
> Yeah, that's pretty much how I would do it.
> 
> There is [some emitter magic in the XPCC generator](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/site_tools/xpcc_generator.py), which also generates C++ code from an abstract description via a Python script. But I remember there's also a SCons builder already part of the nanopb project itself, perhaps copying that would be better?

Jeah I had a quick look into the builder in nanopb/tests which looks pretty sophisticated and I will try to use it for code generation.

salkinium:
Just hack it until it works somewhat, I honestly don't fully understand SCons either even though I wrote most of our setup…

lmoesch:
Sometimes I'd like to question my life choices when it's already 4am and you find out, that the problem was not your code, but a bug in nanopb for windows which they had fixed years ago, but still not put into their latest release version....

Nevertheless I made quite progress with the integration. I could not use their scons integration directly, but I cherry picked most of it's functions into an own Builder. I will update this PR when I have a working example.

ext/nanopb/nanopb.lb
lmoesch: I'm not sure if a default should be provided here. With a default path it it possible to run a nanopb build with no source file to build.

salkinium: It's just that if you don't default the option, lbuild will cannot proceed if the module is selected (see the doc build job in CI).
I would make the behavior so that when the path is empty, everything is still copied, but the `{{ proto_source }}` is just missing from the SConstruct, for the user to fill it in later.

tools/build_script_generator/scons/module.lb
```diff
@@ -220,6 +226,12 @@ def post_build(env):
             "tools": tools,
             "is_modm": repo == "modm",
         })
+
+        # Add protofiles
+        subs["build_proto"] = has_nanopb
+        if has_nanopb:
+            subs["proto_source"] = env.relcwdoutpath(env.get("::protofile.path"))
```
salkinium:
```suggestion
protofile = env.get("::protofile.path")
if protofile:
    subs["proto_source"] = env.relcwdoutpath(protofile)
```

Then in the template
```
%% if proto_source is defined
proto_sources = ["{{ proto_source }}"]
%% else
proto_sources = []
%% endif
```

ext/nanopb/nanopb.lb
```diff
@@ -0,0 +1,75 @@
...
+    env.copy("nanopb/pb_common.c", dest="pb_common.c")
+    env.copy("nanopb/pb_common.h", dest="pb_common.h")
+    env.copy("nanopb/pb_decode.c", dest="pb_decode.c")
+    env.copy("nanopb/pb_decode.h", dest="pb_decode.h")
+    env.copy("nanopb/pb_encode.c", dest="pb_encode.c")
+    env.copy("nanopb/pb_encode.h", dest="pb_encode.h")
+    env.copy("nanopb/pb.h", dest="pb.h")
+
+    env.outbasepath = "modm/scons/site_tools/nanopb_builder"
+    env.copy("nanopb/tests/site_scons/site_tools/nanopb.py", dest="nanopb.py")
```
salkinium: This should go into the `:build:scons` module

ext/nanopb/nanopb.lb
```diff
@@ -0,0 +1,41 @@
...
+
+from pathlib import Path
+
+def init(module):
+    module.name = ":nanopb"
+    module.description = """
+    # Nanopb - Protocol Buffers for Embedded Systems
+    Nanopb is a small code-size Protocol Buffers implementation in ansi C. It is especially suitable
+    for use in microcontrollers, but fits any memory restricted system.
+    """
```
salkinium: Untab due to this being a raw string (thus preserving the tab in `lbuild discover -n :nanopb`)

salkinium (commented):
This looks really good! Do you want to add a small example to have the CI test it?

lmoesch:
Finally I see no erros in the CI 😅 But I tryed another more sophisticated example and it fails due to wrong build order of files. I have to dig into it before squashing everything together

.github/workflows/linux.yml
```diff
@@ -312,6 +312,22 @@ jobs:
...
+        uses: actions/checkout@v2
+      - name: Checkout code and update modm tools
+        run: |
+          (git submodule sync && git submodule update --init --jobs 8) & pip3 install --upgrade --upgrade-strategy=eager modm & wait
+      - name: Compile protobuf example
+        if: always()
+        run: |
+          (cd examples && ../tools/scripts/examples_compile.py protobuf)
+
+
```
salkinium: Please add the example to the nucleo_f429zi folder and remove the extra job (it takes a few minutes to launch a new container for this, way overkill for one example).

ext/nanopb/nanopb.lb
```diff
@@ -0,0 +1,41 @@
...
+def init(module):
+    module.name = ":nanopb"
+    module.description = """
+# Nanopb - Protocol Buffers for Embedded Systems
+Nanopb is a small code-size Protocol Buffers implementation in ansi C. It is especially suitable
+for use in microcontrollers, but fits any memory restricted system.
+"""
+
+def prepare(module, options):
+    module.add_option(PathOption(name="protofile.path", empty_ok=True, absolute=True, description="Path to a protofile which is automatically added by lbuild build.", default=""))
```
salkinium:
```suggestion
    module.add_option(PathOption(name="protofile.path",
                                 description="Path to a protofile.", 
                                 empty_ok=True, absolute=True, default=""))
```

salkinium (commented):
With a working example, I can add support to the other build systems and test it against SCons.

lmoesch:
Okay ... I found a solution ... but ...

image

lmoesch:
Okay, I have no idea why the CI fails a this point ... I mean I move the example into the  nucleo_f429zi directory but why does this crash another example? Or am I reading the log wrong?

lmoesch:
@salkinium any idea why the stm32f4-examples 2 pieline fails? I know I put the nanopb example in there, but it looks like this would cause another example to crash?

lmoesch:
> @salkinium any idea why the stm32f4-examples 2 pieline fails? I know I put the nanopb example in there, but it looks like this would cause another example to crash?

Okay, nevermind. Nanopbs SCons integration is even more broken on Linux than on Windows ... I'm not exactly sure how their tests can be "passing" when they complete fail when checked out on a local machine ...

lmoesch:
> @salkinium any idea why the stm32f4-examples 2 pieline fails? I know I put the nanopb example in there, but it looks like this would cause another example to crash?

Alright. I've been running the CI code in a local docker a couple of time now and AFAICT there is some kind of race-condition, because everytime I do a clean checkout I get different errors, e.g:

> scons: building terminated because of errors.arm-none-eabi-g++: error: modm/src/modm/math/utils/bit_operation.cpp: No such file or directory

or

> scons: building terminated because of errors.arm-none-eabi-g++: error: modm/src/modm/platform/clock/systick_timer.cpp: No such file or directory


By running 

> ../tools/scripts/examples_compile.py nucleo_f429zi/nanopb

When running the command often enough it builds after 2-3 times. And it's always other files that are "missing".

When running Scons directly in the directoy it fails compiling the ext/gcc repo due to missing files, which "fixes" itself by running the command three times.

No idea what is going on there. The dependency injection from nanopb just adds the generated *.cpp files to the sources and flags them dependant on the nanopb header (Which is the dirty fix because normally this would be part of the compile step). Funny thing is on my Windows hostmachine it builds without any issue 🙈

lmoesch:
Jeah I feel kinda genius for figuring out what the f**** is happening... nanopbs action builder uses chdir for running the command which is - NOT THREADSAFE. So all other commands that  run parallel are also moved to this directory. Thatswhy it cannot find the files anymore, because they are now relative to the protocol directory .... 

I thought it was a good Idea to import code from their scons directly but apparently I end up rewriting every function now ...

lmoesch:
🍾🍾🍾

Finally ... I do some cleanup tomorrow and will then squash everything together, what a mess 😂😂😂

examples/nucleo_f429zi/nanopb/main.cpp
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) 2022, Petteri Aimonen <jpa at nanopb.mail.kapsi.fi>, Lucas Moesch
+ */
```
salkinium, line 12: MPLv2?

examples/nucleo_f429zi/nanopb/main.cpp
```diff
@@ -0,0 +1,67 @@
...
+        message.lucky_number = 13;
+
+        /* Now we are ready to encode the message! */
+        status = pb_encode(&stream, SimpleMessage_fields, &message);
+        message_length = stream.bytes_written;
+
+        /* Then just check for any errors.. */
+        if (!status)
+        {
+            return 1;
```
salkinium: Ok, but FYI this will reboot the device eventually.

tools/build_script_generator/scons/module.lb
```diff
@@ -108,6 +108,9 @@ def build(env):
     elif device["core"].startswith("avr"):
         tools.update({"size", "avrdude"})
 
+    if env.has_module(":nanopb"):
+        tools.add("nanopb")
```
salkinium, line 113: Documentation about what this does in `module.md`?

ext/nanopb/nanopb.lb
```diff
@@ -0,0 +1,41 @@
...
+
+from pathlib import Path
+
+def init(module):
+    module.name = ":nanopb"
+    module.description = """
+# Nanopb - Protocol Buffers for Embedded Systems
+Nanopb is a small code-size Protocol Buffers implementation in ansi C. It is especially suitable
+for use in microcontrollers, but fits any memory restricted system.
+"""
```
salkinium, line 40: Perhaps add some text about how to use the `protofile.path` option?

examples/nucleo_f429zi/nanopb/main.cpp
lmoesch: Right, I just copy-pasted the nanopb example, hence the different copyright. I will adjust it so it runs on the uC.

examples/nucleo_f429zi/nanopb/main.cpp
lmoesch: zlib License stated in the repo - never heard of, but stated as campatible with GPL

ext/nanopb/nanopb.lb
lmoesch: Is there an example on how to describe lib options in the string?

salkinium:
I've added some changes to the docs and the SCons wrapper, mostly removing the dependency injection, which I think is done better this way.
I also added a output path to the generator to put it in line with the other generators.

Is there only one protofile to include, or do you expect the PathOption to take multiple paths? Cos that would require a `module.add_set_option(PathOption)` to enable multiple comma-separated paths in the `projects.xml`. If you want to support that, the example should be augmented.

lmoesch:
> I've added some changes to the docs and the SCons wrapper, mostly removing the dependency injection, which I think is done better this way. I also added a output path to the generator to put it in line with the other generators.
> 
> Is there only one protofile to include, or do you expect the PathOption to take multiple paths? Cos that would require a `module.add_set_option(PathOption)` to enable multiple comma-separated paths in the `projects.xml`. If you want to support that, the example should be augmented.

Heys! I would prefer the option to add multiple paths, but I did not know how to do it :'D I will add this then ;)

salkinium:
I added the set option and added a new protofile to the example, which needs to be adapted further.
There's still an issue with the `cmdstr` of the CommandAction, which I'm trying to fix.

salkinium:
> There's still an issue with the cmdstr of the CommandAction, which I'm trying to fix.

Fixed now, was very simple with `env.subst`.

salkinium:
I've extended the example, but I couldn't get the separate `.options` file to work, so the options need to be specified inline (which I find better anyways).
As far as I could tell the protoc compiler just ignores any `.options` file and I couldn't find any tests in nanopb that uses the options non-inline. So I guess their docs are out of date or something like that.

I would consider this PR now complete, can you take a look if something is missing?

salkinium (approved):

lmoesch:
Hey @salkinium Great job with reworking my nanopb-ripoff code :) In my eyes this PR is better then ever and has everything in it that I still had in mind (especially the linked protofiles).

PR #664, opened 2021-07-29
[driver] Ili9341 support Resumable Functions

TomSaw:
To sweep the yard for #665 , here's an _as good as possible_ pealed out update on ILI9341.
**Let's concentrate on the implementation of the Resumables for now.**
Drawing functions and a lot more will be cured with the coming up graphics API rewrite.

## Changes
- Resumable Functions. I could not backport RF for all drawing functions cause was too painfull.
- optimized drawing functions for Ili9341. The clipping-feature is your friend for squary shapes!
- `clear()` now also takes a color to clear with: `clear(Rgb565 color)`
- Dropped `backgroundColor`. Renamed `foregroundColor` -> `color`. To fullfill this change, clear(Rgb565 color) needs to become sooper fast and resource unhungry. A concept witch gets final grip with the coming up graphics API rewrite ;)
- `set(Toggle toggle, bool value)` substitutes lots of `xyzOn()`, `xyzOff()` methods.
- `set(ReadWrite reg, uint8_t value)` and `get(ReadWrite reg, uint8_t value)` substitute lots of `setXyz()`, `getXyz()` methods.
- Simplified Ili93441<->Interface implementation. I think `modm::Ili9341Spi<Ili9341InterfaceSpi<Spi, Cs, Dc> Reset>` reads and explains itself so much better than "simplified" `modm::Ili9341Spi<SpiMaster, Cs, Dc, Reset>`.
- Removed `<..., typename Backlight, ...>` from Driver
   - A simple `Backlight::setOutput(modm::Gpio::High)` subsitutes _the Feature_. However, sometimes a developer wants more control f.e. backlight PWM-Dimmed or hardwired to VCC so is better to keep the LED out of Ili9341 and not bloat the drivers template arguments.
- Optimized color::Rgb565 for performance reasons.
- Renamed color::BrightnessT -> GrayscaleT (Should be another PR 😬 or is ok ?)

# ToDo
- [ ] Update examples


salkinium:
Can this be closed in favor of #665?

TomSaw:
Jupp.

PR #667, opened 2021-08-05
[ext] Embedded Template Library ETL

rleh:
- [x] Submodule of partial [ETLCPP/etl](https://github.com/ETLCPP/etl) repo: https://github.com/modm-ext/etl-partial
- [x] lbuild module
- [x] Minimal example
- [x] Ignore `profiles/` folder and `doxygen.h`
- [x] `ETL_ASSERT` etc should be rerouted to `modm_assert`. I think replacing the `error_handler.h` file should be enough for that.
- [x] Verify that all compatible `ETL_*` defined are defaulted, like `ETL_CPP17_SUPPORTED` etc. 
- [x] `<etl_local.h>` mechanism? => ETL uses `etl_profile.h` themselves.
- [x] Add ETL as external project to Readme

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,6 @@
 [submodule "ext/lvgl/lvgl"]
 	path = ext/lvgl/lvgl
 	url = https://github.com/modm-ext/lvgl-partial.git
+[submodule "ext/etlcpp/etl"]
+	path = ext/etlcpp/etl
+	url = https://github.com/modm-ext/etl-partial.git
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ git clone --recurse-submodules --jobs 8 https://github.com/modm-io/modm.git
 - Integration of useful third-party software:
 	- [FreeRTOS][] and [FreeRTOS+TCP][].
 	- [CMSIS][] and [CMSIS-DSP][].
+	- [ETL][].
 	- [TinyUSB][].
 	- [FatFS][].
 	- [ROSserial][].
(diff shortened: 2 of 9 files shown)


salkinium:
Hm, ETL is certainly very large, I was wondering if there's any way to modularize ETL just a little. Something along the lines of the documentation grouping: https://www.etlcpp.com/documentation.html That would allow us to use ETL inside modm without forcing a lot of file to be generated all the time. We don't have to do it right away, only when we want to start using ETL inside modm.

TomSaw:
Waaa!! You're reading my thoughts. I've just scratched my head on how to get a python-like dict in C++ and here it is!

kikass13:
I have also added etl to our modm repositories and we are keen on using it excessively :+1: 

don't forget to add the etl::string to your MODM_LOGGER class, so that we can print etl::strings as well! :) 
I am using something like this for now
```
etl::string<255> etlStr{"xaxa"};
std::string str(etlStr.data(), etlStr.data_end());
MODM_LOG_INFO << str.c_str() << modm::endl;
```

which is awkward but fine, i guess :p

salkinium:
@rleh is it possible to address my todo list above until the Q3 release in 10 days? I don't think we can modularize the library since they update their code so much…

rleh:
Hopefully I'll find some time next week...

salkinium:
We should probably have a better example.

salkinium:
@rleh please review and test.

rleh (commented):
Approve ✅

salkinium (approved):

PR #669, opened 2021-08-20
Initial support for Smart Response XE

twasilczyk:
This change requires merging https://github.com/modm-io/modm-devices/pull/68 first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,9 @@ git clone --recurse-submodules --jobs 8 https://github.com/modm-io/modm.git
 
 ## Microcontrollers
 
-modm can create a HAL for <!--allcount-->3138<!--/allcount--> devices of these vendors:
+modm can create a HAL for <!--allcount-->3166<!--/allcount--> devices of these vendors:
 
-- STMicroelectronics STM32: <!--stmcount-->2593<!--/stmcount--> devices.
+- STMicroelectronics STM32: <!--stmcount-->2621<!--/stmcount--> devices.
 - Microchip SAM: <!--samcount-->163<!--/samcount--> devices.
 - Microchip AVR: <!--avrcount-->382<!--/avrcount--> devices.
 
--- a/README.md
+++ b/README.md
@@ -74,11 +74,11 @@ git clone --recurse-submodules --jobs 8 https://github.com/modm-io/modm.git
 
 ## Microcontrollers
 
-modm can create a HAL for <!--allcount-->3166<!--/allcount--> devices of these vendors:
+modm can create a HAL for <!--allcount-->3172<!--/allcount--> devices of these vendors:
 
 - STMicroelectronics STM32: <!--stmcount-->2621<!--/stmcount--> devices.
 - Microchip SAM: <!--samcount-->163<!--/samcount--> devices.
-- Microchip AVR: <!--avrcount-->382<!--/avrcount--> devices.
+- Microchip AVR: <!--avrcount-->388<!--/avrcount--> devices.
 
 Here is a table with all device families and the peripheral drivers they support:
 
--- a/README.md
+++ b/README.md
@@ -525,6 +525,7 @@ We have out-of-box support for many development boards including documentation.
 </tr><tr>
 <td align="center"><a href="https://modm.io/reference/module/modm-board-stm32_f4ve">STM32-F4VE</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-board-stm32f030_demo">STM32F030-DEMO</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-board-srxe">Smart Response XE</a></td>
 </tr>
 </table>
 <!--/bsptable-->
--- a/examples/srxe/blink/main.cpp
+++ b/examples/srxe/blink/main.cpp
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2021, Tomasz Wasilczyk
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+
+int
+main()
+{
+	Board::initialize();
+
+	while (true)
+	{
+		Board::LedDebug::toggle();
+		modm::delay(1s);
+	}
+}
--- a/src/modm/board/srxe/board.hpp
+++ b/src/modm/board/srxe/board.hpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2021, Tomasz Wasilczyk
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include <modm/platform.hpp>
+
+using namespace modm::platform;
+
+/// @ingroup modm_board_srxe
+namespace Board {
+
+using namespace modm::literals;
+
+using SystemClock = modm::platform::SystemClock;
+
+using LedDebug = GpioB0;
+using Leds = SoftwareGpioPort<LedDebug>;
+
+namespace Display {
+
+using DC = GpioD6;
+using CS = GpioE7;
+using RST = GpioG2;
+
+}  // namespace Display
+
+inline void
+initialize() {
+	SystemClock::enable();
+
+	LedDebug::setOutput();
+
+	enableInterrupts();
+}
+
+}  // namespace Board
--- a/src/modm/board/srxe/board.xml
+++ b/src/modm/board/srxe/board.xml
@@ -1,0 +1,15 @@
+<library>
+  <repositories>
+    <repository>
+      <path>../../../../repo.lb</path>
+    </repository>
+  </repositories>
+
+  <options>
+    <option name="modm:target">atmega128rfa1-zu</option>
+    <option name="modm:platform:core:f_cpu">16000000</option>
+  </options>
+  <modules>
+    <module>modm:board:srxe</module>
+  </modules>
+</library>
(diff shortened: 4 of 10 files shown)


salkinium:
You can actually update the modm-devices submodule to your fork's commit (use git add remote to add your fork, then choose the commit) without merging and the CI will work fine. That way you can work on both PRs simultaneously and we can merge them together. I would also prefer to have the BSP and an example in this PR so that the CI can perhaps catch additional issues.

twasilczyk:
Indeed it works pretty well. However the modm-devices pointer in ext/ hasn't been updated in a while so my commit pulls in more changes than just for ATmega128RFA1.

I will add some initial SRXE BSP and example (blinker and maybe uart), but the final implementation will need some separate PRs (e.g. I need to write LCD display driver etc etc).

README.md
```diff
@@ -312,7 +312,7 @@ Please [discover modm's peripheral drivers for your specific device][discover].
 <td align="center">○</td>
 <td align="center">○</td>
 <td align="center">✗</td>
-<td align="center">✗</td>
+<td align="center">○</td>
```
twasilczyk: This pops in, because the RF chips have the following xml tag:
```
<module name="FLASH">
  <instance name="FLASH" caption="FLASH Controller">
    <register-group name="FLASH" name-in-module="FLASH" offset="0x00" address-space="data" caption="FLASH Controller"/>
  </instance>
</module>
```

I don't see anything useful here and I guess this wasn't the intent of this field. Should I remove these lines via patch?

salkinium: Yes, but you should remove it in the Python code so it's universal: https://github.com/modm-io/modm-devices/blob/develop/tools/generator/dfg/avr/avr_device_tree.py#L74-L88

README.md
```diff
@@ -366,7 +366,7 @@ Please [discover modm's peripheral drivers for your specific device][discover].
 <td align="center">✅</td>
 <td align="center">✅</td>
 <td align="center">✗</td>
-<td align="center">✗</td>
+<td align="center">○</td>
```
twasilczyk: Huh. This is actually an obscure bug that took me a while to dig into. Not related to my change.

TL;DR: due to a bug in parsing stm32 clock module, all devices with flash module will have this one selected too. I'll file a separate bug for that.

twasilczyk: Filled as https://github.com/modm-io/modm/issues/671

Since I fixed the flash tick, this problem is not applicable to my PR anymore.

README.md
twasilczyk: Fixed.

twasilczyk:
Okay, I think I fixed all the immediate problems. Please let me know if blinky example is good enough for initial PR or it should be more complete with UART and logger.

src/modm/board/srxe/board.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include <modm/platform.hpp>
+
```
salkinium, line 15: Our BSPs have [a top-level `using namespace modm::platform;`](https://github.com/modm-io/modm/blob/develop/src/modm/board/disco_f072rb/board.hpp#L22) to avoid all the typing.

examples/srxe/blink/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+
+int
+main()
+{
+	using namespace Board;
+
+	initialize();
```
salkinium:
```suggestion
	Board::initialize();
```

examples/srxe/blink/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+
+int
+main()
+{
+	using namespace Board;
```
salkinium:
```suggestion
```

salkinium (changes requested):
Otherwise excellent!

src/modm/board/srxe/board.hpp
twasilczyk: Ouch, this is something I wanted to start a discussion later, but since you called it out...

Your own [coding conventions](https://github.com/modm-io/modm/blob/develop/docs/coding_convention.md#statements) disallow this :). This is not just to avoid typing or not - using namespaces in header files pollute the namespaces where it's used in. Think of this as an API and imagine how would it look when you generated documentation for all the symbols in your namespace tree:

```
::modm::Platform::GpioB0
::modm::Platform::GpioB1
::modm::Platform::GpioB2
::GpioB0
::GpioB1
::GpioB2
::Board::GpioB0
::Board::GpioB1
::Board::GpioB2
::Board::Led = GpioB1
```

You generally have three types of this kind of violation:
- using platform namespace in Board namespace - doesn't pollute global namespace, but duplicates platform namespace onto board. Unless this is exactly what you want :)
- using platform namespace in global namespace of board.hpp - pollutes global namespace for all users of board.hpp
- using board namespace (I've seen it at least in arduino board definition) - like the one above, but this also kind of counters the point of having board namespace in the first place

I would generally strongly advise for cleaning `using namespace` usage through modm. One exception you might consider could be literals inside board namespace (I don't have strong feelings about this one). Yes, this forces you to type a little bit more in the header files. But the resulting API is clean. And the libraries are about APIs, not how many LOC they use. Check how stdlib++ looks like in this regard :).

At the end of the day it's your project, so I'll comply with whatever you say. Just please reconsider this one :).

salkinium: I agree with your assessment, however, for consistency we should do this in this board too. It's a bit annoying to refactor all our examples again.

> but this also kind of counters the point of having board namespace in the first place

The idea was that user code can extend a namespace with additional definitions, which isn't possible with a struct. But it's rarely used, most people define their own completely new BSP.

> I would generally strongly advise for cleaning using namespace usage through modm.

Yes, I'm aware of this issue. I've wrapped our [`using namespace modm::literals` into a macro](https://github.com/modm-io/modm/blob/develop/src/modm/board/board.hpp#L14-L17) so if your project defines it, it won't be included.

Perhaps we can rename it to `MODM_DISABLE_GLOBAL_NAMESPACES` or something like that (and actually document it \*cough\*) and add the `using namespace modm::platform` to this file and remove it from the individual `board.hpp` files. That way our examples won't need to be touched (again) and you can still disable this behaviour for your project? (I think this may require adding `modm::platform` to all definitions in every board.hpp files though…)

twasilczyk: Yeah, an option to disable global namespaces sounds like a solution indeed. Do you think it would be possible to enable it somehow globally for a board (is it possible to set a default for a define via module.lb? Somewhere around env.collect)? It would be great to have it as a default at some point.

Making such a change (even in a limited form like doing `using namespace Board` in main.cpp) in all examples is a non-negligible effort, but modifying 260 local files isn't as nearly as problematic as pushing such update to 20 external projects depending on your library :). It's much easier to clean up APIs before the industry starts using it. And I really feel this project is showing direction in which the whole embedded industry should head (if only they weren't so settled with C).

Having said that, I've updated this PR according to suggestions.

src/modm/board/srxe/board.xml
```diff
@@ -0,0 +1,15 @@
...
+      <path>../../../../repo.lb</path>
+    </repository>
+  </repositories>
+
+  <options>
+    <option name="modm:target">atmega128rfa1-zu</option>
+    <option name="modm:platform:core:f_cpu">16000000</option>
+  </options>
+  <modules>
+    <module>modm:board:srxe</module>
```
twasilczyk, line 13: One extra question - is the `srxe` short name right? Besides Smart Response XE, there are two more devices having some attention from tinkering community:
- Smart Response PE - unlikely to be supported by modm, because it's based on TI's CC2430 and being much less appealing HW (numerical keyboard, simpler screen)
- Smart Response XE Receiver - very similar architecturally to Smart Response XE, but doesn't have a screen and does have a USB connector

I might add a board for the latter at some point. It could be named `srxe-recv` to complement `srxe`.

src/modm/board/srxe/board.hpp
salkinium: > Do you think it would be possible to enable it somehow globally for a board

Yes, but then it should be a "positive flag" in the board's `module.lb`:

```python
env.collect(":build:cppdefines", "MODM_ENABLE_GLOBAL_NAMESPACES=0")
```

Then in the `project.xml` you can overwrite it (modm will place it last on the command line):

```xml
<library>
  <collectors>
    <!-- Enable using global namespaces -->
    <collect name="modm:build:cppdefines">MODM_ENABLE_GLOBAL_NAMESPACES=1</collect>
  </collectors>
</library>
```

The macro would then be explicitly defaulted:

```cpp
#ifndef MODM_ENABLE_GLOBAL_NAMESPACES
#define MODM_ENABLE_GLOBAL_NAMESPACES 1
#endif

#if MODM_ENABLE_GLOBAL_NAMESPACES
using namespace modm::platform;
using namespace modm::literals;
using namespace std::chrono_literals;
#endif
```

However, I would still like this to be consistent for all boards, otherwise I get confused and will fix it in a year when I've forgotten about this conversation.

> pushing such update to 20 external projects depending on your library :)

Meh, we can do breaking changes in modm, we've done so many times already, see our Changelog. Not sure if this is a good reason though, maybe devs will get annoyed.

> before the industry starts using it. 

I really hope they skip C++ and head right to Rust since C/C++ integer promotion rules alone drive me insane 🤪

But [the plan is to give modm-devices a proper Python API](https://github.com/modm-io/modm-devices/pull/35) and move all the STM32 specific queries from this repo into there with documentation and then let people use that to help generate their own project/company-specific HAL. That way different languages and HAL designs can still collaborate via modm-devices an ease the burden of data collection.

src/modm/board/srxe/board.xml
salkinium: I think it's fine, especially with the example showing how to use it and both being listing and linked on the homepage.

salkinium: The alternatives are `modm:board:smart-response-xe` and `modm:board:smart-response-xe-receiver`, but meh, I don't think it's a hugely important. We can always rename it later with a deprecation alias as done with the black-pill* boards.

salkinium (approved):
Excellent work!

Issue #670, opened 2021-08-20
Support grayscale displays


twasilczyk:
modm::ui::display either supports monochrome or color displays. The two displays I've been working with (SSD1322 and ST7586S) are 16- and 4-level grayscale.

One option would be to make ColorGraphicDisplay a template allowing other color profiles than Rgb565; then add 16/4 level color profile. Then maybe even MonochromeGraphicDisplay could just use the same template with 1-bit color profile. There's a challenge though: in 16-level grayscale displays, two pixels are encoded in a single byte.

While we're at modm::ui::display, there's also one benefit of using these controllers that modm doesn't utilize: you can select a fragment of the display and not update the whole thing.

salkinium:
@TomSaw is our display expert now, his https://github.com/modm-io/modm/pull/665 is addressing some of these issues.

TomSaw:
Yepp I've also had my concerns with the display implementations in the beginning of this year.
The #665 rewrite considers any type of color including grayscale of any depth.

However, not all cases **where multiple pixels are stuffed into one byte** are covered yet:

- `modm::color::Grayscale<>` needs implementation for n-bit grayscale-depths to cover f.e. 4 and 16 level Grayscale.

- `modm::graphic::Buffer` needs according implementation.

I've added this to the TODOs of #665 and also should have some SSD1322 around so I can test it. Being on a journey until next week but I'll implement this as next. Stay tuned!

twasilczyk:
This is awesome and I'm glad to hear you're considering multi-pixel-per-byte cases. There's one extra complication with SSD1322 that you need to be aware of: addressing is per word, not per byte. So a single 16-bit cell in memory represents four 4-bit pixels - this affects the widths of partial drawing sprites.

Did you consider extending support to drawing just parts of the display? This speeds up most of the use cases *a lot*. In my own project I easily hit north of 100 FPS (that's where my counter ends) even with a very inefficient SPI driver. No wonder why - each frame was sending just a handful of bytes. Added benefit is that you don't need much RAM - you can drive a large OLED with something as small as atmega8.

TomSaw:
`a single 16-bit cell in memory represents four 4-bit pixels - this affects the widths of partial drawing sprites.`
Ok, good to know! This should only affect the transfer-algorithm of the display-driver. I'll keep this in mind when debugging ;).

`Did you consider extending support to drawing just parts of the display`
[...]

PR #665, opened 2021-07-29
[graphic] Rewrite graphics - Round 1

TomSaw:
# Synopsis of added goods
- **Generic colors**
  - Feature complete
    - All operations with **saturation arithmetics**
    - **Conversion** between any instances of **Hsv <-> Rgb** 
  - **Bit granularity** for `modm::Gray`. Same for `modm::Rgb` and `modm::Hsv` cause they're made form 3x `modm::Gray`
    - Symetric scaling. Example convertions 2-bit Gray ---> 4-bit Gray.
        - 0b00 --> 0b0000
        - 0b01 --> 0b0101 // No stupid lshift, the gap is filled up
        - 0b10 --> 0b1010
        - 0b11 --> 0b1111 // Saturated stays saturated, white stays white
  - **C++20 concepts** for colortypes and colorgroups (currently _ColorPlane_ and _ColorPalletizing_). These concepts are very handy to modelate overload resolutions on specialision of buffers and their methods. They're also the key for good readability.
- **Generic graphic buffers**
  - **Arbitrary Color** and **Resolution**
  - Full support for `Buffer` **interoperation**: Copy/Convert buffers from sources (Flash, Ram, _SD-Card not yet implemented_) to destinations (Ram, Display).
  - **Pixel palletizing** for 1-, 2- and 4-bit grayscale (1-bit == Monochrome)
  - **Indexed colors** for compressed, canned Images but also for resource-friendly drawing and texting.
- **Buffer manipulations**
  - **Optimized** drawing of primitives (**Point, Line, Circle, Ellipse**)
  - <s>`foregroundColor`</s>, <s>`backgroundColor`</s> -> replaced by a **colormap***
  - **DMA** whenever possible: each of `clear()` and `Buffer` copies without color-conversion / -mapping
  - Generic **fill patterns** (gradients, checkerboard, stripes). See examples
- **Support external hardware**
  - All features also work efficiently on **external hardware** using **resumable functions**: Currently implemented for ili9341
  - ili9341::`draw(Point / HLine / VLine / Rectangle, Section)` is hyper-fast cause no pixel-pushing is involved. Just some register setup and DMA, filling the area. See #666 . This is great for Prototyping and Tools: You can f.e. draw/update colored Bars with next to zero resource-footprint.

## Roadmap
- [ ] Merge #781
- [ ] Merge #782
- [ ] Refactor graphic directory structure
- [ ] New generic `graphic::Buffer` including Display driver write-methods
[...]

PR #673, opened 2021-08-30
Initial implementation of Smart Response XE display

twasilczyk:
Implementation is monochrome only for now (the display is 4-level grayscale) until the new display APIs get merged from TomSaw's branch.

### DIFF
--- a/src/modm/platform/spi/at90_tiny_mega/spi_master.hpp.in
+++ b/src/modm/platform/spi/at90_tiny_mega/spi_master.hpp.in
@@ -61,9 +61,9 @@ public:
 	DataMode : uint8_t
 	{
 		Mode0 = 0,	///< clock normal,   sample on rising  edge
-		Mode1 = (1 << CPOL{{ id }}),	///< clock normal,   sample on falling edge
-		Mode2 = (1 << CPHA{{ id }}),	///< clock inverted, sample on rising  edge
-		Mode3 = (1 << CPOL{{ id }}) | (1 << CPHA{{ id }}),	///< clock inverted, sample on falling edge
+		Mode1 = (1 << CPHA{{ id }}),	///< clock normal,   sample on falling edge
+		Mode2 = (1 << CPOL{{ id }}),	///< clock inverted, sample on falling  edge
+		Mode3 = (1 << CPOL{{ id }}) | (1 << CPHA{{ id }}),	///< clock inverted, sample on rising edge
 	};
 
 public:


src/modm/ui/display/monochrome_graphic_display_horizontal_impl.hpp
```diff
@@ -18,22 +18,22 @@ template<int16_t Width, int16_t Height>
 void
 MonochromeGraphicDisplayHorizontal<Width, Height>::setPixel(int16_t x, int16_t y)
 {
-	if ((x < Width) and (y < Height)) { buffer[y][x / 8] |= (1 << (x % 8)); }
+	if ((x < Width) and (y < Height)) { this->buffer[y][x / 8] |= (1 << (x % 8)); }
```
twasilczyk: Seems that no test/example code uses this class. Adding `this->` fixes the build, but I didn't dig to why didn't it work in the first place.

examples/srxe/display/main.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <modm/board.hpp>
+
+int
+main()
+{
+	Board::initialize();
+
+	auto& display = Board::display;
+	display.drawRectangle(0, 0, 10, 10);
```
twasilczyk: This is just testing code. I will make it more meaningful in the final version of this PR.

src/modm/board/srxe/board.hpp
```diff
@@ -25,21 +27,40 @@ using SystemClock = modm::platform::SystemClock;
...
 
 inline void
 initialize() {
 	SystemClock::enable();
 
 	LedDebug::setOutput();
 
+	spi::SpiMaster::connect<spi::Sck::Sck, spi::Mosi::Mosi, spi::Miso::Miso>();
+	spi::SpiMaster::initialize<SystemClock, 1_MHz>();
+	// clock is high when inactive, data is sampled on clock trailing edge
```
twasilczyk: Something's off either in documentation of DataMode or the description of Mode 3 I found on the Internet. I'll file a separate bug about this.

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,119 @@
...
+{
+	sendCommand(Command::WriteDisplayData);
+	CS::reset();
+	// TODO: support windows other than full screen
+	// TODO: transfer the whole memory area, not individual pixels
+	for (uint16_t y = 0; y < Height; y++) {
+		for (uint16_t x = 0; x + 2 < Width; x += 3) {
+			uint8_t cell = 0;
+
+			// TODO: getting pixels this way is not very speedy
```
twasilczyk: @TomSaw look at this encoding. 4-level gray scale, but encoded at 3 pixels per byte.

Check page 18 of datasheet: https://www.topwaydisplay.com/sites/default/files/2020-04/ST7586S.pdf

twasilczyk:
Okay, I think this is ready to go. I'm pretty happy with this code and with how your library supports good code.

Honorable mentions:
- all commands+data is sent to the display via full structs and not a bunch of uint8_t's
- no magic numbers, entire protocol is pretty self-explanatory (see initialization code)
- `LittleEndian<>` and `BigEndian<>` templates, I always wanted to code them. See `SetVop` on how I used it.

Challenges:
- it's not easy to run unit tests on Ubuntu (I have 20.04.2 LTS) which doesn't ship with gcc 10
- avr-gcc shipped with modm doesn't support `std::is_constant_evaluated` (that's C++20 symbol)
- why SPI::transferBlocking requires uint8_t* and doesn't just accept void* pointers? See `sendCommand`
- is register type explosion expected? I got `NLineInversion`, `NLineInversion_t`, `LineInversionType` and `LineInversionType_t` for a single register
- I'm still not sure why `MonochromeGraphicDisplayHorizontal` didn't work. I didn't dig deep though.

src/modm/driver/display/st7586s_impl.hpp
TomSaw: Is included in #665 . You can choose any of 1 - 64 bpp for your Grayscale-type

src/modm/ui/display/monochrome_graphic_display_horizontal_impl.hpp
TomSaw: My failure on missing `this->` whilst refactor. You're the first applying the horizontal monochrome since then.

src/modm/driver/display/st7586s_impl.hpp
TomSaw: Means more options on instantiating the `Buffer<...>`. Do you know other Displays out there with such a wild mapping @twasilczyk  ? If it's an exception, i prefer to add something like a `BufferST7586<Resolution>` instead of making the generic `graphic::Buffer<...>` capable to cover such cases so the code keeps clean.

src/modm/driver/display/st7586s_protocol.hpp
```diff
@@ -0,0 +1,187 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+#ifndef MODM_ST7586S_HPP
+#error "Do not include this file directly. Include st7586s.hpp instead."
+#endif
```
TomSaw: #pragma once AND include guards?

src/modm/driver/display/st7586s_protocol.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+	constexpr SetColumnRow(uint8_t start, uint8_t end) : start(start), end(end) {}
+};
+
+enum class DisplayControl : uint8_t {
+	ComInc = 0,
+	ComDec = (1 << 7),
+	SegInc = 0,
+	SegDec = (1 << 6),
+};
```
TomSaw: Optionally there's `#include <modm/math/utils/bit_constants.hpp>` but may be overkill.
`enum class DisplayControl : uint8_t {`
`ComInc = 0,`
`ComDec = Bit7,`
`SegInc = 0,`
`SegDec = Bit6,`
`};`

src/modm/driver/display/st7586s_protocol.hpp
```diff
@@ -0,0 +1,187 @@
...
+enum class SetBooster : uint8_t {
+	x1 = 0b000,
+	x2 = 0b001,
+	x3 = 0b010,
+	x4 = 0b011,
+	x5 = 0b100,
+	x6 = 0b101,
+	x7 = 0b110,
+	x8 = 0b111,
+};
```
TomSaw: I think ints express this better and are easier to read:
`x1 = 0,`
`x2 = 1,`
`x3 = 2,`
`x4 = ...`

TomSaw (changes requested):
Just a quick look for now.
For consistence to `ssd1306_register.hpp`, `st7586s_protocol.hpp` should be `st7586s_register.hpp`.

src/modm/ui/display/monochrome_graphic_display_horizontal_impl.hpp
twasilczyk: Any quick explanation why was it needed in the first place?

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: Displays yes (it’s Smart Response XE and Lego Mindstorms EV3), but they are both using the same controller. I’m fine doing buffer implementation locally.

twasilczyk:
> Just a quick look for now.
> For consistence to `ssd1306_register.hpp`, `st7586s_protocol.hpp` should be `st7586s_register.hpp`.

I also had this thought, but these are not registers (but commands and their payloads) and definitely not defines. Thoughts?

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: Not precisely :). Ifndef protects from including this file before the main header, pragma protects from including it after.

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: It’s 1:1 transcribed from data sheet. But yeah, looks weird. I’ll change these to base 10.

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: Will do.

src/modm/driver/display/st7586s_protocol.hpp
TomSaw: These datasheet-writers do what they want.
When the valid values for an register are obvious an incrementing integral, i  also write in intergrals. This also demonstrates the reader, there's nothing special to care for. Lists of binaries instead grab your special attention.

src/modm/driver/display/st7586s_protocol.hpp
TomSaw: EDIT <strike>I'm a student (as well) - never have seen such a construct. Any guess @salkinium ?</strike>

TomSaw:
> > Just a quick look for now.
> > For consistence to `ssd1306_register.hpp`, `st7586s_protocol.hpp` should be `st7586s_register.hpp`.
> 
> I also had this thought, but these are not registers (but commands and their payloads) and definitely not defines. Thoughts?

Hardware-interfaces are likely a chaotic mix of registers, commands, protocols... more precise distinctions are taken up with the naming of enum's  ->  check other driver-headers!

I think it's an old convention to collect such things in one box called _register_ for the benefit of consistence.

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: I think primarily because modm keeps implementation in "cpp" header files included at the end of the actual header files and not just in source files. It's already an unique setup.

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: It's obvious when it's working :). I try to stick to datasheet as close as possible for starters until I see my code actually working with a given peripheral. Otherwise it may be a hell to debug and figure out I made a mistake converting binary to decimal.

As I mentioned before, I already changed this to decimal.

twasilczyk:
> > > Just a quick look for now.
> > > For consistence to `ssd1306_register.hpp`, `st7586s_protocol.hpp` should be `st7586s_register.hpp`.
> > 
> > 
> > I also had this thought, but these are not registers (but commands and their payloads) and definitely not defines. Thoughts?
> 
> Hardware-interfaces are likely a chaotic mix of registers, commands, protocols... more precise distinctions are taken up with the naming of enum's -> check other driver-headers!
> 
> I think it's an old convention to collect such things in one box called _register_ for the benefit of consistence.

modm doesn't seem to stick to a single name for this stuff. Interface definitions are scattered around files named: `_defines`, `_transport`, `_register`, `_definitions`, `_data`, `_constants`. `register` is only used by ssd1306 driver actually.

@salkinium, what's your take? I don't think this library should stick to old conventions, but that's not my call at the end of the day :)

src/modm/ui/display/monochrome_graphic_display_horizontal_impl.hpp
twasilczyk: I found some explanation myself:
- https://www.cs.technion.ac.il/users/yechiel/c++-faq/nondependent-name-lookup-members.html
- https://eli.thegreenplace.net/2012/02/06/dependent-name-lookup-for-c-templates

Interesting read!

src/modm/board/srxe/board.hpp
twasilczyk: Pushed fix in #674

TomSaw:
Have  proposed my best solution of grayscale-support for st7586 in #670

src/modm/driver/display/st7586s.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <modm/ui/display/monochrome_graphic_display_horizontal.hpp>
+
+#include "st7586s_protocol.hpp"
+
+namespace modm
+{
+
+template <typename SPI, typename CS, typename RST, typename DC, int Width = 384, int Height = 160>
+// TODO: this controller has pixels packed by 3, not 8 per byte
```
TomSaw: should be 4 not 8

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+	sendCommand(Command::DisplayDuty, uint8_t(Height - 1));
+	sendCommand(Command::InverseOff);
+	setWindow(0, 0, Width, Height);
+	update();  // clear the display
+	sendCommand(Command::DisplayOn);
+}
+
+template <typename SPI, typename CS, typename RST, typename DC, int Width, int Height>
+void
+St7586s<SPI, CS, RST, DC, Width, Height>::setWindow(uint16_t x, uint16_t y, uint16_t w, uint16_t h)
```
TomSaw, line 79: Same method is already called setClipping(...), let's follow the scheme

src/modm/driver/display/st7586s.hpp
```diff
@@ -0,0 +1,47 @@
...
+	void sendCommand(Command cmd, const void *data, size_t len);
+	template <typename Data>
+	void sendCommand(Command cmd, Data data) {
+		sendCommand(cmd, &data, sizeof(data));
+	}
+	void sendCommand(Command cmd) {
+		sendCommand(cmd, nullptr, 0);
+	}
+
+	void setWindow(uint16_t x, uint16_t y, uint16_t w, uint16_t h);
```
TomSaw: Same method is already named setClipping(...) in other display-drivers. let's follow the scheme

src/modm/driver/display/st7586s.hpp
twasilczyk: I mean MonochromeGraphicDisplay packs by 8. Is basically a todo to wait for your PR.

src/modm/driver/display/st7586s.hpp
twasilczyk: Sure. Ideally setClipping should be a part of the API, not just a naming scheme.

twasilczyk (commented):
> Have proposed my best solution of grayscale-support for st7586 in #670

I’ll debate a little on the details later, but agree overall. I’ll wait with grayscale mode until you merge your PR.

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+St7586s<SPI, CS, RST, DC, Width, Height>::setClipping(uint16_t x, uint16_t y, uint16_t w, uint16_t h)
+{
+	namespace payload = impl::st7586s::payload;
+
+	modm_assert(x < Width, "st4586s.sc.x", "x >= Width", x);
+	modm_assert(x + w <= Width, "st4586s.sc.xw", "x + w >= Width", x + w);
+	modm_assert(y < Height, "st4586s.sc.y", "y >= Height", y);
+	modm_assert(y + h <= Height, "st4586s.sc.yh", "y + h >= Height", y + h);
+	modm_assert((x % pixelsPerByte) == 0, "st4586s.sc.x%", "x is not a multiply of PBB", x);
+	modm_assert((w % pixelsPerByte) == 0, "st4586s.sc.w%", "w is not a multiply of PBB", w);
```
TomSaw: These asserts evaluate all-over-again runtime :/

-> Why do you need setClipping to be public? Witch case requires the app-writer to call this directly?

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+	}
+	CS::set();
+	// exit with data mode on
+}
+
+template <typename SPI, typename CS, typename RST, typename DC, int Width, int Height>
+void
+St7586s<SPI, CS, RST, DC, Width, Height>::initialize()
+{
+	namespace payload = impl::st7586s::payload;
```
TomSaw: You can place `using payload = impl::st586s::payload` within class declaration to spare repetitions.

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+	setClipping(0, 0, Width, Height);
+	update();  // clear the display
+	sendCommand(Command::DisplayOn);
+}
+
+template <typename SPI, typename CS, typename RST, typename DC, int Width, int Height>
+void
+St7586s<SPI, CS, RST, DC, Width, Height>::setClipping(uint16_t x, uint16_t y, uint16_t w, uint16_t h)
+{
+	namespace payload = impl::st7586s::payload;
```
TomSaw: Here's the repetition ;)

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+template <typename SPI, typename CS, typename RST, typename DC, int Width, int Height>
+void
+St7586s<SPI, CS, RST, DC, Width, Height>::initialize()
+{
+	namespace payload = impl::st7586s::payload;
+
+	// Configure GPIO
+	RST::reset();
+	RST::setOutput();
```
TomSaw: RST::setOutput(false); // same result

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+	CS::set();
+	CS::setOutput();
```
TomSaw: CS::setOutput(true); // same result

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+	DC::setOutput();
+
+	// Reset display
+	modm::delay_us(10);
```
TomSaw: modm::delay(10us); // modern literal syntax

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+	RST::set();
+	modm::delay_ms(120);
```
TomSaw: modm::delay(120ms); // modern literal syntax

src/modm/driver/display/st7586s_impl.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+	// Power ON operation flow (see datasheet)
+	sendCommand(Command::SleepOff);
+	sendCommand(Command::DisplayOff);
+	modm::delay_ms(50);  // t_{ON-V2}
```
TomSaw: ...

TomSaw (changes requested):
Apologizes for the chopped review. I'm fighting with my IDEs github extension, for now falling back to the browser. So here's part 3/3 :P

I leave the review of your endianness-edits @salkinium cause haven't enough background.

src/modm/driver/display/st7586s_protocol.hpp
TomSaw: > modm doesn't seem to stick to a single name for this stuff. Interface definitions are scattered around files named: `_defines`, `_transport`, `_register`, `_definitions`, `_data`, `_constants`. `register` is only used by ssd1306 driver actually.
> 
> @salkinium, what's your take? I don't think this library should stick to old conventions, but that's not my call at the end of the day :)

Confirm, we need a convention for future drivers so this unifies. I like `_defines` the most cause it takes the classic culture from ANSI-C. Anything else is already to tight.

TomSaw:
> modm doesn't seem to stick to a single name for this stuff. Interface definitions are scattered around files named: `_defines`, `_transport`, `_register`, `_definitions`, `_data`, `_constants`. `register` is only used by ssd1306 driver actually.
> 
> @salkinium, what's your take? I don't think this library should stick to old conventions, but that's not my call at the end of the day :)

There should be a standard!? I prefer `_defines` cause it takes the classic culture from ANSI-C. Anything else is already too tight.

src/modm/driver/display/st7586s_protocol.hpp
TomSaw: Most of the implementations are templates, therefore these `*_impl.hpp`. It's a `Header-only` concept but split into declaration and definition for readability.

You're right 😯 without `#pragma once` the headers, one can do stupid things like this:
`#include st7586s.hpp // defines MODM_ST7586S_HPP`
`#include st7586s_impl.hpp // redefinition errors without #pragma once`

If this is a general problem, `#pragma once` should go into every `*.hpp` containing `#error "Don't include this file directly, use '*.hpp' instead!"`

TomSaw (commented):
_

src/modm/math/utils/bit_operation.hpp
```diff
@@ -34,7 +34,7 @@ namespace modm
 	 * 0xab => 0xba
 	 * \endcode
 	 */
-	modm_always_inline uint8_t
+	constexpr modm_always_inline uint8_t
```
salkinium: Let's remove all the `modm_always_inline`, since constexpr already implies `inline` and inlining should be left up to the compiler.

src/modm/driver/display/st7586s_protocol.hpp
salkinium: It's mainly just historic convention to have just one header file which pulls in everything that is needed. We can remove this construct completely and instead just `#pragma once` and then `#include "st7586s.hpp"`, which will also make IDE code completion work in this file, since it's pulling in the correct definitions.

(Ok, this file is dependency free, so you don't need the `#include "st7586s.hpp"` part, but in the `st7586s.hpp_impl.hpp` file that can be used).

src/modm/driver/display/st7586s_impl.hpp
salkinium: Yeah, this is going to be slow. However, you can use `modm_assert_continue_fail_debug`, which will still evaluate at runtime, however only in debug profile. I think that's a good compromise to find nasty bugs.

src/modm/driver/display/st7586s_protocol.hpp
```diff
@@ -0,0 +1,187 @@
...
+// ----------------------------------------------------------------------------
+
+#pragma once
+#ifndef MODM_ST7586S_HPP
+#error "Do not include this file directly. Include st7586s.hpp instead."
+#endif
+
+#include <modm/math/utils/endianness.hpp>
+
+namespace modm::impl::st7586s
```
salkinium: Typically the naming convention is `modm::detail::st7586s` similarly as in the std lib.

salkinium:
> avr-gcc shipped with modm doesn't support std::is_constant_evaluated (that's C++20 symbol)

[GCC supports it since v9](https://gcc.gnu.org/projects/cxx-status.html), so I suspect maybe a problem with combining `constexpr` and `modm_always_inline`?

> why SPI::transferBlocking requires uint8_t* and doesn't just accept void* pointers? See sendCommand

`void*` just requires more casting for the average use-case, since you typically store buffers as `uint8_t*`.

> modm doesn't seem to stick to a single name for this stuff. Interface definitions are scattered around files named: _defines, _transport, _register, _definitions, _data, _constants. register is only used by ssd1306 driver actually.

There isn't really a convention from my side. We typically allow the wild west to sprout until it creates enough annoyance to warrant a alignment to a convention. So far the convention has been: There should be only one header file per driver (ideally identically named as the lbuild module) that you can include that pulls in everything you need, the rest is implementation detail.

TomSaw:
> It's mainly just historic convention to have just one header file which pulls in everything that is needed. We can remove this construct completely and instead just #pragma once and then #include "st7586s.hpp", which will also make IDE code completion work in this file, since it's pulling in the correct definitions.
> 
> (Ok, this file is dependency free, so you don't need the #include "st7586s.hpp" part, but in the st7586s.hpp_impl.hpp file that can be used).

😋 Smaller Boilerplate... Wunderbar! Copy that

src/modm/math/utils/bit_operation.hpp
twasilczyk: Done.

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: Genius. Thanks!

I didn't add the include to the *protocol details* ( ;) ) file, since it doesn't depend on the main one indeed and there may be valid uses for it separately.

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: Holy cow. Didn't you have a longer name for a regular assertion that doesn't get executed in debug mode? ;)

I was rather hoping for something along the lines:
- CHECK(condition) - evaluates in debug build only
- CHECK_ALWAYS(condition) - evaluates both in debug and release

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: I'm afraid you can't define a namespace inside a class.
```
error: 'payload' in namespace 'modm::impl::st7586s' does not name a type
```

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: Ditto.

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: Hah. I was actually hoping it would work this way, but the API scared me off. As it turns out, it wasn't API, but a mere implementation with swapped argument naming:

gpio_D4.hpp:
```
void setOutput(bool status);
```

While the TLD interface (gpio.hpp) has it:
```
void setOutput(bool value);
```

@salkinium: is this a typo? I was digging what `status` means and I had an impression it's something about pull-ups or port config.

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: Done.

src/modm/driver/display/st7586s_impl.hpp
twasilczyk: Done.

src/modm/driver/display/st7586s_protocol.hpp
twasilczyk: Done.

twasilczyk:
> > avr-gcc shipped with modm doesn't support std::is_constant_evaluated (that's C++20 symbol)
> 
> [GCC supports it since v9](https://gcc.gnu.org/projects/cxx-status.html), so I suspect maybe a problem with combining `constexpr` and `modm_always_inline`?

Looks like it's just not there:
```
error: 'is_constant_evaluated' is not a member of 'std'
```

I know the root cause is likely in modm's avr-gcc fork and it's fixable there. But I'm already pretty deep in this rabbit hole :).

> > why SPI::transferBlocking requires uint8_t* and doesn't just accept void* pointers? See sendCommand
> 
> `void*` just requires more casting for the average use-case, since you typically store buffers as `uint8_t*`.

Does it? I'm pretty sure there's actually no casting needed in user code if the function accepts `void*` or `const void*`. Is there?

salkinium (approved):
I'm happy with this, thanks!

src/modm/driver/display/st7586s_impl.hpp
salkinium: It's a typo, it should reflect the architecture interface API.

src/modm/driver/display/st7586s_impl.hpp
salkinium: Yeah, not happy about it either. Here's some explanation of what I'm trying to describe: https://modm.io/reference/module/modm-architecture-assert/#assertion-types

twasilczyk:
Sorry for push+force push (the first one was a WIP commit on a new driver, the second was to remove it). Current state should be the same as approved.

salkinium:
Whoops I wanted to rebased, but pushed *my* develop branch to your fork. I merged it manually.
Let's use feature branches in future ;-P

salkinium:
> I know the root cause is likely in modm's avr-gcc fork and it's fixable there. But I'm already pretty deep in this rabbit hole :).

No, it's probably an issue with our custom libstdcpp not having set the [`_GLIBCXX_HAVE_BUILTIN_IS_CONSTANT_EVALUATED `](https://github.com/modm-io/avr-libstdcpp/blob/master/include/type_traits#L3286) define? It's a bit weird, since [the lib config](https://github.com/modm-io/avr-libstdcpp/blob/master/include/bits/c%2B%2Bconfig.h#L644-L670) should enable it automatically? cc @chris-durand

chris-durand:
@twasilczyk Did you include the `type_traits` header for `std::is_constant_evaluated`? I can't reproduce the issue.

twasilczyk:
Huh, it was missing `type_traits`. Apparently when built for ARM, this header gets pulled in by some other header. I'll make a follow-up PR, but not today.

PR #674, opened 2021-09-03
Fix SPI mode descriptions

twasilczyk:
Comments here contradict what `src/modm/platform/spi/stm32/spi_master.hpp.in` (and Wikipedia) says.

### DIFF
--- a/src/modm/architecture/interface/spi.hpp
+++ b/src/modm/architecture/interface/spi.hpp
@@ -32,8 +32,8 @@ struct Spi
 	{
 		Mode0 = 0b00,	///< clock normal,   sample on rising  edge
 		Mode1 = 0b01,	///< clock normal,   sample on falling edge
-		Mode2 = 0b10,	///< clock inverted, sample on rising  edge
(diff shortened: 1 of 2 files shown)


chris-durand (approved):
Thanks!

chris-durand:
I just noticed the AVR SPI driver has [mode 1 and 2 backwards](https://github.com/modm-io/modm/blob/a173bde68b7ce0b4b285e82c1433c49c4c76dc83/src/modm/platform/spi/at90_tiny_mega/spi_master.hpp.in#L64), **Mode 1** shouldn't have **CPOL** set but **CPHA**. And the comments are wrong as well.
I fixed it although it might break someone who relies on the old behaviour. @salkinium What do you think?

salkinium (approved):
> I fixed it although it might break someone who relies on the old behaviour. @salkinium What do you think?

We fix the bug and mention it in the changelog. Kinda weird that in all the years nobody noticed this bug on AVRs, probably since it wasn't used? The breakage can't be too bad then… #InfallibleLogic

PR #679, opened 2021-09-09
Support USB device on samg family

mcbridejc:


src/modm/board/samg55_xplained_pro/board.hpp
```diff
@@ -51,5 +66,12 @@ initialize()
 	Button::setInput();
 }
 
+inline void initializeUsbDevice()
```
salkinium:
```suggestion
inline void initializeUsbFs()
```

The idea is to be able to compile the [`example/generic/usb` example](https://github.com/modm-io/modm/blob/develop/examples/generic/usb/main.cpp#L41), by only switching the BSP in the `project.xml`. Not entirely sure if there are more changes required though.

examples/samg55_xplained_pro/usbserial/main.cpp
```diff
@@ -0,0 +1,31 @@
+
+#include "modm/platform.hpp"
+
+#include "modm/board.hpp"
```
salkinium:
```suggestion
#include <modm/board.hpp>
```

to indicate library vs local files. Just a convention though.

salkinium (approved):
Nice! Squash?

PR #685, opened 2021-09-13
Fix ADC shared interrupt generation error and fail loudly on bad IRQ names in MODM_ISR

### LINKED ISSUE #663: ADC Undefined_Handler called
Hi Folks,
Just a quick sanity check please. STM32F429 (F4 discovery) ADC
There is a single global ADC interrupt on this device
image

Is modm incorrectly generating 3 separate named IRQHandlers?
Could this be why modm_abandon() reports Undefined_Handler exception when trying to use interrupts?
````
name = 0x08078570 "nvic.undef"
"An undefined NVIC interrupt was raised!"
context = 18
````
According to the Linker MAP file Cross Reference Table:
Symbol    
````
ADC1_IRQHandler                                   ../modm/CMakeFiles/modm.dir/src/modm/platform/adc/adc_interrupt_1.cpp.obj
ADC2_IRQHandler                                   ../modm/CMakeFiles/modm.dir/src/modm/platform/adc/adc_interrupt_2.cpp.obj
ADC3_IRQHandler                                   ../modm/CMakeFiles/modm.dir/src/modm/platform/adc/adc_interrupt_3.cpp.obj
ADC_IRQHandler                                    ../modm/CMakeFiles/modm.dir/src/modm/platform/core/vectors.c.obj
````

For the UARTs - Its as expected - the names match:
````
USART1_IRQHandler                                 ../modm/CMakeFiles/modm.dir/src/modm/platform/uart/uart_1.cpp.obj
                                                  ../modm/CMakeFiles/modm.dir/src/modm/platform/core/vectors.c.obj
USART2_IRQHandler                                 ../modm/CMakeFiles/modm.dir/src/modm/platform/uart/uart_2.cpp.obj
                                                  ../modm/CMakeFiles/modm.dir/src/modm/platform/core/vectors.c.obj
USART3_IRQHandler                                 ../modm/CMakeFiles/modm.dir/src/modm/platform/uart/uart_3.cpp.obj
                                                  ../modm/CMakeFiles/modm.dir/src/modm/platform/core/vectors.c.obj
USART6_IRQHandler                                 ../modm/CMakeFiles/modm.dir/src/modm/platform/uart/uart_6.cpp.obj
                                                  ../modm/CMakeFiles/modm.dir/src/modm/platform/core/vectors.c.obj
````

### LINKED ISSUE #684: Should incorrect IRQ name in MODM_ISR fail louder?
One of the problems I had to debug lately was about  incorrectly generated IRQ names for the samg platform. 

At the moment, you can create an ISR with any name, and it just doesn't get called. I got to thinking about how to statically assert that it's a valid vector. I haven't found a way to get the assert to actually fail, but you can generate a compile error. While trying this out, I discovered that the STM32 ADC driver is currently creating ADC1_IRQHandler, while on the STM32F413 the actual interrupt is ADC_IRQHandler. 

```
modm/platform/adc/adc_interrupt_1.cpp:20:10: error: 'ADC1_IRQn' was not declared in this scope; did you mean 'ADC_IRQn'?
   20 | MODM_ISR(ADC1)
```
What I've currently got in interrupt.hpp:
```
#	define MODM_ISR(vector, ...) \
		static_assert(!std::is_void<decltype(vector ## _IRQn)>::value, "No such IRQ " #vector); \
		modm_extern_c void vector ## _IRQHandler(void) \
			__attribute__((externally_visible)) __VA_ARGS__; \
		void vector ## _IRQHandler(void)
```

From https://github.com/mcbridejc/modm/blob/feature/IRQ_assert/src/modm/architecture/interface/interrupt.hpp#L114

 Thoughts?

### PR DESCRIPTION
**Pull request by mcbridejc**
The `shared_irq_ids` was being populated with strings, e.g. `['1']`, but was being tested with integers (e.g. `if any (i in props["shared_irq_ids"] for i in props["instances"])`).

I also went ahead and included the invalid IRQ check that led me to find this. If you see any issue with that though, I can drop it from the PR. 

What's not entirely clear to me is if this worked on some platform? Perhaps on some platform the instance IDs come in as ints, either because the XML attribute is an int or something in modm-devices converts them. 

Thanks!

### DIFF
--- a/src/modm/architecture/interface/interrupt.hpp
+++ b/src/modm/architecture/interface/interrupt.hpp
@@ -62,6 +62,12 @@
  * This maps to `MODM_ISR_NAME({vector})()`.
  * @note You may have to forward declare the interrupt using `MODM_ISR_DECL({vector})`.
  *
+ * On Cortex-M the vector name is validated at compile time against the device's
+ * vector table, so that misspelled vector names get caught during compilation.
+ * If this is undesirable you can manually call the vector using
+ * `MODM_ISR_NAME(vector)()` or disable the validation by defining
+ * `MODM_ISR_DISABLE_VALIDATION`.
+ *
  * @param vector
  *        The name of the interrupt without any suffix (neither `_vect`, nor `_IRQHandler`).
  *
@@ -78,6 +84,10 @@
  * On AVRs this maps to `ISR(MODM_ISR_NAME({vector}), args)`.
  * On Cortex-M and Hosted this maps to `void MODM_ISR_NAME({vector})(void) args`.
  * `extern "C"` is automatically added in a C++ environment.
+ *
+ * On Cortex-M the vector name is validated at compile time against the device's
+ * vector table, so that misspelled vector names get caught during compilation.
+ * You can disable the validation by defining `MODM_ISR_DISABLE_VALIDATION`.
  *
  * @param vector
  *        The name of the interrupt without any suffix (neither `_vect`, nor `_IRQHandler`).
@@ -103,13 +113,27 @@
 
 #elif defined MODM_CPU_ARM
 
+
+#ifdef __cplusplus
+#	include <modm/platform/core/vectors.hpp>
+#else
+#	ifndef MODM_ISR_DISABLE_VALIDATION
+#  		define MODM_ISR_VALIDATE(vector_str, vector) \
+			_Static_assert(vector ## _IRQn > -127);
+#	else
+#   	define MODM_ISR_VALIDATE(...)
+#	endif
+#endif
+
 #	define MODM_ISR_NAME(vector) \
 		vector ## _IRQHandler
 #	define MODM_ISR_DECL(vector) \
 		modm_extern_c void vector ## _IRQHandler(void)
 #	define MODM_ISR_CALL(vector) \
+		MODM_ISR_VALIDATE(#vector, vector); \
 		vector ## _IRQHandler()
 #	define MODM_ISR(vector, ...) \
+		MODM_ISR_VALIDATE(#vector, vector); \
 		modm_extern_c void vector ## _IRQHandler(void) \
 			__attribute__((externally_visible)) __VA_ARGS__; \
 		void vector ## _IRQHandler(void)
--- a/src/modm/platform/adc/stm32/adc_impl.hpp.in
+++ b/src/modm/platform/adc/stm32/adc_impl.hpp.in
@@ -250,7 +250,5 @@ modm::platform::Adc{{ id }}::getInterruptFlags()
 void
 modm::platform::Adc{{ id }}::acknowledgeInterruptFlags(const InterruptFlag_t flags)
 {
-	// Flags are cleared by writing a one to the flag position.
-	// Writing a zero is ignored.
-	ADC{{ per }}->SR = flags.value;
+	ADC{{ per }}->SR = ~flags.value;
 }
--- a/src/modm/platform/adc/stm32/adc_interrupt.cpp.in
+++ b/src/modm/platform/adc/stm32/adc_interrupt.cpp.in
@@ -18,7 +18,7 @@ modm::platform::AdcInterrupt{{ id }}::Handler
 modm::platform::AdcInterrupt{{ id }}::handler(modm::dummy);
 
 %% if id not in shared_irq_ids
-MODM_ISR(ADC{{ id }})
+MODM_ISR(ADC{{ irq }})
 {
     if (modm::platform::AdcInterrupt{{ id }}::getInterruptFlags()) {
         modm::platform::AdcInterrupt{{ id }}::handler();
--- a/src/modm/platform/adc/stm32/module.lb
+++ b/src/modm/platform/adc/stm32/module.lb
@@ -49,6 +49,7 @@ class Instance(Module):
         properties["channels"] = sorted(channels)
         global props
         properties["shared_irq_ids"] = props["shared_irq_ids"]
+        properties["irq"] = "1" if target.family in ["l1"] else self.instance
         props["instances"].append(self.instance)
 
         env.substitutions = properties
--- a/src/modm/platform/adc/stm32/module.lb
+++ b/src/modm/platform/adc/stm32/module.lb
@@ -76,7 +76,7 @@ def prepare(module, options):
     props["instances"] = []
 
     if target["family"] in ["f2", "f4", "f7"]:
-        props["shared_irqs"] = {"ADC": listify(device.get_driver("adc")["instance"])}
+        props["shared_irqs"] = {"ADC": listify([int(i) for i in device.get_driver("adc")["instance"]])}
         props["shared_irq_ids"] = props["shared_irqs"]["ADC"]
     else:
         shared_irqs = [v["name"] for v in device.get_driver("core")["vector"]]
--- a/src/modm/platform/can/stm32/can.cpp.in
+++ b/src/modm/platform/can/stm32/can.cpp.in
@@ -121,16 +121,16 @@ modm::platform::Can{{ id }}::initializeWithPrescaler(
 	NVIC_EnableIRQ(CEC_CAN_IRQn);
 %% else
 	// Set vector priority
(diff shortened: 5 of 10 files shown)

### CODE REVIEW

**Comment by salkinium** [comment]
Oops, I forgot about this bug: https://github.com/modm-io/modm/issues/663#issuecomment-888237502
There's another bug in this driver, please also fix it: https://github.com/modm-io/modm/issues/663#issuecomment-888304534

File: src/modm/architecture/interface/interrupt.hpp [hunk 1]
```diff
@@ -110,6 +111,7 @@
 #	define MODM_ISR_CALL(vector) \
 		vector ## _IRQHandler()
 #	define MODM_ISR(vector, ...) \
+		static_assert(!std::is_void<decltype(vector ## _IRQn)>::value, "No such IRQ " #vector); \
```
salkinium: You probably want to `#ifdef __cplusplus` this and the header include, since these macro are also callable from a pure C context.

salkinium: I would prefer some kind of C version of this check, as C17 supports `static_assert` via `<assert.h>`. We can include a platform specific header file that contains a list of some sort with the generated vector table names, so that's easy, I just don't know how to check the membership of a string (?) in a set of strings in C at compile time.

salkinium: Ha, perhaps we can even generate the static_assert message? That would be cool "This is not a valid IRQ, but here's a list of all valid IRQs". We do this for [the STM32F1 remap groups](https://modm.io/how-modm-works/#modm-asserts-at-compile-time).

mcbridejc: Hmm...OK good point. It's pretty easy to use the C assert, and that's probably what we should do. But that makes me a little sad, because I just learned about the `string_view_literal` and something like this with a python generated list of valid IRQ name strings would let us actually fail the assert and print whatever we want in C++. 

```
#include <string_view>

using namespace std::string_view_literals;
constexpr std::array valid_strings = {"Word"sv, "list"sv};

constexpr bool valid_word(const std::string_view &s) {
    for(const auto &valid : valid_strings) {
        if(valid == s) return true;
    }
    return false;
}
static_assert(valid_word("Word"), "Why'd this fail?");
static_assert(valid_word("ABC"), "This should fail!");
```

Sadly, I've no idea how to do it in C.

mcbridejc:
~sigh~...turns out the CAN driver depends on defining extra ISRs: https://github.com/modm-io/modm/blob/develop/src/modm/platform/can/stm32/can.cpp.in#L310

src/modm/architecture/interface/interrupt.hpp
salkinium: That's ok, we can have different implementations for C and C++ via `#ifdef`. It's ok if we don't immediately have a C version, the C++ version is very, very good, you should definitely implement that!

salkinium:
> turns out the CAN driver depends on defining extra ISRs:

Then let's add perhaps another macro: `MODM_ISR_UNCHECKED`?
Edit: Alternatively we should actually check if the ISR exists for the CAN driver rather than using the family as a proxy test.

mcbridejc:
> > turns out the CAN driver depends on defining extra ISRs:
> 
> Then let's add perhaps another macro: `MODM_ISR_UNCHECKED`?
> Edit: Alternatively we should actually check if the ISR exists for the CAN driver rather than using the family as a proxy test.

See what you think of what I just committed. I don't love MODM_ISR_UNCHECKED; I'd rather get the error if the CAN (or any other) driver mistakenly generates the wrong interrupt.

salkinium:
> See what you think of what I just committed.

Yeah, much better to fix this for real than hack around it.

> static_assert(vector ## _IRQn > -127);

This is also good enough for C even if it won't fail gracefully. Still better than nothing!

mcbridejc:
Another failing platform: 
```
 x   stm32l100rct6          42.4s
In file included from modm/src/modm/platform/adc/adc_interrupt_.cpp:15:
modm/src/modm/platform/adc/adc_interrupt_.cpp:20:10: error: 'ADC_IRQn' was not declared in this scope; did you mean 'DAC_IRQn'?
   20 | MODM_ISR(ADC)
```

It seems like what's happening here is that the instance ID is empty for some (all?) lxxx series devices. Probably because the XML doesn't have an instance tag under the adc:

```
</driver>
    <driver name="adc" type="stm32"/>
    <driver device-size="6|8|b" device-variant="" name="comp" type="stm32-v3.4">
      <instance value="1"/>
      <instance value="2"/>
    </driver>
```

The correct interrupt for this platform is ADC1, but in this case I don't think it's trying to do the combined IRQ, I think it just didn't get a valid instance. 

The C version works well. The C++ version is giving me trouble. Something about the preprocessor is breaking it; interrupts which *are* in the list still fail. E.g. MODM_ISR(ADC), passed through the macros fails, but if I just `static_assert(isValidIrqName("ADC"))` that's fine. 

```
// validate_irq.hpp.in
#include <array>
#include <cstdint>
#include <string_view>

namespace modm::platform {

namespace valid_irqs {
	using namespace std::string_view_literals;
	constexpr std::array list = {
	%% for name in short_irq_names
		"{{ name }}"sv {{ "," if not loop.last }}
	%% endfor
	};
} // namespace valid_irqs

/** Allows static validation of an IRQ name against the target IRQs */
constexpr bool isValidIrqName(const std::string_view &name) {
	for(const auto &test_name : valid_irqs::list) {
		if(name == test_name) {
			return true;
		}
	}
	return false;
}
```
```
// interrupt.hpp
// Validate that the IRQ is a valid one for the platform.
// In C++, assertion fails and message is printed. In C, a compiler error is generated
// because the referenced symbol is not found.
#ifdef __cplusplus
#	define IRQ_ASSERT(name) static_assert(modm::platform::isValidIrqName(#name), "Invalid IRQ for platform")
#else
#	define IRQ_ASSERT(name) static_assert(name ## _IRQn >= 0)
#endif

#	define MODM_ISR(vector, ...) \
		IRQ_ASSERT(vector); \
		modm_extern_c void vector ## _IRQHandler(void) \
			__attribute__((externally_visible)) __VA_ARGS__; \
		void vector ## _IRQHandler(void)
```

The output of the C++ isn't so much better than the C. `static_assert` needs the message arg to be a string literal, and so I haven't figured out how to a) get the invalid IRQ name into the message or b) how to get a message generated in the validate_irq.hpp header into the static_assert in interrupt.hpp. Maybe a preprocessor macro could resolve b.

salkinium:
> E.g. MODM_ISR(ADC), passed through the macros fails, but if I just static_assert(isValidIrqName("ADC")) that's fine.

There's something about recursive macro resolving, try using MODM_STRINGIFY(name) instead of #name.

>  a) get the invalid IRQ name into the message or b) how to get a message generated in the validate_irq.hpp header into the static_assert in interrupt.hpp. Maybe a preprocessor macro could resolve b.

Try something like this?

```cpp
#define MODM_ISR_VALIDATE_ERROR_MESSAGE " is not a valid IRQ name."
#define IRQ_ASSERT(name) static_assert(modm::platform::isValidIrqName(MODM_STRINGIFY(name)), MODM_STRINGIFY(name) MODM_ISR_VALIDATE_ERROR_MESSAGE)
```

salkinium:
> It seems like what's happening here is that the instance ID is empty for some (all?) lxxx series devices. Probably because the XML doesn't have an instance tag under the adc:

ST is weird and calls this peripheral just ADC not ADC1. hence there is no instance, therefore we only have `modm:platform:adc` and not `modm:platform:adc:1`. I really don't like that, but I don't have a better idea than to follow the ST naming convention. It's ugly.

Edit: Actually I think your right, this seems to be a bug in modm-devices…

salkinium:
I added a constexpr version of your idea.

salkinium:
A slight disadvantage of this implementation is that it does not "respect" ST's redefinitions of IRQ handlers, thus you have to use the real IRQ name and not ST's idea of "painless codes [sic] migration". 🙄

mcbridejc:
> I added a constexpr version of your idea.

Nice! I've been trying to finish something else up before coming back to this.

salkinium (approved):

PR #741, opened 2021-09-29
[hal] Support 16bit transfers for I2c

TomSaw:
I like the idea of moving 16bit treatment from drivers to Hal. Dozens of I2c-drivers contain endian sanities (more or less any call of 'modm::BigEndian()' in src/modm/driver/** ) before or after talking to I2c.

point_right In addition to cleaner code we gain slightly shrinked compilate and slightly improved performance because data >> 8 in Hal is cheaper than modm::fromBigendian(data) in Drivers.

Same currently happens with Spi #690
I2c is a little more tricky so i postpone this and concentrate on merging #665 first.

### DIFF
--- a/src/modm/architecture/interface/i2c_device.hpp
+++ b/src/modm/architecture/interface/i2c_device.hpp
@@ -40,7 +40,7 @@ namespace modm
  * @author	Niklas Hauser
  * @ingroup modm_architecture_i2c_device
  */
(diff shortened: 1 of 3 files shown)


TomSaw:
Renamed branch to `feature/i2c-16bit-transfer` and created new PR: #812
because it's not too late for this youngster PR to get a better branch-name.

PR #744, opened 2021-10-04
Fix for ambigious gpio

asmfreak:
I've got a problem compiling basic example for blue_pill.
Compiler (`arm-none-eabi-c++ (Arch Repository) 11.1.0`) was complaning about `Gpio` struct ambiguity. The this PR fixes that ambiguity.

### DIFF
--- a/src/modm/platform/gpio/common/unused.hpp.in
+++ b/src/modm/platform/gpio/common/unused.hpp.in
@@ -106,13 +106,13 @@ public:
 	struct BitBang
 	{
 		using Data = detail::DataUnused;
-		static constexpr Gpio::Signal Signal = Gpio::Signal::BitBang;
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Thanks, but be aware that the official arm-none-eabi-gcc from ARM is still at v10.2, therefore our CI won't catch this. There may also be other weird behavior, due to Arch using different build scripts and patches.

salkinium:
Actually I take that back, GCC v10 does warn about it!

> modm/src/modm/platform/gpio/unused.hpp:52:7: warning: direct base 'modm::Gpio' inaccessible in 'modm::platform::GpioUnused' due to ambiguity [-Winaccessible-base]
   52 | class GpioUnused : public Gpio, public ::modm::GpioIO

rleh (approved):

PR #749, opened 2021-10-15
[stm32] CanFilter on devices with two bxCan peripherals

rleh:
Enable `CanFilter::setStartFilterBankForCan2(uint8_t)` for all devices with two CAN peripherals.

The code was not generated on some F0, F7 and L4 devices, e.g. STM32L496/STM32L4A6.

Previously the code was generated for all STM32F1, F2 and F4 targets, even if there is no second CAN peripheral. This (breaking) change could therefore lead to errors in some existing (incorrect) application code, as in our STM32F103 examples.

### DIFF
--- a/src/modm/platform/can/stm32/module.lb
+++ b/src/modm/platform/can/stm32/module.lb
@@ -184,6 +184,7 @@ def build(env):
         properties["combined_isr"] = True
     else:
         properties["combined_isr"] = False
+    properties["target_has_can2"] = (("instance" in driver) and len(driver["instance"]) > 1)
 
     if "instance" not in driver:
         properties.update(get_substitutions(0, device, env))
--- a/examples/blue_pill_f103/can/main.cpp
+++ b/examples/blue_pill_f103/can/main.cpp
@@ -41,12 +41,9 @@ main()
 
 	MODM_LOG_INFO << "CAN Test Program" << modm::endl;
 
-	MODM_LOG_INFO << "Dividing filter bank..." << modm::endl;
-	CanFilter::setStartFilterBankForCan2(14);
+	MODM_LOG_INFO << "Initializing Can..." << modm::endl;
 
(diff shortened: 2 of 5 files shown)


salkinium (approved):
Nice!

src/modm/platform/can/stm32/module.lb
```diff
@@ -184,6 +184,7 @@ def build(env):
         properties["combined_isr"] = True
     else:
         properties["combined_isr"] = False
+    properties["target_has_can2"] = env.has_module(":platform:can:2")
```
salkinium: You probably want to check the device file, rather than the lbuild config for this?

rleh: It is useless to call `setStartFilterBankForCan2(_)` if `:can:2` is not included in the project, but I admit it might be better to have the function generated independent of the usage of `:can:2`.

rleh: What is the best way to query the device file for the existence of the second can peripheral?
Some other platform driver as an example?

salkinium: I know, we really need a proper Python API for this crap.

See how the instances are accessed in the prepare method just above.

src/modm/platform/can/stm32/module.lb
```diff
@@ -184,6 +184,7 @@ def build(env):
         properties["combined_isr"] = True
     else:
         properties["combined_isr"] = False
+    properties["target_has_can2"] = (len(device.get_driver("can")["instance"]) > 1)
```
salkinium:
```suggestion
    properties["target_has_can2"] = (len(device.get_driver("can").get("instance", [])) > 1)
```

chris-durand (approved):

rleh:
Merge?

PR #750, opened 2021-10-18
[freertos] allow use FreeRTOS without heap dependency

andryblack:
Move dependency :platform:heap from  :freertos  to :processing:rtos.
This allow use FreeRTOS directly without injecting heap support.

Signed-off-by: andryblack


salkinium (approved):
A simple and effective change, thanks!

I fixed some formatting, simplified the config a little and added an example so that this behaviour gets checked by the CI for future refactorings.

PR #755, opened 2021-10-19
[stm32] G4 / L4+ core voltage boost mode

chris-durand:
For reliable operation at full frequency of G4 and L4+ devices ST recommends enabling the core voltage scaling boost mode. The setting is added to the clock driver and implements a logic equivalent to ST's G4 CubeHAL. The present G4 examples are adapted to set the mode accordingly.

The change has been tested on a Nucleo G474RE.


salkinium (approved):
Thanks!

PR #757, opened 2021-10-20
[driver] TMP12x temperature sensor

chris-durand:
Simple driver for TMP121/TMP123/TMP125 temperature sensors. The driver has been in use for some time on a custom board with an STM32F042K6. I'll test the Nucleo example tomorrow.

### DIFF
--- a/src/modm/driver/temperature/tmp12x.hpp
+++ b/src/modm/driver/temperature/tmp12x.hpp
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2020, Christopher Durand
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_TMP12x_HPP
+#define MODM_TMP12x_HPP
+
+#include <modm/architecture/interface/spi_device.hpp>
+#include <modm/processing/resumable/nested_resumable.hpp>
+#include <array>
+#include <ratio>
+
+namespace modm
+{
+
+/**
+ * TMP12x temperature data
+ * Use aliases Tmp121Temperature, Tmp123Temperature, Tmp125Temperature
+ * for the respective sensor.
+ *
+ * @ingroup modm_driver_tmp12x
+ * @author	Christopher Durand
+ *
+ * @tparam numBits Number of bits of measured temperature
+ * @tparam offsetBits Invalid LSB of raw sensor value
+ * @tparam numerator
+ * @tparam denominator
+ */
+template<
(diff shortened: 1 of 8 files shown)


rleh:
The only difference between TMP121 und TMP123 are the swapped GND and NC pins, or am I missing something?

src/modm/driver/temperature/tmp12x.hpp
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	modm::ResumableResult<Temperature>
+	read()
+	{
+		RF_BEGIN();
+
+		RF_WAIT_UNTIL(this->acquireMaster());
+		Cs::reset();
+		SpiMaster::setDataMode(SpiMaster::DataMode::Mode0);
```
rleh: This should be done in the [`ConfigurationHandler`](https://docs.modm.io/develop/api/stm32f042t6y6/classmodm_1_1SpiDevice.html), like e.g. in the the [DRV832x driver](https://github.com/modm-io/modm/blob/02660d04bf633ae619daa52ce5e19d24e546ede9/src/modm/driver/motor/drv832x_spi_impl.hpp#L22-L25).

src/modm/driver/temperature/tmp12x.hpp
```diff
@@ -0,0 +1,163 @@
...
+			Cs::set();
+		}
+
+		RF_END_RETURN(Temperature(buffer_[1] | (buffer_[0] << 8)));
+	}
+private:
+	std::array<uint8_t, 2> buffer_{};
+};
+
+/**tmp12x_test
```
rleh:
```suggestion
/**
```

rleh (commented):
Nice driver 👍🏽

chris-durand:
> The only difference between TMP121 und TMP123 are the swapped GND and NC pins, or am I missing something?

Most likely the only difference is the chip marking. For the ones I had the NC pins were bonded to GND.

src/modm/driver/temperature/tmp12x.hpp
chris-durand: Shouldn't we also set the data size? If a DRV832x was shared on a bus with another device with a different setting the configuration wouldn't be reset to 8 bit.

rleh: Yes.

chris-durand: A lot of SPI drivers which use the default settings don't have configuration handlers at all. We should fix that at some point.

rleh (approved):

src/modm/driver/temperature/tmp12x.hpp
rleh: Absolutely! Somebody™️ just has to do (and test) it...

salkinium (approved):
Very nice!

chris-durand:
Please don't merge until I have tested the example.

chris-durand:
I have tested the example and it works.

PR #762, opened 2021-10-23
[cortex] Fix overflow in millisecond delay

chris-durand:
The current implementation of Cortex-M millisecond delay assumes calling `delay_us` with a delay of 1s is allowed. This is not true for all supported targets. The worst case right now are STM32G4 at 170 MHz for which a delay of 790 ms overflows.
The limit is `max_us_delay = (2^32-1) / (core_clock_MHz * 2^us_shift)`. [For G4 the shift is 5, for H7 3.](https://github.com/modm-io/modm/blob/6e9f000e0ce22bc66af4e42f60255d041ed6234a/src/modm/platform/core/stm32/module.lb#L38)

~~With this fix delay_ms will call `delay_us(500'000)` instead of `delay_us(1'000'000)` to implement long delays.
Tested in hardware on Nucleo H723ZG at 550 MHz and Nucleo G474RE at 170 MHz.~~

`us shift` will be reduced for G4 and H7 to allow for a maximum us delay of at least 1s.

### DIFF
--- a/src/modm/platform/core/sam/module.lb
+++ b/src/modm/platform/core/sam/module.lb
@@ -33,19 +33,22 @@ def build(env):
     env.template("startup_platform.c.in")
 
     # delay code that must be tuned for each family
-    # (cycles per loop, setup cost in loops, us shift)
+    # (cycles per loop, setup cost in loops, max cpu frequency)
     tuning = {
-        "d": (3, 4, 5), # CM0 tested on D21 in RAM
-        "g": (6, 4, 5), # G55
-        "v": (4, 4, 3), # V70
-    }.get(target.family, (4, 4, 5))
+        "d": (3, 4,  48), # CM0 tested on D21 in RAM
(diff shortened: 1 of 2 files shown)


salkinium:
No, the `us_shift` needs to be reduced, it's only an optimization to allow less error to accumulate during the looping.
The 1000ms/µs/ns delays are written into the documentation and it would be a breaking change to reduce that now.

salkinium:
Yes, I'll also add a change that validate the setting in `module.lb`, since this is a really dumb bug that's super hard to debug. Kinda annoyed at myself that I didn't do this right.

salkinium:
I've added max-frequency from CubeMX to the STM32 device files and derive the us_shift directly now. This also makes porting a little easier.

salkinium:
I also changed the SAM module to use (manually specified) max frequency instead of us_shift. Just way easier for porting.

salkinium (approved):

PR #763, opened 2021-10-23
[stm32] Fix linkerscript veneer offset issues

chris-durand:
There is a linkerscript bug that can cause code placed in RAM to be loaded to the wrong address because alignment is ignored. By chance the bug showed on some F0/L0 after the `delete` stub implementations were added 9 days ago.

We could reproduce the bug with the blink examples on F042, F030 and L031 devices but more devices are affected. It caused the delay to run with random durations. The instructions loading `modm::platform::delay_fcpu_MHz` from ram to a register were never executed.

This bug was most likely introduced in the last [linkerscript refactoring](https://github.com/modm-io/modm/commit/996e361d9ed9b59f0a4928bc3ddec22862d6665b) in July.

Thanks to @salkinium for coming up with the fix after it was clear what the issue was.

### DIFF
--- a/src/modm/platform/core/cortex/linker.macros
+++ b/src/modm/platform/core/cortex/linker.macros
@@ -15,24 +15,32 @@
 %% endmacro
 
 %% macro prefix()
-MEMORY
-{
-%% for memory in (memories + cont_ram_regions)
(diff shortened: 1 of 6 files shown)


salkinium:
I still need to validate this for all other devices, so not merging yet.

salkinium (approved):

salkinium:
Validated in hardware on all linkerscripts variants.
I also cleaned up some things in the templates so that looking at the generated linkerscripts is a little less painful.

PR #765, opened 2021-10-25
Extend pin connect types

mcbridejc:
I was taking the approach of having drivers add these as they were implemented, but I've realized that having them all allows drivers created outside of modm repo to implement connect() methods.

### DIFF
--- a/src/modm/platform/gpio/sam/pin.hpp.in
+++ b/src/modm/platform/gpio/sam/pin.hpp.in
@@ -33,6 +33,8 @@ enum class PeripheralPin
 	BitBang,
 	Rx,
 	Tx,
+	Rts,
+	Cts,
 	ExtInt,
 	Dm,
 	Dp,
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Sure, however, keep in mind that it's better to collect all the signal names from the device file and then generate this list otherwise you're going to probably miss some signals.

mcbridejc:
> Sure, however, keep in mind that it's better to collect all the signal names from the device file and then generate this list otherwise you're going to probably miss some signals.

Perhaps, but I think there is a fair bit of oddball naming to work around, and some of it I'm not at all sure you can work around. There are redundant names, e.g. where a signal is listed as miso and spi3_miso. OK so throw out names with an underscore maybe. But then you get into the samd parts, and I think some pins are just called "pad", connected to the sercom device. There's no 'scl' or 'miso' to be found in the xml, so if you want to connect by this name -- and I think you do -- it needs to get hardcoded somewhere. 

I have no motiviation right now to dive into these weeds, or put the effort needed to make sure it doesn't break on other chips, so I'm just figuring a somewhat more complete list is somewhat better than a less complete one! :)

PR #770, opened 2021-10-30
[samg] Make Adc::getPinChannel consistent with STM32 interface

mcbridejc:
Previously, you pass GpioA17::Ad as the template argument, but in the
stm32 driver you pass just GpioA17. This makes the SAMG driver
consistent with that.


salkinium (approved):
Good catch!

PR #771, opened 2021-10-30
[samg] Add faster transmit only method to SPI driver

mcbridejc:
Synchronous transfer is slowed down by the having to wait for a complete
transfer to read the received byte. When only transmit is required, this
method allows for writing the next transmit byte while the current byte
is being written out to increase throughput.

Also removes the 'setOutput' calls in the connect() method, as these are not necessary.


mcbridejc:
I was seeing some pretty significant inter-byte delay running at 6MHz SCK: 
image
(before and after)

I'd like to add a similar method to the STM32 driver as well.

salkinium:
Sure, but make it protected and call it `transferBlocking ` via operator overloading, since we're not extending the interface to call that function directly.

salkinium (approved):
Thanks!

PR #772, opened 2021-10-30
[stm32] H7 DMA + DMA data fixes + fix DAC DMA

chris-durand:
This adds support for H7 DMA. The device data [has to be fixed first](https://github.com/modm-io/modm-devices/issues/74). H7 devices use the same DMA IP as F4/F7 devices but use the DMAMUX peripheral for request multiplexing.

The updates in modm-devices fix broken DMA request data for many G0, L4, H7 and WL devices.

- [x] Adapt DMA driver for H7
- [x] Fix requests mappings in modm-devices data
- [x] Handling for L4+ DMAMUX data with shared request for two peripherals (DCMI/PSSI)
- [x] Fix DAC DMA driver for F2, F4, F7, H7
- [x] Add DAC DMA example if missing and test
   - [x] G4
   - [x] H7
   - [x] L4
   - [x] F4
   - [x] F7
- [x] Test H7 DMA

### DIFF
--- a/examples/nucleo_f446ze/dac_dma/main.cpp
+++ b/examples/nucleo_f446ze/dac_dma/main.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2021, Christopher Durand
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <modm/board.hpp>
+
+#include <numbers>
+#include <cmath>
+#include <array>
+
+using namespace Board;
+
+constexpr auto computeSinTable(float frequency = 1.f)
+{
+	std::array<uint16_t, 100> data{};
+	constexpr auto HalfOutput = ((1 << 12) - 1) / 2; // 12 bit dac
+	for (size_t i = 0; i < data.size(); ++i) {
+		constexpr auto pi = std::numbers::pi_v<float>;
+		data[i] = HalfOutput * (1 + sin(i * (2 * pi * frequency / data.size())));
+	}
+	return data;
+}
+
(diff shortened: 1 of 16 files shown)


chris-durand:
DMA is working on H7 but I there is an issue with the DAC DMA driver on H7 and, most likely, also on F2, F4 and H7.

On STMG4 the DAC DMA driver only works if the 16 bit samples are written as a 32 bit transfer. When both memory and peripheral sizes are set to "Half-Word" (16 bit) nothing happens.
The stream-channel DMA IP will however force FIFO mode when peripheral and data sizes don't match. Thus, it writes one sample to each half of the data register instead of zero-extending the 16 bit sample to 32 bit as it happens on STM32G4. That results in only every second sample being output by the DAC and the second half of the output is garbled data. 

I will add DAC DMA examples for every STM32 series I have a devboard at home of, test what works and fix the driver.

chris-durand:
DAC DMA is now also working on F4. F3 is not working yet, L4 and F7 are untested.

chris-durand:
@salkinium If the CI succeeds, this is ready to be merged for the release. H7 DMA is working fine and I tested the DAC DMA driver on all platforms listed above. As I don't have the F3 board with me I could not test that one. The [device data fix](https://github.com/modm-io/modm-devices/pull/75) should go first.

examples/nucleo_f446ze/dac_dma/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+	Timer4::setMode(Timer4::Mode::UpCounter);
+	Timer4::setPrescaler(1);
+	Timer4::setOverflow(90 - 1);
+	Timer4::applyAndReset();
+	Timer4::start();
+
+	// Enable trigger out for timer 4
+	TIM4->CR2 |= TIM_CR2_MMS_1;
+
+	//GpioOutputA4::setOutput(true);
```
chris-durand: This line should not be here

salkinium (approved):
Very nice as always! I'll merge it tomorrow afternoon.

PR #780, opened 2021-11-15
[fix] modm::Saturated

TomSaw:
To PR the generic colors nextly from #665 , i would like to PR this fixed `modm::Saturated` first.

### TODO
- [x]  Add missing operator+= / -=
- [x] Extend tests to more types and more cases

PR #781, opened 2021-11-15
[color] Very generic colors: Rgb, Hsv and Gray

TomSaw:
These Colors may be used on their own. F.e. passed to a colored LED driver or being emitted by color-sensor drivers.

They also represent single pixels for `modm::graphic::Buffer` and `modm::graphic::Display` (Work in progress, see [feature/rewrite-graphic](https://github.com/TomSaw/modm/tree/feature/rewrite-graphic)

## Features
- **Bit granularity** for `color::Gray`. Same for `color::Rgb` and `color::Hsv` cause they're made from 3x `color::Gray`
  - Proportional conversion between grays with different digits. 2-bit Gray ---> 4-bit Gray f.e. converts as follows:
      - 0b00 --> 0b0000
      - 0b01 --> 0b0101 // No stupid lshift, the gap is filled up
      - 0b10 --> 0b1010
      - 0b11 --> 0b1111 // Saturated stays saturated, white stays white
- **C++20 concepts** for colortypes and colorgroups (currently _ColorPlane_ and _ColorPalletizing_). These concepts support overall readability and enabled very nice overload resolution modulation for `modm::graphic::Buffer` and `modm::graphic::Display` (Coming soon, see [feature/rewrite-graphic](https://github.com/TomSaw/modm/tree/feature/rewrite-graphic)
- Any imaginable **color conversion** works - type, granularity, palletized or planar... doesn't matter.
- Operators with **saturation arithmetics**

### TODO
- [x] Requires #780 being merged
- [x] Implement some missing operators
- [x] Optional range check for value in `ProportionalUnsigned::ProportionalUnsigned(T value)`
- [x] Update header Timestamps


TomSaw:
This is ready for a first review @salkinium
I also got my fingers on `modm::Saturated` again, improved the API and added another test `SaturationTest::test__uint8_t_ref`.

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+	using T = uint_t<D>::least;
+	static constexpr T min = 0;
+	static constexpr T max = bitmask<D>();
+
+	constexpr ProportionalUnsigned() = default;
+
+	// FIXME want both:
+	// COMPTIME: static_assert(value <= max, "value out of range")
+	// RUNTIME: std::min(value, max) or modm_assert(value <= max, ...)
```
salkinium: Maybe something with `if (std::is_constant_evaluated()) { static_assert } else { modm_assert }`
https://en.cppreference.com/w/cpp/types/is_constant_evaluated

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,114 @@
...
+	constexpr ProportionalUnsigned(T value)
+		: value(std::min(value, max)) {
+		// TODO disable via lbuild option
```
salkinium: Let's not overload lbuild options for something that is rare and can also solved via a defaulted CPP macro that can be overwritten with compile flags.

src/modm/math/saturation/saturated.hpp
```diff
@@ -55,36 +55,63 @@ class Saturated
...
+	template <typename U>
+	constexpr bool
+	operator>=(const U& other) const
+	{ return value >= other; }
+
+	template <typename U>
+	constexpr bool
+	operator<=(const U& other) const
+	{ return value <= other; }
+
```
salkinium: Isn't the spaceship operator supposed to implement these implicitly?

src/modm/math/utils/integer_traits.hpp
```diff
@@ -66,4 +66,19 @@ struct fits_any {
...
+ * @brief  		Generates a bitmask with N bits set
+ * 
+ * @tparam N 	Number of bits to set
+ */
+template<unsigned int N>
+struct bitmask {
+	using value_type = uint_t<N>::least;
+	static constexpr value_type value = std::pow(2, N) - 1;
+	constexpr operator value_type() const noexcept { return value; }
+};
```
salkinium: You could add a shift for stuff like 0b11111000?

src/modm/ui/color/concepts.hpp
```diff
@@ -0,0 +1,95 @@
...
+concept ColorRgbStacked = is_instance<C, RgbStackedD>::value;
+
+template<class C>
+concept ColorHsv = is_instance<C, HsvD>::value;
+
+/**
+ * @brief 	Concept to ident any colortype instance
+ */
+template<class C>
+concept Color = ColorGray<C> || ColorRgb<C> || ColorHsv<C> || ColorRgbStacked<C>; // conjunction
```
salkinium: Can you  still create `class Color` if `concept Color` exists? ie. are the names colliding, or are they implicitly namespaced?
I feel uncomfortable with not being able to distinguish Color concept from Color class just by the name.
On the other hand using `Color_c` to mark the concept (analog to `uint32_t` to mark a type) is ugly?
Perhaps namespace `modm::color::concept::Color` (also a little ugly)?

src/modm/ui/color/gray.hpp
```diff
@@ -0,0 +1,216 @@
...
+ * 				and saturating arithemtics.
+ *
+ * @tparam D 	Number of Digits
+ * @tparam P	When forming a Buffer with this color, this would be the internal storage.
+ * 				Should be MCUs fastest unsigned int -> just keep the default.
+ * 				Some Displays require a specific type therefore it's a template argument.
+ *
+ *
+ * @author		Thomas Sommer
+ * @ingroup		modm_color_gray
```
salkinium:
```suggestion
 * @ingroup		modm_ui_color
```
Unless I missed a `@defgroup modm_color_gray` somewhere (which is not recommended).

src/modm/ui/color/gray.hpp
```diff
@@ -0,0 +1,216 @@
...
+ */
+template <int D>
+class GrayD : public modm::ProportionalUnsigned<D>
+{
+	static_assert(D > 0, "Positive number of digits required for grayscale / colorchannel type.");
```
salkinium: Use unsigned int?

src/modm/ui/color/gray.hpp
```diff
@@ -0,0 +1,216 @@
...
+};
+
+template<typename E>
+using GrayT = GrayD<std::numeric_limits<E>::digits>;
+
+using Monochrome = GrayD<1>;
+using Gray2 = GrayD<2>;
+using Gray4 = GrayD<4>;
+using Gray8 = GrayT<uint8_t>;
+using Gray16 = GrayT<uint16_t>;
```
salkinium, line 318: Very nice!

src/modm/ui/color/hsv.hpp
```diff
@@ -13,115 +9,119 @@
...
+	void iostream_human_friendly(IOStream& io) {
+		using CalcTypeHue = modm::WideType<typename HueType::T>;
+		io << (CalcTypeHue(hue.getValue()) * 360 / hue.max) << "deg\t";
+
+		using CalcTypeSaturation = modm::WideType<typename SaturationType::T>;
+ 		io << (CalcTypeSaturation(saturation.getValue()) * 100 / saturation.max) << "%\t";
+
+		using CalcTypeValue = modm::WideType<typename T::T>;
+		io << (CalcTypeValue(value.getValue()) * 100 / value.max) << "%";
+	}
```
salkinium: Needs to be guarded by `#if __has_include(<modm/io/iostream.hpp>)` to prevent people being forced to include the expensive modm IOStream.

test/modm/math/saturation/saturation_test.hpp
```diff
@@ -19,14 +19,17 @@ class SaturationTest : public unittest::TestSuite
 {
 public:
 	void
-	testSigned8bit();
+	test__int8_t();
```
salkinium:
```suggestion
	test_int8_t();
```
Or is there a reason for the dual underscore?

salkinium (commented):
So far I like it a lot, very nice design with the generic bit width!

src/modm/ui/color/gray.hpp
TomSaw: This wasn't intended. thx

src/modm/ui/color/hsv.hpp
TomSaw: Nice. I've also friended `ostream_human_friendly()` and moved it out of the class, next to `operator<<()` to keep things assorted.

src/modm/ui/color/gray.hpp
TomSaw: Even better: `template <std::unsigned_integral D>`

src/modm/ui/color/concepts.hpp
TomSaw: having a concept "Color" corresponds to the very nature of [concepts intent](https://en.cppreference.com/w/cpp/language/constraints):
> The intent of concepts is to model semantic categories (Number, Range, RegularFunction) rather than syntactic restrictions (HasPlus, Array). According to ISO C++ core guideline T.20, "The ability to specify meaningful semantics is a defining characteristic of a true concept, as opposed to a syntactic constraint."

Concrete Color classes should be specific in how they are a color / how they implement the Color concept, hence will get a specific name like Rgb, Hsv a.s.o. 

**I even hope** that defining a class named abstractly "Color" is permitted ... aaaaaaand.... it is 😊:

> error: 'struct modm::color::Color' redeclared as different kind of entity

src/modm/math/saturation/saturated.hpp
TomSaw: The spaceship operator also rests above this block of handcrafted comparison operators. Unfortunately it only works with the exact same template class as argument:

> error: defaulted member 'constexpr auto modm::Saturated<T>::operator<=>(const T&) const' must have parameter type 'const modm::Saturated<T>&'

src/modm/math/proportional_unsigned.hpp
TomSaw: Ahh this looks good, thanks for the input!

src/modm/math/utils/integer_traits.hpp
TomSaw: You've added such a feature to [Register Configurations](https://modm.io/reference/module/modm-architecture-register/#register-configurations). Here's a clear benefit of encapsulated the shifting information cause it's widely reused.

The `bitmask<..>` instead will likely be always just used to define constants. It's not a class... just a helper and there's no advantage of `bitmask_t<5, 3>` over `bitmask_t<5> << 3` to get `0b11111000`. The later is more intuitive to read i think.

src/modm/ui/color/gray.hpp
TomSaw: Even better: Make use of concepts: I've added `requires (D > 0)` also to the template args of the color classes. Unfortunately [a concept for natural numbers as follows](https://stackoverflow.com/a/63952541) doesn't work:
```
// This is fine ...
template <int D>
concept natural_int = D > 0;

// ... but we can't use it like this
template <natural_int D>

// Instead this works
template <int D>
requires natural_int<D>

// So i've omitted the concept idea and have been repetive with plain requires
template <int D>
requires (D > 0)
```

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,116 @@
...
+	constexpr ProportionalUnsigned() = default;
+
+	constexpr ProportionalUnsigned(T value) {
+		if(std::is_constant_evaluated())
+			this->value = std::min(value, max);
+		else {
+			modm_assert_continue_fail_debug(value <= max, "ProportionalUnsigned", "value out of range");
+			// Constructing runtime should a.f.a.p.
+			this->value = value;
+		}
```
salkinium: I think clamping by default without error is fine here. I'm uncomfortable with only sometimes getting a runtime error.

```cpp
constexpr ProportionalUnsigned(T value) value(std::min(value, max)) {}
```

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	constexpr void setValue(T value) {
+		if(std::is_constant_evaluated())
+			this->value = std::min(value, max);
+		else {
+			modm_assert_continue_fail_debug(value <= max, "ProportionalUnsigned", "value out of range");
+			// Calling setValue() runtime should a.f.a.p.
+			this->value = value;
+		}
```
salkinium: same here

src/modm/math/proportional_unsigned.hpp
TomSaw: In the meanwhile, I think std::clamp is bad and there really should be a switch to enable/disable the assert:

In normal program flow this assert should - by any means -  be omitted: Pixels aggregate 1 to 3 of these ProportionalUnsigned. There may be lot's of ProportionalUnsigned constructions **with well ranged values** inside the `graphic::Buffer` layer and you don't want the std::clamp overhead.

Exceeding the range of ProportionalUnsigned on construction is an issue, throwing "out of range" to the developer is much better than silently clamping bad values.

TomSaw:
Apologize my enduring "push --force" thunderstorm. I'm transitioning from **poke mode** to **test local, push after success mode** 😬

Tests passed @salkinium! What's left is a good solution for the `ProportionalUnsigned` constructor 🙄

TomSaw:
I've added another idea for more power @compile 😍 :

```
// Nothing new
const Rgb565 red = html::Red;
const Rgb565 blue = html::Blue;

// This works because of `modm/ui/color/constexpr_operators.hpp` and is very usefull for UI theming
const auto purple = red + blue;
const auto red_part = purple.getRed();

// You can't mix types with current quick implementation e.g. ColorRgbStacked + ColorRgb
const Rgb565 blub = red + html::Blue; // ERROR
```

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,141 @@
...
+	}
+    return modm::Abandonment::DontCare;
+}
+
+// Register handler for all builds
+MODM_ASSERTION_HANDLER(bad_gray_value);
+// Or register only for debug builds
+// MODM_ASSERTION_HANDLER_DEBUG(log_assertion);
+#endif
+
```
TomSaw: Not my final intend, but does what is required: **assertion code should not go into the release** cause the constructor / assignment operator will be called very, very often in the coming soon `graphic::Buffer`.

I am not getting smarter from reading [modm assertion](https://modm.io/reference/module/modm-architecture-assert/) up and down. Adding a lbuild option? Is there some cool compiler-flag to control assertions?

Please give me some input @salkinium on how to do this right!

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,141 @@
...
+	static constexpr T max = bitmask<D>();
+
+	constexpr ProportionalUnsigned() = default;
+
+	constexpr ProportionalUnsigned(T value) {
+		#ifdef PROPORTIONAL_UNSIGNED_DEBUG
+		if(not modm_assert_continue_ignore(value <= max, "propun.ctor", "value exceeded max", max)) {
+		 	value_ = std::min(max, value);
+			return;
+		}
```
TomSaw: Here's the optional value-assert for developers. #1

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,141 @@
...
+	constexpr operator T() const
+	{ return value_; }
+
+	void operator=(T value) {
+		#ifdef PROPORTIONAL_UNSIGNED_DEBUG
+		if(not modm_assert_continue_ignore(value <= max, "propun.ctor", "value exceeded max", max)) {
+		 	value_ = std::min(max, value);
+			return;
+		}
+		#endif
```
TomSaw: Here's the optional value-assert for developers. #2

src/modm/math/proportional_unsigned.hpp
salkinium: I would use a `modm_assert_continue_ignore_debug` that only triggers an assertion in debug mode and otherwise continues.
I would provide this handler code with `MODM_ASSERTION_HANDLER_DEBUG()` in the documentation instead.
Otherwise if you provide the handler here by default, you cannot disable it.

```cpp
if(not modm_assert_continue_ignore_debug(value <= max, "propun.ctor", "Input value exceeded maximum, getting clamped.", max)) {
    value = max;
}
```

I don't think it's necessary for the release build or to provide a default debug assertion, since you will likely very much see the issue graphically anyways, which will probably cause people to check their input, and then maybe check the documentation and find the debug handler example code.

salkinium: Just be careful about the constexpr, since the `modm_assert` mechnism is obviously not `constexpr` (would be interesting if that can be redirected to a `static_assert` with `std::is_constant_evaluated`, but not sure)!

salkinium: Also put the handler code into the example of course.

TomSaw: Ok, this ...
```
constexpr ProportionalUnsigned(T value) {
  if(not modm_assert_continue_ignore_debug(value <= max, "propun.ctor", "value exceeded max", max)) {
    value_ = std::min(max, value);
    return;
  }
  value_ = value;
}
```
... removes the handler but still asserts and applies `std::min(..)`.
But what i want for release is:
```
constexpr ProportionalUnsigned(T value) {
  value_ = value;
}
```

salkinium: Oh, I see. You can guard this code with `#ifdef MODM_DEBUG_BUILD ` which is set in debug profile. (which apparently is undocumented?!?)

TomSaw: > I would provide this handler code with MODM_ASSERTION_HANDLER_DEBUG() in the documentation instead

Moved the handler to `proportional_unsigned.md`

TomSaw: > Oh, I see. You can guard this code with `#ifdef MODM_DEBUG_BUILD ` which is set in debug profile. (which apparently is undocumented?!?)

👉 Indeed! This is what i've been looking for. And is not documented.. at least not at https://modm.io/reference/module/modm-architecture-assert/ where it belongs to by any means

salkinium: Yup, I'm already fixing it and adding it to `modm/architecture/utils.hpp` and `:build` module too.

TomSaw:
**That's it - Can be merged!**

TomSaw:
What's missing here @salkinium ?

` FileNotFoundError: [Errno 2] No such file or directory: '/__w/modm/modm/src/modm/math/propertional_unsigned.md'`

src/modm/math/proportional_unsigned.lb
```diff
@@ -0,0 +1,26 @@
...
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = ":math:proportional_unsigned"
+    module.description = FileReader("propertional_unsigned.md")
```
salkinium:
```suggestion
    module.description = FileReader("proportional_unsigned.md")
```

TomSaw: Gnaa... this day was too long. Need sleep. Ciao

TomSaw:
Merge please!

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+	template <int E>
+	requires (D <= E)
+	constexpr ProportionalUnsigned(const ProportionalUnsigned<E>& other)
+		: value_(other.value_ >> (E - D)) {}
+
+	template <int E>
+	requires (D <= E)
+	constexpr ProportionalUnsigned(ProportionalUnsigned<E> &&other)
+		: value_(other.value_ >> (E - D)) {}
```
chris-durand: Specialization for rvalue-reference makes no sense here. It does exactly the same as the `const&` version and can be removed. Even if the value was moved from in this function the machine code would be exactly identical to the copy version.

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,120 @@
...
+	requires (D > E)
+	constexpr ProportionalUnsigned(const ProportionalUnsigned<E>& other)
+		: value_(other.value_ * max / other.max)
+	{}
+
+	template <int E>
+	requires (D > E)
+	constexpr ProportionalUnsigned(ProportionalUnsigned<E> &&other)
+		: value_(other.value_ * max / other.max)
+	{}
```
chris-durand: can be removed, same as above

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,120 @@
...
+public:
+	static constexpr int Digits = D;
+
+	using T = uint_t<D>::least;
+	static constexpr T min = 0;
+	static constexpr T max = bitmask<D>();
+
+	constexpr ProportionalUnsigned() = default;
+
+	constexpr ProportionalUnsigned(T value) {
```
chris-durand: I am not really convinced this class is a sane abstraction. The implicit conversions allow for a lot of confusing behaviour, for example:

```
modm::ProportionalUnsigned<2> a = 1;
modm::ProportionalUnsigned<4> b = a;
modm::ProportionalUnsigned<4> c = a + 0;
MODM_LOG_INFO << "b: " << b << ", c: " << c << modm::endl;
```
prints `b: 5, c: 1`

What happens in line 3 is that `a + 0` implicitly converts `a` to an unsigned int type and the `ProportionalUnsigned<4>` is constructed from that without performing the scaling.

The purpose of this class is to associate some special behaviour with an integer value through its type. If you can easily lose that type information, it's very error prone to use.

src/modm/math/proportional_unsigned.md
```diff
@@ -0,0 +1,22 @@
...
+
+```cpp
+static modm::Abandonment
+bad_proportional_unsigned_value(const modm::AssertionInfo &info)
+{
+	if(!strncmp(info.name, (char*)"propun", 6)) {
+    	MODM_LOG_ERROR << info.name << " " << info.description << " " << info.context << modm::endl;;
+    	return modm::Abandonment::Ignore;
+	}
+    return modm::Abandonment::DontCare;
```
chris-durand: Indentation

src/modm/math/utils/integer_traits.hpp
```diff
@@ -66,7 +66,20 @@ struct fits_any {
...
+	using fits_any_t = typename fits_any<Ts...>::type;
 
+/// @ingroup modm_math_utils
+
+/// Generates a bitmask with N bits set
+/// @author		Thomas Sommer
+template<unsigned int N>
+struct bitmask {
+	using value_type = uint_t<N>::least;
+	static constexpr value_type value = std::pow(2, N) - 1;
```
chris-durand: `std::pow(2, N)` will only be `constexpr` in gcc as a non-standard extension and, for example, [not compile with clang](https://gcc.godbolt.org/z/eP1Koa68z). Generally, the C++ standard does not allow most math functions to be `constexpr` because they have to be evaluated according to the runtime floating point environment (rounding mode, etc.). `std::pow` is computed in floating-point and it is unclear if it is always rounded the right way here. I'd do `(1ull << N) - 1`.

src/modm/math/utils/integer_traits.hpp
```diff
@@ -66,7 +66,20 @@ struct fits_any {
...
+	constexpr operator value_type() const noexcept { return value; }
+};
```
chris-durand: What is the advantage of `bitmask` being a type and not a simple free function returning `value_type`? It converts implicitly anyway.

src/modm/math/utils/misc.hpp
```diff
@@ -56,65 +57,53 @@ pow(uint32_t base, uint8_t exponent)
...
+constexpr T vmin(T a, T b, Ts&&... cs)
 {
-	if (a < b)
-		return b;
-	else
-		return a;
+    return a < b ?
+        vmin(a, std::forward<Ts>(cs)...) :
+            vmin(b, std::forward<Ts>(cs)...);
 }
```
chris-durand: There is no advantage here to use `&&` and `std::forward`. Comparing is a non-modifying operation and does not profit from rvalue references.
```
template<typename T, typename... Ts>
constexpr T vmin(T a, T b, Ts... cs)
{
    return a < b ?
        vmin(a, cs...) :
        vmin(b, cs...);
}
```
If this should be efficient with expensive to copy types or work with non-copyable types the values have to be passed by `const&`.

src/modm/math/utils/misc.hpp
```diff
@@ -56,65 +57,53 @@ pow(uint32_t base, uint8_t exponent)
...
+}
+
+template<typename T, typename... Ts>
+constexpr T vmax(T& a, T& b, Ts&... cs)
 {
-	return ( ( (b > c) ? b : c ) > a ) ?
-	         ( (b > c) ? b : c) : a;
+    return a > b ?
+        vmax(a, std::forward<Ts>(cs)...) :
+            vmax(b, std::forward<Ts>(cs)...);
```
chris-durand: Same as above.

src/modm/ui/color/concepts.hpp
```diff
@@ -0,0 +1,101 @@
...
+ * @author		Thomas Sommer
+ * @ingroup		modm_ui_color
+ */
+namespace modm::color {
+
+
+// Forward declarations
+template <int D>
+requires (D > 0)
+class GrayD;
```
chris-durand: What does the `D` stand for in `GrayD`?

src/modm/ui/color/gray.hpp
```diff
@@ -0,0 +1,239 @@
...
+	using T = ProportionalUnsigned<D>::T;
+
+	using modm::ProportionalUnsigned<D>::operator=;
+
+  	template<ColorRgb C>
+	constexpr GrayD(const C& rgb)
+		: ProportionalUnsigned<D>((
+			2125 * (modm::WideType<T>)(GrayD(rgb.red())) +
+			7154 * (modm::WideType<T>)(GrayD(rgb.green())) +
+			0721 * (modm::WideType<T>)(GrayD(rgb.blue()))
```
chris-durand: `0721` is an octal literal and equal to 465. That is probably not intended.

src/modm/ui/color/rgb.hpp
```diff
@@ -24,102 +19,210 @@
...
+		blue_ -= other.blue();
+		return *this;
+	}
+
+	RgbD& operator*=(const RgbD& other) {
+		red_ *= other.red();
+		green_ *= other.green();
+		blue_ *= other.blue();
+		return *this;
+	}
```
chris-durand: Is component-wise multiplication / division of color values a sensible operation? What does this mean?

src/modm/ui/color/concepts.hpp
rleh: RGBD: red, green, blue, depth (output of a depth camera like Realsense/...), so maybe GrayD is "color" of an b/w depth camera? 😉 😉 😜

(I know, D here probably is for _Number of Digits_.)
Why can't we just call it `Gray` instread of `GrayD` (same for `RgbD`/...)? Otherwise people using depth cameras and modm at the same time might get confused...

src/modm/ui/color/rgb.hpp
rleh: It's a cheap ways to darken/lighten a color. Looks useful to me.

chris-durand: > It's a cheap ways to darken/lighten a color. Looks useful to me.

If the data type allowed for fractional values that could work but the components are integer here. Also, semantically multiplying colors makes not much sense and I would model this as a different data type or just allow multiplication with a scalar.

src/modm/ui/color/gray.hpp
```diff
@@ -0,0 +1,239 @@
...
+			) / 10000
```
chris-durand: Could we change the constants in a way that the number we divide by is a power of 2? Then the divide is a left shift. That would be more efficient, especially on Cortex-M0(+) or AVR.

src/modm/ui/color/concepts.hpp
TomSaw: Yes _D_ is for _Digits_
I've attached the prefix in analogy to GrayT<typename T>

```
template<typename E>
using GrayT = GrayD<std::numeric_limits<E>::digits>;

using Monochrome = GrayD<1>;
using Gray2 = GrayD<2>;
using Gray4 = GrayD<4>;
using Gray8 = GrayT<uint8_t>;
using Gray16 = GrayT<uint16_t>;
```
Also many drivers declare local types, so plain `Rgb` is kinda reserved:
`using Rgb = RgbB<16>` or `using Rgb = RgbT<uint16_t>`

**-> How about _B_ for Bits? Would look acceptable: `GrayB<B>`, `RgbB<B>`, `HsvB<B>`**

src/modm/ui/color/gray.hpp
TomSaw: What's the orginal source of this conversion formula from uPainter @salkinium?
For now, i've rounded down: 2125 -> 2124, 721 -> 720

salkinium: I think I got that from [the Wikipedia Grayscale article](https://en.wikipedia.org/wiki/Grayscale#Converting_colour_to_grayscale). The formula seems to be slightly different though: 



A possible adaptation for a bit shift value could be.

- Scaled by 2^13/10000: `(1742 * R + 5859 * G + 591 * B) >> 13`
- Scaled by 2^14/10000: `(3483 * R + 11718 * G + 1183 * B) >> 14`

src/modm/ui/color/concepts.hpp
TomSaw: ... or W for Width: RgbW

src/modm/ui/color/gray.hpp
TomSaw: Calculation takes place in 32 bit anyway.
'Scaled by 2' seems to be the right pick.

TomSaw: I've dump copied and tested your _Scaled by 2^13/10000_ @salkinium . Works like expected, thanks for tinkering this out 👍‍

-> If needed, let's optimize this later in a real application or once the Buffers and Displays got their PR-updates. Will be much easier to confirm the conversion quality if one can generate and inspect interference patterns and such on a screen.

src/modm/math/proportional_unsigned.hpp
TomSaw: Confirm, the implicit conversion of `ProportionalUnsigned` is way too volatile and error prone.

The `ProportionalUnsigned` abstraction is needed for feature completeness of `color::HueD`: It's **hue component is a ProportionalUnsigned** because **Hue should wrap rather than saturate!**

### So i've fixed `ProportionalUnsigned`

- removed the confusing implicit conversion.
- The plain value of `ProportionalUnsigned<...>` now and only is accessible with `ProportionalUnsigned<...>::value()`
- Implemented basic operators for ProportionalUnsigned with wrapping behaviour
- Added a test in `color_test.cpp` to test/demonstrate hue-wrapping of `HsvD<...>`

@chris-durand  's code evaluates correctly now: 
```
...
MODM_LOG_INFO << "b: " << b.value() << ", c: " << c.value() << modm::endl;

```
prints: b: 5, c: 5

src/modm/ui/color/concepts.hpp
TomSaw: Okay... after more concideration, simple feels right -> Have removed the **D** ;)

src/modm/math/proportional_unsigned.hpp
TomSaw: What do you think guys?

TomSaw:
These colors have been proven in the field. Adding code snippets to the docs (`color.md`) rounds up the package.

- We're left open with the [issue of `modm::ProportinoalUnsigned`](https://github.com/modm-io/modm/pull/781#pullrequestreview-891852531). Please have a look on my solution @chris-durand when you find a minute.

chris-durand:
> Please have a look on my solution @chris-durand when you find a minute.

I'll try to find some time tomorrow to look at the PR.

src/modm/math/proportional_unsigned.hpp
chris-durand: Hi, your fix solves that case.

I very much like the idea of having a limited-precision unsigned integer type with wrapping arithmetic, but I am not convinced the proportional conversion should be implemented inside that class via non-explicit conversion constructors.

Have you thought of just having a fixed width wrapping integer class and doing the scaling with a non-member function? The first benefit is it would let the user choose the intended form of conversion (scaling, wrapping, truncating, etc.) making the type more flexible.

Secondly, doing scaling in an implicitly called constructor or assignment could make user code quite hard to read if you don't already know what is going on. If you had something like `scale_to_bits<8>(value)` in the code it would be quite apparent what happens.

Furthermore, separating the conversion gets rid of the conversion constructors where you can't return an error condition from and the irregular assignment operators returning `void`. 

Making the scaling a non-member function follows one of the STL's core principles: separating algorithms and data structures. It makes the class more generic while removing irregularities.

I have to admit I haven't thought this completely through, but it seems a very good idea to me at this point.

src/modm/math/utils/misc.hpp
```diff
@@ -56,74 +57,58 @@ pow(uint32_t base, uint8_t exponent)
...
- * temporary expressions, since they are only evaluated once, unlike a
- * preprocessor macro.
+ * @return   	The biggest object
+ *
+ * @see			https://stackoverflow.com/questions/23815138/implementing-variadic-min-max-functions
  */
 template<typename T>
-constexpr T
-max(const T a, const T b, const T c)
+constexpr T vmax(T& a, T& b)
```
chris-durand:
```suggestion
constexpr T vmax(const T& a, const T& b)
```
The provided implementation does not accept constants, only non-const lvalues.

src/modm/math/utils/misc.hpp
```diff
@@ -56,74 +57,58 @@ pow(uint32_t base, uint8_t exponent)
...
 {
-	return ( ( (b > c) ? b : c ) > a ) ?
-	         ( (b > c) ? b : c) : a;
+    return a > b ? a : b;
+}
+
+template<typename T, typename... Ts>
+constexpr T vmax(T& a, T& b, Ts&... cs)
```
chris-durand:
```suggestion
constexpr T vmax(const T& a, const T& b, const Ts&... cs)
```

src/modm/math/proportional_unsigned.hpp
```diff
@@ -0,0 +1,167 @@
...
+	operator+(U value)
+	{
+		return {T(value_ + value) % (ProportionalUnsigned::max + 1)};
+	}
+
+	template<std::integral U>
+	ProportionalUnsigned
+	operator-(U value)
+	{
+		return {T(value_ - value) % (ProportionalUnsigned::max + 1)};
```
chris-durand: The maximum values are all `2^n - 1`. Therefore, the slow modulus operation can be expressed as a bit-and (`& ProportionalUnsigned::max`). The compiler should perform the optimization anyway but I would write it in terms of `&` to make really sure it works even with `-O0`.

TomSaw: Good point. Thank you!

TomSaw:
Wanna get some stuff off the table. This PR feels robust, if you find a minute @salkinium 😅...

src/modm/ui/color/color.md
```diff
@@ -0,0 +1,90 @@
...
+red = white; // -> Rgb888(255, 255, 255)
+```
+## Saturation Arithmetics
+
+Every arithmetic operations yields in burned out or underexposed colors.
+
+Only the Hue component of Hsv ColorType wraps around just like integers do.
+
+## Flexible Widgets
+- [ ] Complete this snippet and make the code actually working
```
salkinium, line 62: ?

We generally don't promise future code in the documentation, cos we usually don't get it done in time ;-P

salkinium (commented):
I'm fine with these changes, modulo my doc comment.
I'm two weeks from handing in my thesis, so cannot give you a technical review, but if @chris-durand approves this PR, I'll merge it.

chris-durand:
@TomSaw Did you see what I wrote in https://github.com/modm-io/modm/pull/781#discussion_r900488911 ? The discussion got so lengthy that github hides most of it by default :sweat_smile:

TomSaw:
> @TomSaw Did you see what I wrote in https://github.com/modm-io/modm/pull/781#discussion_r900488911 ? The discussion got so lengthy that github hides most of it by default :sweat_smile:

Nope. Totally missed that and check it tomorrow.

src/modm/math/proportional_unsigned.hpp
TomSaw: Yes, its too monolithic. I'll rewrite this!

TomSaw: Solid points @chris-durand.

It's too monolithic! As long as there's no further application, the scaling shouldn't even live in this limited precision type.

I'll rewrite this and put the scaling into color::Gray<..>.

salkinium:
I'm still very interested in merging this, unfortunately, I've completely lost the overview of the review. I think it's almost done, perhaps you want to address the least comments and then we can merge it?

TomSaw:
Hey there. Yes it was almost done. Only chris requested to extract the "arbitrary integer" logic into its own class for good reasons, see https://github.com/modm-io/modm/pull/781#discussion_r900488911 and https://github.com/modm-io/modm/pull/781#discussion_r911941799

Simple task but then, a strong force pulled me into another universe. One with a ton of PHP actually 🤕

-
I've just read the processing::fibers readme and it sounds gorgeous 🤩 congrats for completing this!

TomSaw:
Nothing changed yet.. just cleaned up the tree

PR #782, opened 2021-11-15
[c++20] Rewrite Vector, Location2D, Matrix and LUDecomposition

TomSaw:
This is an excerpt from #665.

`modm::Vector<T, 2>` is passed as template-parameter `Resolution` to `modm::graphic::Buffer<..>`
**-> Thus the constructors require `constexpr`**

By the chance, i may give some maintenance to all of these vector files.
<strike>Just modernization, no changes in logic</strike> C++20 flavoured, more feature complete and 100% backwards compatible.
### Key changes
- Constexpr constructors everywhere
- using `<algorithm>` everywhere (did i miss an oportunity? let me know!)
- Applying `std::round` whenever float->integral conversion occurs
  - `GeometricTraits<T>::round()` could be removed
- `[[depricated]]` all `Vector<>::convert()` in fave of fresh _conversion constructors_ .
- Added some missing routines here and there, f.e IOStream `operator<<`
- Improved readability
  - `using declaratives` for local instances of `Vector<T, N>`
  - (Re)assorted methods

  - Migrated all method definitions into `vector{N}.hpp`

# TODO
- [x] Re-add `Vector<T, 2>::convert()` and `Location2D<U>::convert()` for backwards-comp and depricate it
- [x] Treat deprecations
- [x] `geometric_traits.hpp` avr specialisation 
- [ ] Simplify the Vector tests


salkinium:
Yes please. Constexpr constructors are particularly useful as they reduce to constinit .data section to be copied on startup instead of constructed by instructions.

TomSaw:
`Vector` and `Matrix` use `uint8_t` as template and function parameter type.
Shouldn't this be `std::size_t` or would it be an performance drawback when used as function parameter?

TomSaw:
`Vector` and `Matrix` use `uint8_t` as template and function parameter type for size metrics.
Shouldn't this be `std::size_t` or would it be an performance drawback when used as function parameter?

I think it may be desired to restrict these metrics to 0-255.

chris-durand:
> `Vector` and `Matrix` use `uint8_t` as template and function parameter type for size metrics.
> Shouldn't this be `std::size_t` or would it be an performance drawback when used as function parameter?

That code was written for 8 bit AVR more than ten years ago. On those there is a penalty since arithmetic registers are 8 bit only but `size_t` is 16 bit. `std::size_t` is the right idiomatic choice. I guess `std::size_t` would be fine for `operator[]` on AVR. The compiler will inline it anyway.

TomSaw:
> > `Vector` and `Matrix` use `uint8_t` as template and function parameter type for size metrics.
> > Shouldn't this be `std::size_t` or would it be an performance drawback when used as function parameter?
> 
> That code was written for 8 bit AVR more than ten years ago. On those there is a penalty since arithmetic registers are 8 bit only but `size_t` is 16 bit. `std::size_t` is the right idiomatic choice. I guess `std::size_t` would be fine for `operator[]` on AVR. The compiler will inline it anyway.

Great, so i'll take this to 2020!

TomSaw:
Got some CI claims @windows only but no machine around.. 
Is there a smarter way than setting up a win-VM @salkinium ???

chris-durand:
There are the same errors on Linux. You could be lucky and not need Windows to debug this.

TomSaw:
Okay much better now. There's lot's more i could do but time is rare.

Cause this code is so sensitive to existing projects, **i've been spacious with commits** so you may follow my intends easier. Let's squash everything to the very end.

src/modm/math/geometry/geometric_traits.hpp
```diff
@@ -26,118 +27,59 @@ namespace modm
...
 		static const bool isValidType = true;
 
-		typedef float FloatType;
-
+		using FloatType = float;
 		// Usually the range of a int32_t is big enough so that no
 		// conversion to int64_t is required. This exception is made because
 		// 64-bit operations are very, very slow on an AVR.
-		typedef int32_t WideType;
+		using WideType = int32_t;
```
TomSaw: What's with this 8-bit specialisation? Drop or jinja2 for avr only?

src/modm/math/geometry/geometric_traits.hpp
```diff
@@ -26,118 +27,59 @@ namespace modm
...
 		}
 	};
 
-	template <>
-	struct GeometricTraits<int16_t>
+	template <std::floating_point T>
+	struct GeometricTraits<T>
 	{
+		// DEPRICATED Don't validate with isValidType, use requires modm::arithmetic<T>
 		static const bool isValidType = true;
```
TomSaw, line 52: I've added numerous // DEPRICATED flags...
How do treat these? [C++ depricated](https://en.cppreference.com/w/cpp/language/attributes/deprecated) ?

src/modm/math/geometry/geometric_traits.hpp
```diff
@@ -26,118 +27,59 @@ namespace modm
...
 		}
 	};
 
+	// FIXME only avr
+	// %% if target == "avr"
 	template <>
 	struct GeometricTraits<int32_t>
 	{
+		// DEPRICATED Don't validate with isValidType, use requires std::integral || std::floating_point<T>
 		static const bool isValidType = true;
```
TomSaw: I've added numerous `// DEPRICATED` marks.
How may i treat these? [C++ depricated](https://en.cppreference.com/w/cpp/language/attributes/deprecated) ?

src/modm/math/geometry/location_2d.hpp
```diff
@@ -85,62 +80,29 @@ namespace modm
...
 
-		/// TODO
-		Vector<T, 2>
-		translated(const Vector<T, 2>& vector) const;
-
-		/// Convert between Location-objects with different base-types
-		template <typename U>
-		Location2D<U>
-		convert() const;
-
```
TomSaw, line 104: Here's another occurency of depricated `::convert()`

TomSaw:
This starts looking good. Some changes to notice:
- Depricated ::convert() methods in fave of redefined _conceptuated conversion constructors_ .
- Moved definitions of all constructors from `vector{N}_impl.hpp` to `vector{N}.hpp` cause it's easier to distinct their numerous variants.

-> I think one-liner methods should be moved from `vector{N}_impl.hpp` to `vector{N}.hpp` as well. Confirm @salkinium ?

src/modm/math/geometry/vector1_impl.hpp
```diff
@@ -67,65 +27,17 @@ modm::Vector<T, 1>::operator = (const modm::Matrix<T, 1, 1> &rhs)
...
-}
-
 // ----------------------------------------------------------------------------
 template<typename T>
 const T&
-modm::Vector<T, 1>::operator [] (uint8_t index) const
+modm::Vector<T, 1>::operator [] (std::size_t index) const
 {
 	return reinterpret_cast<const T*>(this)[index];
 }
```
TomSaw: These one-liners overcomplicate the reading.

salkinium:
Yes, makes sense, go for it!

src/modm/math/geometry/vector2.hpp
```diff
@@ -12,373 +13,409 @@
...
+	template<typename U>
+	requires std::integral<T> && std::floating_point<U>
+	void set(U x, U y) {
+		this->x = std::round(x);
+		this->y = std::round(y);
+	}
+
+	// accessors
+	T& operator[] (std::size_t index)
+	{ return reinterpret_cast<T*>(this)[index]; }
```
chris-durand: That code is undefined behaviour. You can actually cast a pointer of a standard-layout type to the pointer type of its first member and dereference it but the array indexing into members is not allowed. I tried many things to exactly keep the class interface the same and make it defined but all these solutions had bad code gen.

A possible way to go could be to store the data as an array and provide member function accessors for x and y that return a reference. Users would need to change `.x` to `.x()` though. We could make basically everything `constexpr` in here if we fixed this. I would not feel comfortable to ship the UB version in production code.

 @salkinium Do you think this change in interface would be acceptable?

TomSaw:
We could go an additional step and destill Vector down to a single generic template class. Some Design thoughts:

### Generic Operator methods
Most operators in current `vector.hpp` use `std::transform` - sometimes `std::accumulate` - to manipulate arrays of T. Compilers optimize aggressively and STL stuff is also predestined to optimize well. Maybe the compiler can even be nailed down to always loop-unroll the operators. ... at least, this would be nice for some profiling.

### Access dims with letters X, Y, Z, W
↥ Note @chris-durand comment above ↥

Data is always stored as `T coords[N]`. The API serves `T& get{X, Y, Z, W}()` depending on `<..., N>` as usuall.

Could be realized <strike>using "recursive inheritance" (if you understand what i mean) or simple `std::enable_if` or</strike> `requires` on hardcoded methods in the generic Vector class.

### Constructor with Parameter Packs
Replace these numerous variations with one variadic template Constructor that takes Parameter Packs of mixed types of Vector<..>

### IMPLEMENT_VECTOR_ACCESSOR{N}

- Not sure with this. Maybe a clever signature being passed const...?

salkinium:
I'm perfectly happy to completely redesign this vector API with breaking changes. It was adapted from very old code that isn't even available anymore, so feel free to change it.

TomSaw:
**Here's a complete rewrite: `modm::Vector` is now an `std::array` with support for arithmetic operations.**

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,552 @@
...
+*
+* \author	Fabian Greif
+* \author	Niklas Hauser
+* \author	Thomas Sommer
+* \ingroup	modm_math_geometry
+*/
+template<class T, std::size_t N>
+requires (N > 0)
+class Vector : public std::array<T, N>
+{
```
TomSaw, line 53: The inheritance on std::array adds lot's of comfort f.e. using `modm::Vector` with [STL ranges](https://en.cppreference.com/w/cpp/ranges)

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,552 @@
...
-	 * Basic data type of all geometric operations. Used to represent vectors
-	 * as well as particular points in the coordinate system.
-	 *
-	 * \section	point_vector	Point vs. vector
+	 * @param args		scalars, Vectors and Arrays in your desired order
+	 */
+	template <class... Args>
+	constexpr Vector(Args... args)
+	{ assign<0>(args...); }
+
```
TomSaw: This constructor replaces the original dozens of variants, satisfies even more cases and is easily extended.

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,552 @@
...
+	// accessors for x, y, z, w, v, u
+	constexpr T x() const { return this->operator[](0); }
+	constexpr T& x() { return this->operator[](0); }
+
+	constexpr T y() const requires (N > 1) { return this->operator[](1); }
+	constexpr T& y() requires (N > 1) { return this->operator[](1); }
+
+	constexpr T z() const requires (N > 2) { return this->operator[](2); }
+	constexpr T& z() requires (N > 2) { return this->operator[](2); }
+
```
TomSaw, line 147: !!! API-change: Vector components x, y, z, w are no more plain properties but methods. It's obligatory for the single-header rewrite, Also note @chris-durand 's _undefined behaviour_ comment above.

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,552 @@
...
+		// std::transform(this->begin(), this->end(), other.begin(), res.begin(), Func{});
+
+		// ... but this optimizes perfectly.
+		res.x() = BinFunc{}(this->x(), other.x());
+		if constexpr(N > 1) res.y() = BinFunc{}(this->y(), other.y());
+		if constexpr(N > 2) res.z() = BinFunc{}(this->z(), other.z());
+		if constexpr(N > 3) res.w() = BinFunc{}(this->w(), other.w());
+		if constexpr(N > 4) res.v() = BinFunc{}(this->v(), other.v());
+		if constexpr(N > 5) res.u() = BinFunc{}(this->u(), other.u());
+
```
TomSaw: For `operator`s returning a new `Vector`, the `<algorithm>`-path did not optimize sufficiently (Line:174) - execution time has tripled. I've found a workaround (Line:177-182) but there's for sure a more elegant solution.

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,552 @@
...
+	constexpr bool hasInf() const { return false; }
+
+	// methods for 2D only
+
+	constexpr Vector
+	operator~ () const
+	requires std::is_signed_v<T> && (N == 2)
+	{ return Vector(y(), -x()); }
+
+	constexpr float getAngle() const requires (N == 2)
```
TomSaw: Some of the 2D methods are constrained to N == 2 instead of generalizing them. Let's improve this on demand, if ever.

test/modm/math/geometry/vector3_test.cpp
```diff
@@ -21,99 +21,97 @@ Vector3Test::testConstructor()
...
-	TEST_ASSERT_EQUALS(q.z, 7);
+	TEST_ASSERT_EQUALS(p.x(), 100);
+	TEST_ASSERT_EQUALS(p.y(), 9);
+	TEST_ASSERT_EQUALS(p.z(), 4);
+
+/* 	modm::Vector3i q(m);
+	TEST_ASSERT_EQUALS(q.x(), -4);
+	TEST_ASSERT_EQUALS(q.y(), 5);
+	TEST_ASSERT_EQUALS(q.z(), 7); */
 }
```
TomSaw, line 115: Re-implementation the Matrix-constructors is a left over TODO.

TomSaw (commented):
**Here's a complete rewrite, maybe a little too radical but feels like natural: `modm::Vector` is now an `std::array` with support for arithmetic operations.**
I've benchmarked against the original code and found the exact same - in some cases even faster - execution times.

chris-durand:
@TomSaw I am quite busy with work right now, but will try to review tomorrow.

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,518 @@
...
+ *  1) strait line: third point behind the second (returns 1)
+ *  2) last point between the other two (returns 0)
+ *  3) third point before the first one (returns -1)
+ *
+ * This definition is useful for inclusion or intersection testing.
+ */
+template<typename T>
+int8_t
+ccw(Vector<T, 2> a, Vector<T, 2> b, Vector<T, 2> c) {
+	using WideType = std::conditional<std::is_floating_point_v<T>, T, modm::WideType<T>>::type;
```
chris-durand: `std::conditional` is not needed here. `modm::WideType<T>` for floating point is `T`.

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,518 @@
...
+ *
+ * If all three points are in a line there are three possibilities:
+ *  1) strait line: third point behind the second (returns 1)
+ *  2) last point between the other two (returns 0)
+ *  3) third point before the first one (returns -1)
+ *
+ * This definition is useful for inclusion or intersection testing.
+ */
+template<typename T>
+int8_t
```
chris-durand:
```suggestion
constexpr int8_t
```

src/modm/math/geometry/line_2d_impl.hpp
```diff
@@ -79,7 +66,7 @@ modm::Line2D<T>::getDistanceTo(const Vector<T, 2>& point) const
 	FloatType d = c1 / c2;
 
 	// calculate the closest point
-	Vector<T, 2> closestPoint = this->point + d * this->directionVector;
+	Vector<T, 2> closestPoint = this->point + this->directionVector * d;
```
chris-durand: More `const` :wink: 
```suggestion
	const Vector<T, 2> closestPoint = this->point + this->directionVector * d;
```

src/modm/math/geometry/line_segment_2d.hpp
```diff
@@ -42,9 +42,11 @@ namespace modm
 		typedef typename GeometricTraits<T>::FloatType FloatType;
 
 	public:
-		LineSegment2D();
+		constexpr LineSegment2D() = default;
```
chris-durand: More function in this class can be `constexpr`, not just the constructors.

src/modm/math/geometry/location_2d.hpp
```diff
@@ -37,39 +36,60 @@ namespace modm
...
+		constexpr Location2D() = default;
+		constexpr Location2D(const Vector<T, 2>& position, const float& orientation)
+			: position(position), orientation(orientation) {}
+
+		[[deprecated("Use 'setPosition({x, y}, orientation)' instead!")]]
+		constexpr Location2D(const T& x, const T& y, const float& orientation)
+			: position(x, y), orientation(orientation) {}
+
+		template<typename U>
+		constexpr Location2D(const Location2D<U> &l) : position(l.position), orientation(l.orientation) {}
```
chris-durand: This enables silent and implicit narrowing conversions. This constructor should be explicit.

src/modm/math/geometry/location_2d.hpp
```diff
@@ -37,39 +36,60 @@ namespace modm
...
-		void
-		setPosition(const Vector<T, 2>& point);
-
-		void
-		setPosition(const T& x, const T& y);
+		Vector<T, 2> position;
+		float orientation = 0;
+
+		constexpr Location2D() = default;
+		constexpr Location2D(const Vector<T, 2>& position, const float& orientation)
```
chris-durand: Small types like `float` should be passed by value. Then it is passed in a register instead of on the stack on many platforms.

src/modm/math/geometry/location_2d.hpp
```diff
@@ -37,39 +36,60 @@ namespace modm
...
+
+		[[deprecated("Use 'setPosition({x, y}, orientation)' instead!")]]
+		constexpr Location2D(const T& x, const T& y, const float& orientation)
+			: position(x, y), orientation(orientation) {}
+
+		template<typename U>
+		constexpr Location2D(const Location2D<U> &l) : position(l.position), orientation(l.orientation) {}
+
+		// getters and setters
+		void setPosition(const Vector<T, 2>& position) { this->position = position; }
```
chris-durand: All the member functions can be `constexpr`.

src/modm/math/geometry/location_2d.hpp
```diff
@@ -37,39 +36,60 @@ namespace modm
...
+
+		Vector<T, 2> getPosition() const { return position; }
+		inline float getOrientation() const { return orientation; }
+		T getX() const { return position.x(); }
+		T getY() const { return position.y(); }
+
+		bool operator== (const Location2D &other) const {
+			return (
+				position == other.position and
+				std::abs(orientation - other.orientation) < __FLT_EPSILON__
```
chris-durand: Can position be float? Should we handle that case somehow?

Not sure if `__FLT_EPSILON__` is the right thing to do. For numbers between 1.0 and 2.0 it's the distance between adjacent floats. For numbers greater than 2.0 the comparison will do the same thing as `==`.

src/modm/math/geometry/location_2d.hpp
```diff
@@ -37,39 +36,60 @@ namespace modm
...
+			);
+		}
+		bool operator!= (const Location2D &other) const {
+			return (
+				position != other.position or
+				std::abs(orientation - other.orientation) > __FLT_EPSILON__
+			);
+		}
```
chris-durand: Couldn't we implement this in terms of `!(*this == other)`?

src/modm/math/matrix.hpp
```diff
@@ -48,33 +47,40 @@ namespace modm
...
 		 *     3, 4,
 		 *     5, 6,
-		 * };
-		 *
-		 * modm::Matrix<int16_t, 3, 2> a(m);
+		 * });
 		 * \endcode
 		 */
-		Matrix(const T *data);
+		constexpr Matrix(const T *data) {
```
chris-durand: That constructor is dangerous, it takes an array as a raw pointer without any size information. I would change this to  take a `std::span<const T, ElementCount>`. Then users have to be explicit when doing something nasty.

src/modm/math/matrix.hpp
```diff
@@ -48,33 +47,40 @@ namespace modm
...
-	template<typename T, uint8_t ROWS, uint8_t COLUMNS>
+	template<typename T, std::size_t ROWS, std::size_t COLUMNS>
 	class Matrix
 	{
 	public:
+		static constexpr std::size_t RowCount = ROWS;
+		static constexpr std::size_t ColumnCount = COLUMNS;
+		static constexpr std::size_t ElementCount = RowCount * ColumnCount;
+
+		T element[ElementCount];
```
chris-durand: The data has to be initialized for the `constexpr` default constructor to be really `constexpr`.
```suggestion
		T element[ElementCount]{};
```

src/modm/math/geometry/vector.hpp
```diff
@@ -11,141 +12,518 @@
...
+		CalcType sum{0};
+
+		for(auto &v: *this)
+			sum += WideType(v) * WideType(v);
+		return sum;
+	}
+
+	template<typename TR = T>
+	constexpr TR getLength() const
+	{ return modm::round_smart<TR, decltype(std::sqrt(getLengthSquared()))>(std::sqrt(getLengthSquared())); }
```
chris-durand: No need for the `decltype`, the second parameter will be deduced from the argument. That is a lot cleaner.
```suggestion
	{ return modm::round_smart<TR>(std::sqrt(getLengthSquared())); }
```

chris-durand (commented):
Thanks for all the effort you put into this. I got about halfway through your changes and will have a more detailed look when I find more time to do so.

TomSaw:
Oops... due to my vector deep dive, i've forgot to mention 🤦‍♂️:
Haven't yet taken the same care for these other geometry classes like i did for vector! This PRs title is kinda misleading.

**If you find another minute, better take a second look on `vector.hpp`** so that the new design is approved.
I'll take your review comments as inspirational boost into a last session with the other geometry candidates.

My pleasure. Also thanks for taking the time to review my code and support my practice! Win-win, i 💙 open-source.

PR #785, opened 2021-11-18
[fix] gdb.py signal handler for linux

### LINKED ISSUE #784: Debug python script cannot handle CTRL-C on Linux
On a linux (Fedora) platform it's not possible to press CTRL-C in GDB to stop execution without GDB terminates completely. I have solved the problem by adding a signal handler to `modm_tools/gdb.py` which does nothing except catching the signal. Does anyone have the same problem, on google it's a known problem then using gdb/openocd, I will be happy to make a PR since it's a annoying feature.

Solution:
``` python
import signal

def signal_handler(sig, frame):
    pass
```
And somewhere in main activate it:
``` python
signal.signal(signal.SIGINT, signal_handler)
```

ghost:
CTRL-C on Linux platform breaks the GDB process and leaves the terminal in a broken state, instead of breaking the debugged program. SIGINT handler added to fix the problem.

### DIFF
--- a/tools/modm_tools/gdb.py
+++ b/tools/modm_tools/gdb.py
@@ -12,9 +12,15 @@
 
 import os
 import subprocess
+import signal
+
+def signal_handler(sig, frame):
+    pass
+
 if __name__ == "__main__":
     import sys
     sys.path.append(os.path.dirname(os.path.dirname(__file__)))
+    signal.signal(signal.SIGINT, signal_handler)
 
 from modm_tools import utils, backend as bem
 from modm_tools import openocd, crashdebug, bmp


salkinium:
I've tested this and it doesn't break anything on macOS, so I'm going to merge this fix. @rleh can you test on Linux?

rleh:
> @rleh can you test on Linux?

I'll check...

chris-durand:
I have also tested it on Ubuntu and it didn't break anything. It is weird that I couldn't reproduce the original bug neither on Arch nor Ubuntu.

chris-durand:
Also when I tried it with `gnome-terminal` it didn't occur, but let's merge it if it does not cause any issues for anyone else.

salkinium (approved):

chris-durand (approved):

Issue #789, opened 2021-12-04
RTT on Blue Pill


StevenMacias:
Hi, I am currently trying to get RTT working on a blue pill. I read this issue https://github.com/modm-io/modm/issues/501 and as far as I understood the STLink-v2 (3v3, GND, SWDIO, SWCLK) should be enough to test this functionality. 

I did the following: 
```
# Copy the rtt example from stm32f3_discovery into the blue_pill_f103 examples
cd modm/examples
cp -r stm32f3_discovery/rtt blue_pill_f103/
# Copy the openocd.cfg since the blue pill doesn't have built-in programmer 
cp blue_pill_f103/blink/openocd.cfg blue_pill_f103/rtt 
```
Then I replace the `LedNorth` by `LedGreen` in the `blue_pill_f103/rtt/main.cpp`

Finally, edit the `blue_pill_f103/rtt/project.xml` by updating the `<extends>` field with the blue pill target and by adding the openocd.cfg option. It looks like this: 

```
<library>
  <extends>modm:blue-pill-f103</extends>
  <options>
    <option name="modm:build:build.path">../../../build/blue_pill_f103/rtt</option>
    <option name="modm:platform:rtt:buffer.rx">16</option>
    <option name="modm:build:openocd.cfg">openocd.cfg</option>
  </options>
  <modules>
    <module>modm:platform:rtt</module>
    <module>modm:processing:timer</module>
    <module>modm:build:scons</module>
    <module>modm:debug</module>
  </modules>
</library>

```
Now, I call: 
- `cd blue_pill_f103/rtt`
- `lbuild build`
- `scons -j8`
- `scons program profile=debug`
- `scons log-rtt`

This is how the final result looks like, no errors but no loop debug traces:
```
(base)  steven@legion  ~/workspace/steven/modm/examples/blue_pill_f103/rtt   develop  scons log-rtt
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
╭───OpenOCD───> Real Time Transfer
╰─────RTT────── stm32f103c8t6
Open On-Chip Debugger 0.11.0
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
WARNING: interface/stlink-v2.cfg is deprecated, please switch to interface/stlink.cfg
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
Info : clock speed 1000 kHz
[...]

Issue #501, opened 2020-11-16
ITM example does not work


strongly-typed:
I am currently not quite sure how to run the example in [examples/nucleo_f103rb/itm](https://github.com/modm-io/modm/tree/develop/examples/nucleo_f103rb/itm).

I slightly modified the example to run on a Blue Pill: [feature/itm](https://github.com/strongly-typed/modm/commit/bbeb364569b7985c12ad80bcdb694c0f05e2482b)

When I run (after lbuild and programming)

```
scons log-itm fcpu=64000000
```

I just get
```
➜  itm git:(feature/itm) ✗ scons log-itm fcpu=64000000
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
.----OpenOCD--> Single Wire Viewer
'------SWO----- stm32f103c8t6
Open On-Chip Debugger 0.10.0+dev-01514-ga8edbd020-dirty (2020-11-16-21:53)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
modm_program
```

and no further output. I tried with an OpenOCD version from January and the latest HEAD.

Any clues?

strongly-typed:
When I change the blink frequency to 5 Hz (delay 100ms), I notice a change of blink frequency when I run `scans log-itm`. Seems to me that the (hard) fault handler caught in. 

Good to have the same blink pattern in the blinky example and the hard fault handler. So whatever happens, the blinky example always works! modm **has** a marketing department!

salkinium:
ITM logs over the SWO pin, which isn't connected on the Blue-Pill?

We should probably implement [RTT logging](https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/), which only requires a debug connection, and works asynchronously.

strongly-typed:
> ITM logs over the SWO pin, which isn't connected on the Blue-Pill?

OK, got it. I only connected `SWCLK` and `SWDIO` to a cheap ST-Link V2 clone. And my expectation was (without reading the doc) that this already bidirectional data connection is also used for logging.

With a cheap ST-Link V2 clone, some [HW mod is required](https://lujji.github.io/blog/stlink-clone-trace/). I'll look into that today.

strongly-typed:
Jupp, works!

```
➜  itm git:(feature/itm) ✗ scons log-itm fcpu=72000000
[...]

PR #791, opened 2021-12-05
Integration of RTT example from stm32f3_discovery to blue_pill_f103

StevenMacias:
As explained in https://github.com/modm-io/modm/issues/789, I wanted to test the RTT functionality in the Blue Pill but currently there is no example for that (also mentioned on https://github.com/modm-io/modm/issues/501#issuecomment-728665271). 

I know it may look redundant since I just needed to change a couple of lines from the original code, but I think this functionality is really attractive for newcomers. Having a plug and play example that shows that you don't need an external UART is a really nice feature. 

CHANGES: 
- main.cpp: changed LedNorth to LedGreen and added myself to the header comment (wanna make my mama proud).
- openocd.cfg: changed stlink-v2.cfg to stlink.cfg to avoid OpenOCD deprecated config warning.
- project.xml: updated accordingly.

### DIFF
--- a/examples/blue_pill_f103/rtt/main.cpp
+++ b/examples/blue_pill_f103/rtt/main.cpp
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2021, Niklas Hauser
+ * Edited by Steven Macias
+ *
+ * This file is part of the modm project.
+ *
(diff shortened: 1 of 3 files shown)


salkinium (approved):
Makes sense!

chris-durand (approved):
Thanks!

PR #795, opened 2021-12-17
[stm32] Fix off-by-one error in exti irq handler

chris-durand:
The lowest interrupt flag bit is not cleared correctly for EXTI interrupts shared between multiple pins. The controller is stuck in the interrupt forever.


salkinium (approved):
🙈

PR #796, opened 2021-12-21
[driver] ADC ADS816x

rleh:
- [x] Driver
- [x] Example
- [x] Test on real hardware

### DIFF
--- a/src/modm/driver/adc/ads816x.hpp
+++ b/src/modm/driver/adc/ads816x.hpp
@@ -1,0 +1,150 @@
+// coding: utf-8
+// ----------------------------------------------------------------------------
+/*
+ * Copyright (c) 2021, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_ADS816X_HPP
+#define MODM_ADS816X_HPP
+
+#include <array>
+#include <modm/architecture/interface/spi_device.hpp>
+#include <modm/architecture/interface/gpio.hpp>
+#include <modm/processing/resumable.hpp>
+#include <modm/processing/timer/timeout.hpp>
+
+namespace modm
+{
+
+/// @ingroup modm_driver_ads816x
+struct ads816x
+{
+	enum class
+	Register : uint16_t
+	{
+		REG_ACCESS		= 0x000,
+		PD_CNTL			= 0x004,
+		SDI_CNTL		= 0x008,
+		SDO_CNTL1		= 0x00C,
+		SDO_CNTL2		= 0x00D,
+		SDO_CNTL3		= 0x00E,
+		SDO_CNTL4		= 0x00F,
+		DATA_CNTL		= 0x010,
+		PARITY_CNTL		= 0x011,
+
+		OFST_CAL		= 0x018,
+		REF_MRG1		= 0x019,
+		REF_MRG2		= 0x01A,
+		REFby2_MRG		= 0x01B,
+
+		AIN_CFG			= 0x024,
+		COM_CFG			= 0x027,
+
+		DEVICE_CFG		= 0x01C,
+		CHANNEL_ID		= 0x01D,
+		SEQ_START		= 0x01E,
+		SEQ_STOP		= 0x01F,
+		ON_THE_FLY_CFG	= 0x02A,
+		AUTO_SEQ_CFG1	= 0x080,
+		AUTO_SEQ_CFG2	= 0x082,
+
+		// Alert, etc...
+	};
+
+	enum class
+	Command : uint8_t
+	{
+		NoOperation		= 0b00000 << 3,
+		Write			= 0b00001 << 3,
+		Read			= 0b00010 << 3,
+		SetBits			= 0b00011 << 3,
+		ClearBits		= 0b00100 << 3,
+	};
+
+	enum class
+	Mode : uint8_t
+	{
+		Manual			= 0b00,
+		OnTheFly		= 0b01,
+		AutoSequence	= 0b10,
+		CustomSequence	= 0b11,
+	};
+
+	enum class
+	DataFormat : uint8_t
+	{
+		OnlyResult			= (0b00 << 4),
+		ResultChannel		= (0b01 << 4),
+		ResultChannelStatus	= (0b10 << 4),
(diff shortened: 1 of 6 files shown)


src/modm/driver/adc/ads816x_impl.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+	// Read conversion results
+	for (buffer2[0] = 0; buffer2[0] < std::popcount(channelsBitmask); buffer2[0]++) {
+		timeout.restart(4us);
+		RF_WAIT_UNTIL(timeout.isExpired());
+
+		RF_WAIT_UNTIL(this->acquireMaster());
+		Cs::reset();
+
+		RF_CALL(SpiMaster::transfer(nullptr, buffer, 2));
```
chris-durand, line 106: If I am reading the datasheet correctly, the first run of the loop will not read the correct sample. According to the sequence diagram the cycle after setting `SEQ_START` will not output the first sample from the sequence but from the input configured before:
image

Data from the first sequence channel is output on the fourth transfer, not the third.

src/modm/driver/adc/ads816x_impl.hpp
```diff
@@ -0,0 +1,134 @@
...
+	if (this->releaseMaster()) {
+		Cs::set();
+	}
+
+	RF_END_RETURN( static_cast<uint16_t>((static_cast<uint16_t>(buffer2[0]) << 8) | buffer2[1]) );
+}
+
+template <typename SpiMaster, typename Cs>
+ResumableResult<void>
+Ads816x<SpiMaster, Cs>::autoSequenceConversion(uint8_t channelsBitmask, std::array<uint16_t, 8>& result)
```
chris-durand: Wouldn't it be better to use `std::span` instead of `std::array&`? There is a choice to be made with span if it should be a fixed size buffer (`std::span<uint16_t, 8>`) or we make it dynamically sized to accomodate for the use-case of sampling less than 8 channels  (`std::span<uint16_t>`). The dynamic size would imply a runtime check for the size.

src/modm/driver/adc/ads816x_impl.hpp
```diff
@@ -0,0 +1,134 @@
...
+	buffer[2] = value;
+
+	RF_WAIT_UNTIL(this->acquireMaster());
+	Cs::reset();
+
+	RF_CALL(SpiMaster::transfer(buffer, nullptr, 3));
+
+	if (command == Command::Read) {
+		Cs::set();
+		timeout.restart(4us);
```
chris-durand: Should we make the period configurable? This would be useful for the chip variants with different sampling speeds.

chris-durand (commented):
Nice!

src/modm/driver/adc/ads816x_impl.hpp
rleh: Correct. I already fixed that behavior in https://github.com/rleh/modm/commit/0eaa41dc82d945e783cbd1d6e073c0533c5b76a8.

But in _auto sequence_ mode the chip still seems to return measurements of the first channel...

src/modm/driver/adc/ads816x_impl.hpp
rleh: Good idea

src/modm/driver/adc/ads816x_impl.hpp
rleh: Done.

rleh:
I commented out the code for the _auto sequence_ mode for now, because it does not seem to switch channels correctly.
Manual mode works on real hardware!
image

Ready to merge from my side.

src/modm/driver/adc/ads816x.hpp
```diff
@@ -0,0 +1,152 @@
...
+	DataFormat : uint8_t
+	{
+		OnlyResult			= (0b00 << 4),
+		ResultChannel		= (0b01 << 4),
+		ResultChannelStatus	= (0b10 << 4),
+		//Reserved			= (0b11 << 4),
+	};
+};
+
+using std::chrono_literals::operator""us;
```
salkinium: nope, this will drag it into the whole modm namespace.

src/modm/driver/adc/ads816x_impl.hpp
```diff
@@ -0,0 +1,155 @@
...
+ResumableResult<void>
+Ads816x<SpiMaster, Cs>::initialize(Mode mode)
+{
+	using namespace std::chrono_literals;
+	RF_BEGIN();
+
+	// Unlock register access by writing magic byte
+	RF_CALL(registerAccess(Command::Write, Register::REG_ACCESS, 0b1010'1010));
+
+    // Power up everything except the internal REFby2 buffer
```
salkinium: indentation police whoop whoop 🚨🚨

salkinium (changes requested):
Good, just some details to polish…

src/modm/driver/adc/ads816x_impl.hpp
rleh: Ouups 😨

salkinium (approved):
Awesome!

src/modm/driver/adc/ads816x_impl.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+	RF_END();
+}
+*/
+
+template <typename SpiMaster, typename Cs>
+modm::ResumableResult<uint8_t>
+Ads816x<SpiMaster, Cs>::registerAccess(Command command, Register reg, uint8_t value)
+{
+	using namespace std::chrono_literals;
```
chris-durand: This line is not required, no user defined literals are used within the function.

src/modm/driver/adc/ads816x_impl.hpp
```diff
@@ -0,0 +1,155 @@
...
+// FIXME: Does not work correctly...
+template <typename SpiMaster, typename Cs>
+template<std::size_t N>
+ResumableResult<void>
+Ads816x<SpiMaster, Cs>::autoSequenceConversion(uint8_t channelsBitmask, std::span<uint16_t, N> result)
+{
+	using namespace std::chrono_literals;
+	RF_BEGIN();
+
+	if (std::popcount(channelsBitmask) > N) {
```
chris-durand, line 85: Just a side-note, if you are ever implementing the auto conversion mode: The most sensible way to use span here would be a `std::span<uint16_t>` with a dynamic size and then have `result.size()` here instead of `N`. That way it would be a normal runtime parameter for the size. The check is at runtime anyway.

chris-durand (approved):
Thanks, I'd be fine with merging it as is.

rleh:
Removed the unnecessary `using namespace std::chrono_literals;` statements.

PR #797, opened 2021-12-29
2021q4 release

salkinium:
It's quarterly release time again! 2022-01-01 is the date.

- [x] Merge remaining PRs?
  - [x] STM32H7 DMA
  - [x] ADS816x
- [x] Update all submodules
- [x] Fix submodule issues
  - [x] FreeRTOS 202112.00
  - [x] ETL v20.22.0
  - [x] TinyUsb v0.12.0
  - [x] LVGL v8.1.0
  - [x] modm-devices
  - [x] sam cmsis headers
  - [x] stm32 cmsis headers
- [x] Update release notes
- [x] Check for updated compilers
- [x] Hardware testing
- ~~Fix author.py script to detect Co-authored-by correctly.~~ next time™

PR #799, opened 2022-01-01
[docs] Add workaround for doxypress crash for mcp7941x driver

chris-durand:
Fix docs generator crash caused by the mcp7941x driver in the Q4 release (see #797)

### DIFF
--- a/src/modm/driver/rtc/mcp7941x.hpp
+++ b/src/modm/driver/rtc/mcp7941x.hpp
@@ -7,7 +7,6 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
- * chip datasheet available at http://ww1.microchip.com/downloads/en/DeviceDoc/22266A.pdf
  */
 // ----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


rleh (approved):

chris-durand:
License header issue fixed as well

rleh (approved):
Thanks!

salkinium (approved):
Thanks!

salkinium:
I'll check if I can teach the docs python script in the build-docs-test job to fail on segfault, that would prevent this issue in the future. So there may be some test commits on this branch.

chris-durand:
Our doxypress config has `"clang-parsing": false`. Seems we are using the broken legacy doxygen lex parser that does not really understand modern C++. Is that on purpose?

salkinium:
> Is that on purpose?

Probably not, I think we took the default and just changed what was needed. If it works without having clang installed, then we can change it.

chris-durand:
`"clang-parsing": true` segfaults for me :scream_cat: 
Let's not mess with it right now ...

salkinium:
Failing properly now: https://github.com/modm-io/modm/runs/4681342161?check_suite_focus=true
And not failing with the fix.

salkinium:
When you have time tomorrow @rleh, you can copy the develop folder to 2021q4 release to archive it.

rleh:
~~Already on my to-do list :)~~
Done ✅

PR #803, opened 2022-01-03
[ci] Explicit specify docker image version

rleh:


rleh:
```
The workflow is not valid. .github/workflows/linux.yml (Line: 12, Col: 14): Unrecognized named-value: 'env'. Located at position 1 within expression: env.IMAGE_TAG .github/workflows/linux.yml (Line: 95, Col: 14): Unrecognized named-value: 'env'. Located at position 1 within expression: env.IMAGE_TAG

```

rleh:
There seems to be no way to specify the container version using some kind of variable, so the string is now in the YAML file 33 times. 😒

Merge?

chris-durand (approved):

salkinium (approved):

PR #805, opened 2022-01-05
[driver] IMU ADIS16470

rleh:
- [x] Driver
  - [x] Add some additional functions for configuration
- [x] Example
- [x] Test on real hardware
  - Everything works
  - ~~Except reading back MSC_CTRL register, see https://github.com/modm-io/modm/pull/805#pullrequestreview-852186650~~ Timing issue solved: MSC_CTRL register requires 3ms waiting before reading back.

### DIFF
--- a/examples/nucleo_f429zi/imu_adis16470/main.cpp
+++ b/examples/nucleo_f429zi/imu_adis16470/main.cpp
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2022, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/platform/spi/spi_master_1.hpp>
+#include <modm/driver/inertial/adis16470.hpp>
+#include <optional>
+
+using namespace Board;
+
+using SpiMaster = SpiMaster1;
+using Mosi = GpioB5;
+using Miso = GpioB4;
+using Sck = GpioB3;
+using Cs = GpioA4;
+using Dr = GpioB9;
+using Rst = GpioB8;
+
+using Adis16470 = modm::Adis16470<SpiMaster, Cs>;
+
+Adis16470 imu{};
+
+int
+main()
+{
+	Board::initialize();
+	Leds::setOutput();
+
+	MODM_LOG_INFO << "ADIS16470 IMU demo on Nucleo-F429ZI\n" << modm::endl;
+
+	SpiMaster::initialize<Board::SystemClock, 650_kHz>();
+	SpiMaster::connect<Sck::Sck, Mosi::Mosi, Miso::Miso>();
+
+	Rst::setOutput();
+	Dr::setInput(Gpio::InputType::PullUp);
+
+	// Reset ADS816x chip
+	Rst::reset();
+	modm::delay(10ms);
+	Rst::set();
+
+	// Wait start-up time
+	modm::delay(252ms);
+
+	MODM_LOG_INFO << "Initializing IMU..." << modm::endl;
+	RF_CALL_BLOCKING(imu.initialize());
+
+	// Software reset
+	RF_CALL_BLOCKING(imu.writeGlobCmd(modm::adis16470::GlobCmd::SoftwareReset));
+	// Wait start-up time after reset
+	modm::delay(252ms);
+
+	// Trigger self test
+	RF_CALL_BLOCKING(imu.writeGlobCmd(modm::adis16470::GlobCmd::SensorSelfTest));
+	// Wait for self test to complete
+	modm::delay(14ms);
+
+	modm::adis16470::DiagStat_t diagStat = RF_CALL_BLOCKING(imu.readDiagStat());
+	MODM_LOG_INFO << "Reading DIAG_STAT register = " << diagStat << modm::endl;
+
+	modm::delay(1ms);
+	auto result = RF_CALL_BLOCKING(imu.readRegister(modm::adis16470::Register::PROD_ID));
+	if (result.has_value()) {
+		MODM_LOG_INFO << "ADIS16470 product id = " << *result << " (should be 16470==0x4056)" << modm::endl;
+	}
+	else {
+		MODM_LOG_INFO << "Unable to reag PROD_ID register" << modm::endl;
+	}
+
+	modm::adis16470::MscCtrl_t mscCtrl = RF_CALL_BLOCKING(imu.readMscCtrl());
+	MODM_LOG_INFO << "Reading MSC_CTRL register = " << mscCtrl << modm::endl;
+
+	RF_CALL_BLOCKING(imu.setDataOutputFrequency<10_Hz>());
+
+	RF_CALL_BLOCKING(imu.writeRegister(modm::adis16470::Register::FILT_CTRL, 8)); // N_B = 2^8 = 64
+
+	mscCtrl = modm::adis16470::MscCtrl::DrPolarity | modm::adis16470::MscCtrl::PointOfPercussionAlign;
+	MODM_LOG_INFO << "Writing MSC_CTRL register = " << mscCtrl << modm::endl;
+	RF_CALL_BLOCKING(imu.writeMscCtrl(mscCtrl));
+
+	diagStat = RF_CALL_BLOCKING(imu.readDiagStat());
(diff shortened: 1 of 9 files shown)


rleh:
Chip seems to work (`DR` pin toggles as expected), but `DOUT/MISO` signal is constantly low.
image
Looks like it might be shorted to `CS`, but my multimeter measures approx. 20kOhm from `CS` to `DOUT`.

image

salkinium:
Maybe the MISO needs a pullup?

rleh:
> Maybe the MISO needs a pullup?

I already [enabled the STM32 internal pull-up](https://github.com/modm-io/modm/pull/805/files#diff-b4b02cfce2933b74ab36ecaa929a4f501bf17054ce00db68e63c6cd985330389R40), but maybe it is to weak for 600kHz SPI...

salkinium:
For F4 it's only 30-50kOhm for the pullup (see datasheet, I/O characteristics).

rleh:
I added an external 3k9 pull-up to MISO and still same issue.
Maybe a hardware defect of the IMU (costs 300$)... 😒
I will test the driver with another identical IMUs (but already soldered on custom PCBs).

chris-durand:
> > Maybe the MISO needs a pullup?
> 
> I already [enabled the STM32 internal pull-up](https://github.com/modm-io/modm/pull/805/files#diff-b4b02cfce2933b74ab36ecaa929a4f501bf17054ce00db68e63c6cd985330389R40), but maybe it is to weak for 600kHz SPI...

~~I guess it is. Rise time is 2.2 * R *C. With the pull up R=40kOhm and C=5pF for the STM32 IO we get 2.8 µS. The reciprocal is 362 kHz.~~

EDIT: Ah, the calculation is wrong, I made a typo with the calculator but it's not the issue anyways :man_facepalming: .

salkinium:
Uiuiui, the gloves are off now, at your pencils, ready, math!

src/modm/driver/inertial/adis16470.hpp
```diff
@@ -0,0 +1,233 @@
...
+	 * Improves SPI usage efficiency by 100% compared to `accessRegister(...)`.
+	 *
+	 * @param sequence All registers in the std::span will be read and the
+	 * results will be stored in values.
+	 * @param values Results of the read operations
+	 * @return False in case of any error, e.g. if some register access is not
+	 * permitted.
+	 */
+	modm::ResumableResult<bool>
+	readRegisterSequence(std::span<Register> sequence, std::span<uint16_t> values);
```
chris-durand:
```suggestion
	readRegisterSequence(std::span<const Register> sequence, std::span<uint16_t> values);
```
The register list should be const.

src/modm/driver/inertial/adis16470_impl.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+	for (i = 0; i < sequence.size(); i++) {
+		if (getRegisterAccess(sequence[i]) == AccessMethod::Write) {
+			// Reading this register is not permitted
+			RF_RETURN(false);
+		}
+
+		buffer[0] = uint8_t(sequence[i]) & 0b0111'1111;
+		buffer[1] = 0;
+		RF_CALL(SpiMaster::transfer(buffer.data(), &buffer[2], 2));
```
chris-durand, line 207: Chip select is never reset inside of `readRegisterSequence()`.

examples/nucleo_f429zi/imu_adis16470/main.cpp
```diff
@@ -0,0 +1,103 @@
...
+main()
+{
+	Board::initialize();
+	Leds::setOutput();
+
+	MODM_LOG_INFO << "ADIS16470 IMU demo on Nucleo-F429ZI\n" << modm::endl;
+
+	SpiMaster::initialize<Board::SystemClock, 650_kHz>();
+	SpiMaster::connect<Sck::Sck, Mosi::Mosi, Miso::Miso>();
+	Miso::setInput(Gpio::InputType::PullUp);
```
chris-durand: Doesn't this reset the pin mode from SPI alternate function to plain input?

chris-durand:
The IMU requires SPI mode 3:

spi

This is never configured in the driver.

chris-durand:
The data out should not require a pull-up. I would remove it.

examples/nucleo_f429zi/imu_adis16470/main.cpp
```diff
@@ -0,0 +1,128 @@
...
+	// 		  modm::adis16470::MscCtrl::LinearGCompensationGyro |
+	// 		  modm::adis16470::MscCtrl::PointOfPercussionAlign;
+	MODM_LOG_INFO << "Writing MSC_CTRL register = " << mscCtrl << modm::endl;
+	RF_CALL_BLOCKING(imu.writeMscCtrl(mscCtrl));
+
+	diagStat = RF_CALL_BLOCKING(imu.readDiagStat());
+	MODM_LOG_INFO << "Reading DIAG_STAT register = " << diagStat << modm::endl;
+
+	mscCtrl = RF_CALL_BLOCKING(imu.readMscCtrl());
+	MODM_LOG_INFO << "Reading MSC_CTRL register = " << mscCtrl << modm::endl;
```
rleh: Somehow the write to MSC_CTRL is working, but also no error shows up in DIAG_STAT.

image

chris-durand: The write looks fine. Have you tried to write and read back any other register the same way?

Could it be a timing issue? The wait times for the reset are listed as "typical" values.

rleh: > Have you tried to write and read back any other register the same way?

Not yet.

> Could it be a timing issue? The wait times for the reset are listed as "typical" values.

Could be a timing issue, but the ADIS16470 seems to be ready and producing data even ~50ms earlier.
I'll add more delay.

rleh:
> The IMU requires SPI mode 3: [image ...]
> The data out should not require a pull-up. I would remove it.

Works now with SPI mode 3 and without any pull-up.

src/modm/driver/inertial/adis16470.hpp
```diff
@@ -0,0 +1,332 @@
...
+	std::array<uint8_t, bufferSize> buffer;
+	std::size_t i;
+	uint16_t checksum;
+	std::optional<uint16_t> tmp;
+
+	const modm::ShortPreciseDuration tStall{std::chrono::microseconds(16)};
+	modm::ShortPreciseTimeout timeout{tStall};
+};
+
+// Output operators
```
salkinium: This should be guarded, so that it can be compiled without the `modm:io` module

src/modm/driver/inertial/adis16470.lb
```diff
@@ -0,0 +1,36 @@
...
+        ":math:utils",
+        ":processing:resumable",
+        ":processing:timer")
+    return True
+
+def build(env):
+    env.outbasepath = "modm/src/modm/driver/inertial"
+    env.copy("adis16470.hpp")
+    env.copy("adis16470_impl.hpp")
+    env.copy("adis16470.cpp")
```
salkinium: Only if with `modm:io` module

src/modm/driver/inertial/adis16470.hpp
```diff
@@ -0,0 +1,332 @@
...
+ * \ingroup	modm_driver_adis16470
+ * \author	Raphael Lehmann
+ */
+template<class SpiMaster, class Cs>
+class Adis16470 : public adis16470,
+				  public modm::SpiDevice<SpiMaster>,
+				  protected modm::NestedResumable<2>
+{
+public:
+	Adis16470() = default;
```
salkinium: Is this not implicitly so for all classes?

salkinium (changes requested):
Nice driver, just needs to guard the IO module.

salkinium (approved):
Squash?

rleh:
> Squash?

Yes, but after investigating why the MSC_CTRL register does not read back correctly.

examples/nucleo_f429zi/imu_adis16470/main.cpp
rleh: > The write looks fine. Have you tried to write and read back any other register the same way?

Other registers (e.g. `USER_SCR1`) read back fine.

> Could it be a timing issue? The wait times for the reset are listed as "typical" values.

Yes, waiting 3ms solves the issue and `MSC_CTRL` reads back correctly.
I added a 3ms timeout after writing `MSC_CTRL` in the `writeMscCtrl(MscCtrl_t v)` function https://github.com/modm-io/modm/blob/c0adf493e7f432b962dc9388bb3b5eeac9aa8d63/src/modm/driver/inertial/adis16470_impl.hpp#L144-L147

rleh:
Squashed, tested (again) on real hardware & ready to merge!

salkinium (approved):

chris-durand (approved):
Thanks! Very nice!

PR #807, opened 2022-01-06
[scons] Emit explicit error when chosen gcc isn't found on PATH

WasabiFan:
Previously, if the compiler was not on PATH, shutil.which would return
None and the error from "realpath" would be:

    TypeError: expected str, bytes or os.PathLike object, not NoneType

Someone else should probably test this. I made these edits via copy-paste from elsewhere and haven't run this particular commit (but _have_ confirmed it worked in a separate working copy).

### DIFF
--- a/tools/build_script_generator/scons/resources/SConscript.in
+++ b/tools/build_script_generator/scons/resources/SConscript.in
@@ -37,7 +37,15 @@ env.Append(toolpath=[
 %% for tool in tools | sort
 env.Tool("{{tool}}")
 %% endfor
-env["GCC_PATH"] = dirname(dirname(realpath(shutil.which(env["CC"]))))
+
+c_compiler_name = env["CC"]
+c_compiler_path = shutil.which(c_compiler_name)
+
+if c_compiler_path is None:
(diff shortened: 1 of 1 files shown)


rleh (approved):
Thanks!

chris-durand (approved):
Thanks, working:

```
$ lbuild build; scons
Selected compiler "arm-none-eabi-gcc" not found on PATH. Please add its installation directory to the PATH environment variable.
```

salkinium (approved):

salkinium:
Ah, I merged the wrong PR first, could you rebase this on develop again? (I don't have push access to your remote)

WasabiFan:
I'm not at my PC right now, but it looks like GH was able to merge from the target branch so we should be good. Probably best to squash-merge to avoid the unnecessary merge commit.

salkinium:
Eh, my idiot brain just realized I can just manually merge it…

WasabiFan:
For future reference, commit is: https://github.com/modm-io/modm/commit/78308b0d055049e7fca6e3a9576f32e21b41e17f

Thanks!

PR #810, opened 2022-01-12
[example] Add blue_pill_f103/undefined_irq

TomSaw:
Let's add this expressive example to this mainstream board!


salkinium:
Alternatively you could move it to examples/generic and make it work on more STM32 devices. I don't really like copying examples like this verbatim.

PR #811, opened 2022-01-12
[fix] fixed forgotten modm:blue-pill -> modm:blue-pill-f103

TomSaw:
This has grabbed my attention on CI builds so quickly fixed it.

### DIFF
--- a/examples/generic/ros/can_bridge/project.xml
+++ b/examples/generic/ros/can_bridge/project.xml
@@ -1,5 +1,5 @@
 <library>
-  <extends>modm:blue-pill</extends>
+  <extends>modm:blue-pill-f103</extends>
   <options>
     <option name="modm:build:build.path">../../../../build/generic/ros/can_bridge</option>
     <option name="modm:build:openocd.cfg">openocd.cfg</option>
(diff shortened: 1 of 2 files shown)


salkinium (approved):
Thanks, I should check for warnings like these more often!

rleh (approved):

salkinium:
What's with the commit message? That doesn't sound related to the lbuild alias warning?

```
[fix] fixed forgotten modm:blue-pill -> modm:blue-pill-f103

Previously, if the compiler was not on PATH, shutil.which would return
None and the error from "realpath" would be:

    TypeError: expected str, bytes or os.PathLike object, not NoneType
```

chris-durand:
> What's with the commit message? That doesn't sound related to the lbuild alias warning?

@TomSaw Also, the commit author should probably be you and not @WasabiFan. I guess you accidentally amended [that commit](https://github.com/modm-io/modm/commit/78308b0d055049e7fca6e3a9576f32e21b41e17f).

TomSaw:
OK is fixed - Thanks guys!

chris-durand (approved):
Thanks!

PR #816, opened 2022-01-28
Issues when using TinyUSB and FreeRTOS together

rleh:
I guess this is a bug in our TinyUSB port or a version incompatability between our FreeRTOS and the FreeRTOS version expected by TinyUSB.

But could also be a bug with local setup, let's see if the example fails in CI...

```
$ cd examples/nucleo_f429zi/usb_freetros/
$ lbuild build
$ scons
scons: Building targets ...
Compiling C++·· release/main.o
In file included from modm/ext/tinyusb/osal/osal.h:53,
                 from modm/ext/tinyusb/tusb.h:38,
                 from main.cpp:12:
modm/ext/tinyusb/osal/osal_freertos.h: In function 'QueueDefinition* osal_semaphore_create(osal_semaphore_def_t*)':
modm/ext/tinyusb/osal/osal_freertos.h:56:10: error: 'xSemaphoreCreateBinaryStatic' was not declared in this scope; did you mean 'xSemaphoreCreateBinary'?
   56 |   return xSemaphoreCreateBinaryStatic(semdef);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |          xSemaphoreCreateBinary
modm/ext/tinyusb/osal/osal_freertos.h: In function 'QueueDefinition* osal_mutex_create(osal_mutex_def_t*)':
modm/ext/tinyusb/osal/osal_freertos.h:99:10: error: 'xSemaphoreCreateMutexStatic' was not declared in this scope; did you mean 'xSemaphoreCreateMutex'?
   99 |   return xSemaphoreCreateMutexStatic(mdef);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |          xSemaphoreCreateMutex
modm/ext/tinyusb/osal/osal_freertos.h: In function 'QueueDefinition* osal_queue_create(osal_queue_def_t*)':
modm/ext/tinyusb/osal/osal_freertos.h:134:10: error: 'xQueueCreateStatic' was not declared in this scope; did you mean 'xQueueCreateSet'?
  134 |   return xQueueCreateStatic(qdef->depth, qdef->item_sz, (uint8_t*) qdef->buf, &qdef->sq);
      |          ^~~~~~~~~~~~~~~~~~
      |          xQueueCreateSet
scons: *** [.../build/nucleo_f429zi/usb_freetros/scons-release/main.o] Error 1
```

cc @salkinium @chris-durand

### DIFF
--- a/ext/hathach/module.lb
+++ b/ext/hathach/module.lb
@@ -20,11 +20,12 @@ def init(module):
     module.description = FileReader("module.md")
 
 def prepare(module, options):
-    if not (options[":target"].has_driver("usb") or
-            options[":target"].has_driver("usb_otg_fs") or
-            options[":target"].has_driver("usb_otg_hs") or
-            options[":target"].has_driver("udp") or
-            options[":target"].has_driver("uhp")):
+    device = options[":target"]
+    if not (device.has_driver("usb") or
+            device.has_driver("usb_otg_fs") or
+            device.has_driver("usb_otg_hs") or
+            device.has_driver("udp") or
+            device.has_driver("uhp")):
         return False
 
     configs = {"device.cdc", "device.msc", "device.vendor", "device.midi", "device.dfu"}
@@ -42,12 +43,15 @@ def prepare(module, options):
                               dependencies=lambda vs:
                                     [":tinyusb:{}".format(v.replace(".", ":")) for v in vs]))
 
-    if options[":target"].has_driver("usb_otg_hs"):
+    # Most devices only have FS=usb, some STM32 have FS/HS=usb_otg_fs/usb_otg_hs
+    # and some STM32 only have HS=usb_otg_hs, so we only define this option if
+    # there is a choice and default this to FS by default.
+    if device.has_driver("usb_otg_hs") and device.has_driver("usb_otg_fs"):
         module.add_option(
                 EnumerationOption(name="speed",
                                   description="USB Port Speed",
                                   enumeration={"full": "fs", "high": "hs"},
-                                  default="high" if options[":target"].has_driver("usb_otg_hs") else "full",
+                                  default="full",
                                   dependencies=lambda s: ":platform:usb:{}s".format(s[0])))
 
     module.add_submodule(TinyUsbDeviceModule())
@@ -121,7 +125,9 @@ def build(env):
 
     has_device = env.has_module(":tinyusb:device")
     has_host = env.has_module(":tinyusb:host")
-    speed = env.get("speed", "fs")
+    # On STM32 with both speeds, the option decides, otherwise we must default
+    # to HS for the few STM32 with *only* usb_otg_hs, all other devices are FS.
+    speed = env.get("speed", "hs" if env[":target"].has_driver("usb_otg_hs") else "fs")
     port = 0 if speed == "fs" else 1
 
     mode = None
--- a/ext/hathach/module.lb
+++ b/ext/hathach/module.lb
@@ -113,11 +113,16 @@ def build(env):
             tusb_config["CFG_TUSB_MCU"] = "OPT_MCU_RP2040"
             env.copy("tinyusb/src/portable/raspberrypi/rp2040/", "portable/raspberrypi/rp2040/")
 
+    # TinyUSB has a OS abstraction layer for FreeRTOS, but it is causes problems with modm
+    # if env.has_module(":freertos"):
+    #    tusb_config["CFG_TUSB_OS"] = "OPT_OS_FREERTOS"
+    #    env.copy("tinyusb/src/osal/osal_freertos.h", "osal/osal_freertos.h")
+
     if env.has_module(":freertos"):
-        tusb_config["CFG_TUSB_OS"] = "OPT_OS_FREERTOS"
-        env.copy("tinyusb/src/osal/osal_freertos.h", "osal/osal_freertos.h")
-    else:
-        env.copy("tinyusb/src/osal/osal_none.h", "osal/osal_none.h")
+        env.log.warning("TinyUSB in modm does not currently use the FreeRTOS abstraction layer"
+                        " and is not thread-safe with FreeRTOS threads.")
+
+    env.copy("tinyusb/src/osal/osal_none.h", "osal/osal_none.h")
 
     if env.has_module(":debug"):
         tusb_config["CFG_TUSB_DEBUG_PRINTF"] = "tinyusb_debug_printf"
--- a/examples/nucleo_f429zi/usb_freertos/main.cpp
+++ b/examples/nucleo_f429zi/usb_freertos/main.cpp
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2022, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <tusb.h>
+
+#include <modm/board.hpp>
+#include <modm/processing/rtos.hpp>
+#include <modm/processing/timer/periodic_timer.hpp>
+
+using namespace Board;
+
+class SomeTask : modm::rtos::Thread
+{
+public:
+	SomeTask() : Thread(2, 2048, "some_task") {}
+
+	void
+	run()
+	{
+		modm::PeriodicTimer tmr{2.5s};
+		uint8_t counter{0};
+
+		while (1)
+		{
+			if (tmr.execute())
+			{
+				MODM_LOG_INFO << "SomeTask: Hello via Uart, counter=" << (counter++) << modm::endl;
+			}
+
+			vTaskDelay(1);
+		}
+
+		vTaskDelete(0);
+	}
+};
+
+class UsbTask : modm::rtos::Thread
+{
+public:
+	UsbTask() : Thread(1, 2048, "usb_task"), usb_io_device0{}, usb_stream0{usb_io_device0} {}
+
+	void
+	run()
+	{
+		MODM_LOG_INFO << "USBTask: Calling Board::initializeUsbFs() ..." << modm::endl;
+		Board::initializeUsbFs(4);
+		MODM_LOG_INFO << "USBTask: Calling tusb_init() ..." << modm::endl;
+		tusb_init();
+		MODM_LOG_INFO << "... done!" << modm::endl;
+
+		modm::PeriodicTimer tmr{2.5s};
+		uint8_t counter{0};
+
+		while (1)
+		{
+			tud_task();
+			if (tmr.execute())
+			{
+				MODM_LOG_INFO << "UsbTask: Hello via Uart, counter=" << counter << modm::endl;
+				usb_stream0 << "UsbTask: Hello via USB CDC, counter=" << counter << modm::endl;
+				counter++;
+			}
+
+			vTaskDelay(1);
+		}
+
+		vTaskDelete(0);
+	}
+
+private:
+	modm::IODeviceWrapper<UsbUart0, modm::IOBuffer::DiscardIfFull> usb_io_device0;
+	modm::IOStream usb_stream0;
+};
+
+SomeTask someTask;
+UsbTask usbTask;
+
+int
+main()
+{
+	Board::initialize();
+	Leds::setOutput();
+	MODM_LOG_INFO << "Nucleo-F429ZI: TinyUSB & FreeRTOS example" << modm::endl;
+
+	MODM_LOG_INFO << "WARNING!" << modm::endl;
+	MODM_LOG_INFO << "TinyUSB in modm does not currently use the FreeRTOS abstraction layer";
+	MODM_LOG_INFO << " and is not thread-safe with FreeRTOS threads." << modm::endl;
+
+	modm::rtos::Scheduler::schedule();
+
+	// we should never get here
+	return 0;
+}
--- a/examples/nucleo_f429zi/usb_freertos/project.xml
+++ b/examples/nucleo_f429zi/usb_freertos/project.xml
@@ -1,0 +1,17 @@
+<library>
+  <extends>modm:nucleo-f429zi</extends>
+  <options>
+    <option name="modm:build:build.path">../../../build/nucleo_f429zi/usb_freertos</option>
+    <option name="modm:tinyusb:config">device.cdc</option>
+  </options>
+  <modules>
+    <module>modm:build:scons</module>
+    <module>modm:freertos</module>
+    <module>modm:processing:rtos</module>
+    <module>modm:processing:timer</module>
+    <module>modm:tinyusb</module>
+  </modules>
+  <collectors>
+    <collect name="modm:build:cppdefines">CFG_TUSB_DEBUG=2</collect>
+  </collectors>
+</library>
--- a/ext/aws/FreeRTOSConfig.h.in
+++ b/ext/aws/FreeRTOSConfig.h.in
@@ -48,7 +48,7 @@ your application. */
 #define configCPU_CLOCK_HZ                      ( SystemCoreClock )
 #define configTICK_RATE_HZ                      ( {{ frequency }} )
 /* Memory allocation related definitions. */
-#define configSUPPORT_STATIC_ALLOCATION         {{ 0 if with_heap else 1 }}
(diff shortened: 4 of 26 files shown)


rleh:
> `xSemaphoreCreateBinaryStatic()`

"configSUPPORT_STATIC_ALLOCATION must be set to 1 in FreeRTOSConfig.h for this RTOS API function to be available." (https://www.freertos.org/xSemaphoreCreateBinaryStatic.html)

salkinium:
This is an issue with out FreeRTOS port that added the ability to use static memory in #750.
It needs to be adapted to allow for both `configSUPPORT_STATIC_ALLOCATION` and `configSUPPORT_DYNAMIC_ALLOCATION` at the same time, currently it's one of the other depending on whether the `modm:platform:heap` module is available, which is not a good choice…

salkinium:
I guess a better implemention would be to enable `configSUPPORT_STATIC_ALLOCATION` always, and only enable `configSUPPORT_DYNAMIC_ALLOCATION` if `modm:platform:heap` is available. That would only be a small change iff FreeRTOS supports both.

salkinium:
Yeah, this seems to be the right way: https://forums.freertos.org/t/static-and-dynamic-allocation-together/8138/3

> Yes it is possible to use both static and dynamic allocation at the same
> time - many of the demos do this and it is a very common scenario in
> production (especially where libraries are using static allocation even
> if application code is not).
>
> However, if you have configSUPPORT_STATIC_ALLOCATION set to 1 then the
> timer/daemon and idle tasks WILL use static allocation and you MUST
> provide the statically allocated memory via callbacks. See the
> following documentation to see how:
> https://www.freertos.org/a00110.html#configSUPPORT_STATIC_ALLOCATION

salkinium:
That's not enough you need to [change this `%% else` to `%% endif`](https://github.com/modm-io/modm/blob/develop/ext/aws/modm_port.cpp.in#L68) to always provide the required static memory. Also [this here](https://github.com/modm-io/modm/blob/develop/ext/aws/FreeRTOSConfig.h.in#L51).

rleh:
Already done, give me 15 seconds to write a commit message (without typos).

salkinium:
Please also update the line in `freertos.md`.

rleh:
Seems to work now with FreeRTOS, but TinyUSB does not work on my Nucleo-F429ZI board. I checked the [`generic/usb/`](https://github.com/modm-io/modm/tree/develop/examples/generic/usb) example: It does not work either.
The call to `tusb_init()` never returns.

@salkinium (On which hardware) have verified TinyUSB after the last update?

salkinium:
Black Pill and F103 😬. Perhaps an issue with [the VBUS sense config](https://github.com/modm-io/modm/blob/develop/src/modm/board/nucleo_f429zi/board.hpp#L160)?

salkinium:
The [TinyUSB BSP for the F439 has the same though](https://github.com/hathach/tinyusb/blob/master/hw/bsp/stm32f4/boards/stm32f439nucleo/board.h#L97)

rleh:
TinyUSB hangs here: https://github.com/modm-ext/tinyusb-partial/blob/6a3450ec4ba9598136c1df345e1eaa1c62cc0f52/src/portable/st/synopsys/dcd_synopsys.c#L504
because Nucleo-F429ZI has USB FS and HS peripherals, modm defaults to the HS peripheral and the example does not set the speed option.

Buit with this fixed I run into an assertion once I plug in the USB to a computer...

rleh:
Another bug: `CFG_TUSB_DEBUG=2` seems to require heap.

salkinium:
>  modm defaults to the HS peripheral

Hmm, [this line seems to be a copy-paste error](https://github.com/modm-io/modm/blob/develop/ext/hathach/module.lb#L50), it should just be `"full"`.

rleh:
So default to `full` for all devices?

chris-durand:
> TinyUSB hangs here: https://github.com/modm-ext/tinyusb-partial/blob/6a3450ec4ba9598136c1df345e1eaa1c62cc0f52/src/portable/st/synopsys/dcd_synopsys.c#L504 because Nucleo-F429ZI has USB FS and HS peripherals, modm defaults to the HS peripheral and the example does not set the speed option.
> 
> Buit with this fixed I run into an assertion once I plug in the USB to a computer...

Just wanted to post that I found it, but you were 8 min earlier :smiley:

salkinium:
> So default to full for all devices?

Yes, otherwise it breaks default behavior of devices without HS peripheral that don't have this option. Very nasty debugging.

chris-durand:
> So default to `full` for all devices?

It is a reasonable default and the most commonly used option. High speed USB requires an external PHY on almost all STM32. Some F730 have an internal high speed PHY but it uses completely different pins. On those users will be aware that they have to make the choice according to the hardware.

rleh:
More bugs while trying to debug:

```
$ scons debug-coredump firmware=9AFFA6659846AE10021C2E4443054683898E8D9A
scons: Building targets ...
scons: *** [gdb_post_mortem_debug] Source `artifacts/9affa6659846ae10021c2e4443054683898e8d9a.elf' not found, needed by target `gdb_post_mortem_debug'.
```

and

```
$ scons profile=debug -j8 artifact
scons: Building targets ...
╭───Artifact─── build/usb/scons-debug/usb.elf
╰────Cache────> artifacts/None.elf
scons: *** [store_artifact] ValueError : Unable to find Build ID for 'build/usb/scons-debug/usb.elf'!
  File "/usr/lib/python3.10/site-packages/SCons/Action.py", line 1279, in execute
    result = self.execfunction(target=target, source=rsources, env=env)
  File ".../modm/examples/generic/usb/modm/scons/site_tools/artifact.py", line 19, in run_store_artifact
    build_id.cache_elf(source[0].path, env["CONFIG_ARTIFACT_PATH"])
  File ".../modm/examples/generic/usb/modm/modm_tools/build_id.py", line 54, in cache_elf
    raise ValueError("Unable to find Build ID for '{}'!".format(source))
ValueError: Unable to find Build ID for 'build/usb/scons-debug/usb.elf'!
```

rleh:
Seems like my build_id is empty:

```
$ arm-none-eabi-objdump -d build/usb/scons-debug/usb.elf

[...]
0800817c g       .rodata	00000000 __build_id
[...]
```

What?

chris-durand:
That commit broke it:
```
commit 996e361d9ed9b59f0a4928bc3ddec22862d6665b
Author: Niklas Hauser <niklas.hauser@rwth-aachen.de>
Date:   Tue Jul 13 04:03:27 2021 +0200

    [core] Refactor linkerscript templates
    
    - Adds a generic linkerscript for all basic Cortex-M devices.
    - Adds .data_{ram} and .noinit_{ram} sections to all memories.
    - Uses MAX() to forward the heap location counter so that a very
      large .data section will collapse the heap to zero and "spill" into
      the next RAM section without overflowing the subsection SRAM1.
    - Generalizes RAM heap section placement via macro.
```

chris-durand:
The `build_id.py` script looks for a `.build_id` section in the elf file. That was removed during the linker script refactoring.

salkinium:
Ooops, I condensed the sections a little because otherwise the output of `scons size` would get spammed with a lot of unreleated stuff incl. `.build_id`. I'm checking how I can fix the modm_tools script. Bad Niklas, BAD!

rleh:
Next problem: 
```
scons debug-coredump
```
does not work at all because:

```
$ file ext/adamgreen/crashcatcher/CrashDebug/bins/lin64/CrashDebug 
ext/adamgreen/crashcatcher/CrashDebug/bins/lin64/CrashDebug: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), corrupted program header size, corrupted section header size
```

(The file is also not marked executable, bit that is solvable...)

rleh:
> corrupted program header size, corrupted section header size

Because decoding executable files as utf-8 and copying them line by line is a bad idea: https://github.com/modm-ext/CrashCatcher-partial/blob/beec94a93c42a8cad924330be16dc4888045ae9a/update.py#L39-L42
🤣

salkinium:
🙈🙈🙈🙈🙈

salkinium:
I didn't find a way to extract the build id without putting it in its own section, as you already fixed it. Instead I filter it out in the size tool and also fixed some duplicated sections I found.

salkinium:
"Raphael, can you fix TinyUSB on the F429?"

chris-durand:
> "Raphael, can you fix TinyUSB on the F429?"

Was so free to push the fix. Working on my Nucleo F446ZE. I have committed it already one hour ago before I noticed @rleh also found the issue.

chris-durand:
> Was so free to push the fix. Working on my Nucleo F446ZE. I have committed it already one hour ago before I noticed @rleh also found the issue.

Fix fails for H7B3 which has no full-speed USB peripheral but only a high speed one. I changed it to "full" as a default if available or else "high" if only that exists.

rleh:
> Next problem:
> 
> ```
> scons debug-coredump
> ```
> 
> does not work [...]

Still does not work:

<details>

```
Remote debugging using | [...]/modm/examples/generic/usb/modm/ext/crashcatcher/bins/lin64/CrashDebug --elf artifacts/846ef225a76bfe28ec3f01d097d6e5cb2b4e3ab1.elf --dump coredump.txt
ERROR: 

CrashDebug - Cortex-M Post-Mortem Debugging Aid (1.7-20210805.0)

Copyright (C) 2021 Adam Green
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Usage: CrashDebug (--elf elfFilename | --bin imageFilename baseAddress)
                   --dump dumpFilename
                  [--alias baseAddress size redirectAddress]
Where: NOTE: The --elf and --bin options are mutually exclusive.  Use one
             or the other but not both.
       --elf is used to provide the filename of the .elf image containing
         the device's FLASH contents at the time of the crash.
       --bin is used to provide the filename of the binary image loaded into
         the device's FLASH when the crash occurred. These binary images are
         typically generated by running:
           "arm-none-eabi-objcopy -O binary input.elf output.bin"
         The baseAddress parameter indicates where the contents of the .bin
         file was loaded into FLASH.  This address will typically be
         0x00000000 unless a boot loader was in use.
       --dump is used to provide the filename of the crash dump which
         contains the contents of RAM and the CPU registers at the time of
         the crash.  See the following link to learn more about generating
         these crash dumps:
           http://github.com/adamgreen/CrashDebug#crash-dump-generation
       --alias is used to trap memory accesses to the region defined
         by baseAddress/size and redirect them to the region at
         redirectAddress. For example acesses to baseAddress will access
         redirectAddress instead).
Encountered unexpected error: 30
Remote communication error.  Target disconnected.: Connection reset by peer.
Section .text, range 0x8000000 -- 0x80066b0: matched.
Section .rodata, range 0x80066b0 -- 0x8008240: matched.
Section .build_id, range 0x8008240 -- 0x8008264: matched.
Section .data, range 0x8008264 -- 0x80082fc: MIS-MATCHED!
Section .fastdata, range 0x80082fc -- 0x8008304: MIS-MATCHED!
warning: One or more sections of the target image does not match
the loaded file
```

</details>

Any idea @salkinium?

rleh:
> [CrashCatcher] still does not work: [...]
> Any idea @salkinium?

Solved be removing newline characters from the coredump.txt file.

rleh:
@chris-durand Do you have the `generic/usb/` example running on your Nucleo-F446ZE?

chris-durand:
> @chris-durand Do you have the `generic/usb/` example running on your Nucleo-F446ZE?

I'll try that now. I have only tested the F446ZE usbserial one so far.

rleh:
I just tested the `nucleo_f446ze/usbserial/` example on the Nucleo-F429ZI (only changing the board in `project.xml`): Works!

Once I add the following line it crashes again:

```diff
usb_stream << "Hello World from USB: " << (counter++) << modm::endl;
+MODM_LOG_INFO << "Hello World from serial connection! counter=" << counter << modm::endl;
```

chris-durand:
@rleh It crashes.

rleh:
Seems to be unrelated to USB:

I can also crash my Nucleo-F429ZI by just creating a second (uart based) IOStream:

```cpp
modm::IODeviceWrapper< modm::platform::Usart2, modm::IOBuffer::BlockIfFull > uart2_io_device;
modm::IOStream uart2_stream(uart2_io_device);

// in main loop:
uart2_stream << "Hello World from UART2: " << (counter++) << modm::endl;
```

chris-durand:
> Seems to be unrelated to USB:
> 
> I can also crash my Nucleo-F429ZI by just creating a second (uart based) IOStream:
> 

It seems to be USB-related. I can reproduce the crash but the code starts working if I don't call `tud_task();` from the main loop.

salkinium:
> Fix fails for H7B3 which has no full-speed USB peripheral but only a high speed one.

I made the option only appear for devices with a real choice and defaulted it properly.

Both the USB generic and the usb_freertos examples hardfault for me. Not sure why…

chris-durand:
I found the reason for the crash, classic off-by-one. The array `_desc_str` has size 32 but 32 elements are written into it starting from index 1 (`tusb_get_device_serial(_desc_str + 1);` in tusb_descriptors.c).

That was actually overwriting the lower half of the pointer to the logger device inside the `modm::log::info` object. When trying to call `write()` on the logger device the vtable pointer was pointing at the wrong address.

rleh:
Now **all** my problems are solved!
I disabled the FreeRTOS abstraction layer in TinyUSB since I don't see it's benefit and it causes problems.

I also added the parameter (`uint8_t priority=3`) to `Board::initializeUsbFs()` for all BSPs with USB support. This does not change existing behavior because `usb::initialize(uint8_t priority=3)` defaulted to the same interrupt priority.

Review & merge?

salkinium (approved):

chris-durand:
>  I disabled the FreeRTOS abstraction layer in TinyUSB since I don't see it's benefit and it causes problems.

The FreeRTOS integration seems to be needed when accessing tinyusb functions from multiple threads. When it is enabled mutexes are added in some places and the tinyusb fifo is implemented in terms of a FreeRTOS thread safe queue. Otherwise you'll get race conditions and undefined behaviour if you do that. I am not sure if we should go with that as the default, at least not without a big warning somewhere. What do you think @salkinium @rleh ?

The PR looks good otherwise.

salkinium:
> I am not sure if we should go with that as the default, at least not without a big warning somewhere. 

I mean, ideally we would solve the FreeRTOS issues in TinyUSB, otherwise we should definitely add a warning to the TinyUSB documentation about missing automatic support for FreeRTOS and perhaps open an issue to track it.

ext/hathach/module.lb
```diff
@@ -109,19 +113,22 @@ def build(env):
...
 
-    if env.has_module(":freertos"):
-        tusb_config["CFG_TUSB_OS"] = "OPT_OS_FREERTOS"
-        env.copy("tinyusb/src/osal/osal_freertos.h", "osal/osal_freertos.h")
-    else:
-        env.copy("tinyusb/src/osal/osal_none.h", "osal/osal_none.h")
+    # TinyUSB has a OS abstraction layer for FreeRTOS, but it is causes problems with modm
+    # if env.has_module(":freertos"):
+    #    tusb_config["CFG_TUSB_OS"] = "OPT_OS_FREERTOS"
+    #    env.copy("tinyusb/src/osal/osal_freertos.h", "osal/osal_freertos.h")
```
rleh, line 119:
```suggestion
    #    env.copy("tinyusb/src/osal/osal_freertos.h", "osal/osal_freertos.h")

    if env.has_module(":freertos"):
        env.log.warning("TinyUSB in modm is currently not using its FreeRTOS abstraction layer and is not thread-safe with FreeRTOS threads.")
```

rleh (commented):
Every other driver in modm is also not thread safe with FreeRTOS, so I hope that all people who are using FreeRTOS are aware of this.

But adding a warning is probably a good idea!

ext/hathach/module.lb
chris-durand: I would add a warning comment in the example source. If we have a specific tinyusb + FreeRTOS example and there is a tinyusb FreeRTOS integration I would assume that is what is being shown there.

rleh: Oh yes, good idea!

rleh:
I added a warning to the example and to the TinyUSB lbuild module. Example works on my hardware. Ready to merge from my side!

salkinium (approved):

chris-durand (approved):
Thanks!

rleh:
Thanks for your help @chris-durand and @salkinium!

PR #819, opened 2022-02-05
[board] Add configuration version for DISCO-F469NI

### LINKED ISSUE #818: STM32F469I-DISCO touch does not work on newer board revisions (B-03)
ST exchanged the touch screen controller [FT6206](https://modm.io/reference/module/modm-driver-ft6x06/) ([datasheet](https://cdn-shop.adafruit.com/datasheets/FT6x06+Datasheet_V0.1_Preliminary_20120723.pdf)) by FT6336 ([datasheet](https://focuslcds.com/content/FT6236.pdf)).

Touch controller on rev B-01:
Touchcontroller_rev_B-01
image

Touch controller on rev B-03:
Touchcontroller_rev_B-03
image

---

The driver is basically the same, but the I²C address has changed from `0x2a` to `0x38`. Should get fixed in the STM32F469I-DISCO board support package / examples.
I'm not aware of any way to detect the board revision from software, maybe we could try to ping the touch screen controller during initialization and then use the I²C address that responded?
Otherwise we could duplicate the lbuild board module or add a revision option to the lbuild module?

cc @kikass13 @macslow

salkinium:
This adds configuration versions for the DISCO-F469NI board `<extends>modm:disco-f469ni:b-03</extends>` to fix the different I2C address of the touch sensor.
See https://github.com/modm-io/modm/issues/818#issuecomment-1030415850.

Fixes #818.

- [x] Depends on https://github.com/modm-io/lbuild/pull/73
- [x] Update documentation
- [x] Test in hardware

cc @rleh

### DIFF
--- a/src/modm/board/disco_f469ni/module.lb
+++ b/src/modm/board/disco_f469ni/module.lb
@@ -14,6 +14,7 @@
 def init(module):
     module.name = ":board:disco-f469ni"
     module.description = FileReader("module.md")
+    # Revisions = [b-03, b-01]
 
 def prepare(module, options):
     if not options[":target"].partname.startswith("stm32f469nih"):
@@ -39,10 +40,16 @@ def build(env):
     env.outbasepath = "modm/src/modm/board"
     env.substitutions = {
         "with_logger": True,
-        "with_assert": env.has_module(":architecture:assert")
+        "with_assert": env.has_module(":architecture:assert"),
+        "touch_address": 0x38 if env[":disco-f469ni"] == "b-03" else 0x2a,
     }
     env.template("../board.cpp.in", "board.cpp")
-    env.copy('.')
+    env.copy("board_display.cpp")
+    env.copy("board_dsi.cpp")
+    env.copy("board_init.cpp")
+    env.copy("board_otm8009a.cpp")
+    env.copy("board_sdram.cpp")
+    env.template("board.hpp.in")
     env.collect(":build:openocd.source", "board/stm32f469discovery.cfg")
 
     env.collect(":platform:cortex-m:linkerscript.memory", linkerscript_memory)
--- a/docs/src/guide/custom-project.md
+++ b/docs/src/guide/custom-project.md
@@ -93,14 +93,16 @@ discover all of modm yet. You can now choose from two levels of customization:
 
 ## Using a Board Support Package
 
-Use lbuild to discover the specific BSP configuration you want to use:
+Use lbuild to discover the specific BSP configuration you want to use.
+Some board configurations support different hardware revisions, so check your
+hardware to select the right one:
 
 ```
  $ lbuild discover
 Parser(lbuild)
 ╰── Repository(modm @ ../../ext/modm)   modm: a barebone embedded library generator
     ├── Option(target) = REQUIRED in [stm32f469ngh6, stm32f469nih6, stm32f469vet6, ...
-    ├── Config(modm:disco-f469ni)   STM32F469IDISCOVERY
+    ├── Config(modm:disco-f469ni:b-03) in [b-01, b-03]   STM32F469IDISCOVERY
    ...
 ```
 
@@ -113,7 +115,7 @@ cannot remove any inherited modules.
 <library>
   <repositories>...</repositories>
 
-  <extends>modm:disco-f469ni</extends>
+  <extends>modm:disco-f469ni:b-03</extends>
 
   <options>...</options>
   <modules>...</modules>
--- a/docs/src/guide/discover.md
+++ b/docs/src/guide/discover.md
@@ -20,7 +20,7 @@ format. The `modm/repo.lb` file must be declared to find all modm modules:
 Parser(lbuild)
 ╰── Repository(modm @ .)   modm: a barebone embedded library generator
     ├── Option(target) = REQUIRED in [stm32f469ngh6, stm32f469nih6, stm32f469vet6, ...
-    ├── Config(modm:disco-f469ni)   STM32F469IDISCOVERY
+    ├── Config(modm:disco-f469ni:b-03) in [b-01, b-03]   STM32F469IDISCOVERY
    ...
 ```
 
@@ -54,7 +54,7 @@ modules for this specific target. We will choose the `stm32f469nih6` device:
 Parser(lbuild)
 ╰── Repository(modm @ .)   modm: a barebone embedded library generator
     ├── Option(target) = stm32f469nih6 in [stm32f407vgt6, stm32f469nih6, ...
-    ├── Config(modm:disco-f469ni)   STM32F469IDISCOVERY
+    ├── Config(modm:disco-f469ni:b-03) in [b-01, b-03]   STM32F469IDISCOVERY
    ...
     ├── Module(modm:board)   Board Support Packages
     │   ╰── Module(modm:board:disco-f469ni)   STM32F469IDISCOVERY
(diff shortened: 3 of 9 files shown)


examples/stm32f469_discovery/game_of_life/project.xml
```diff
@@ -1,5 +1,5 @@
 <library>
-  <extends>modm:disco-f469ni</extends>
+  <extends>modm:disco-f469ni/revB03</extends>
```
rleh: Before merging this we should also fix the actual issue with the B-03 revision. 😉

salkinium: Sure, I just wanted to prototype this idea quickly to get a feel for it.

salkinium: Well… do you think this is a good idea? This is basically just a shortcut for the hidden repo option… Alternatively we could add a shared revision option to the `modm:board` module, but it's not as concise as this and it's not a repo option so cannot influence modules that way.

salkinium:
I've changed the mechanism to be much less hacky and also discoverable by simply turning the config node into an enumeration option that can be queried like any other option. The significant advantage to this is that the specific versions can be declared and input checked like an enumeration option with a clear default value.
The syntax is now `modm:disco-f469ni:b03`, so a "submodule" of the config, but if you use `modm:disco-f469ni` it gets defaulted to the first value of the enumeration.

The versions are grepped from the `board.xml`, but it's not part of the config xml API, it's only convenient to store it there.

salkinium:
Here is the discoverablility:
```
$ lbuild discover
Parser(lbuild)
╰── Repository(modm @ ../../..)   modm: a barebone embedded library generator
    ├── Option(target) = stm32f469nih6 in ...
    ├── Config(modm:disco-f100rb)   STM32VLDISCOVERY
    ├── Config(modm:disco-f303vc)   STM32F3DISCOVERY
    ├── Config(modm:disco-f407vg)   STM32F4DISCOVERY
    ├── Config(modm:disco-f429zi)   STM32F429IDISCOVERY
    ├── Config(modm:disco-f469ni) = b03 in [b01, b02, b03]   STM32F469IDISCOVERY
```

```
 $ lbuild discover :disco-f469ni
>> modm:disco-f469ni  [Configuration]

# STM32F469IDISCOVERY

Discovery kit with STM32F469NI MCU.

This BSPs `Board::initialize()` sets the board up with these settings:

- 180MHz CPU clock frequency in overdrive mode.
- 90MHz APB2 and 45MHz APB1 clock frequency.
- MODM_LOGGER initialized at 115.2kBaud outputting on STLink.
- 16MB SDRAM initialized for heap.
- Disables the display by default.

You can enable and use the on-board display in two ways:

1. Use the `modm::GraphicDisplay` returned by `Board::getDisplay()`, which
   initializes the display on first call and allocates the framebuffer in the
   external SDRAM. This is the simplest option.
2. Call `Board::initializeDisplay()` and then provide your own display buffer
   with `Board::setDisplayBuffer(void*)`.

!!! warning "Caveats"
    No double-buffering is implemented, so you may notice some tearing if you
    update the display fast enough.

Access to the capacitive touchscreen is provided in the `Board::ft6` namespace.
Call `Board::initializeTouchscreen()` to setup the peripherals.

[Product Link](https://www.st.com/en/evaluation-tools/32f469idiscovery.html)

Version: b03
Versions: [b01, b02, b03]
```

src/modm/board/disco_f469ni/board.hpp
```diff
@@ -164,6 +164,7 @@ using Scl = GpioB8;
 using Sda = GpioB9;
 using I2cMaster = I2cMaster1;
 using Touch = modm::Ft6x06< I2cMaster >;
+constexpr uint8_t TouchAddress = {{touch_address}};
```
rleh, line 167: Shouldn't this file get renamed to `board.hpp.in`?

src/modm/board/disco_f469ni/module.lb
```diff
@@ -39,10 +39,16 @@ def build(env):
     env.outbasepath = "modm/src/modm/board"
     env.substitutions = {
         "with_logger": True,
-        "with_assert": env.has_module(":architecture:assert")
+        "with_assert": env.has_module(":architecture:assert"),
+        "touch_address": 0x38 if env.get(":disco-f469ni", "b-03") == "b-03" else 0x2a,
```
rleh: Are you sure revision _B-02_ also has touch controller address `0x38`?

rleh (approved):
Nice, thanks! 👍🏽

src/modm/board/disco_f469ni/module.lb
salkinium: I couldn't find any mention of hardware revisions in the official kit PDF, but I found the [two addresses in the CubeHAL BSP](https://github.com/STMicroelectronics/STM32CubeF4/blob/2f3b26f16559f7af495727a98253067a31182cfc/Drivers/BSP/STM32469I-Discovery/stm32469i_discovery.h#L231-L239), but they [detect the driver address at runtime without mentioning revision](https://github.com/STMicroelectronics/STM32CubeF4/blob/2f3b26f16559f7af495727a98253067a31182cfc/Drivers/BSP/STM32469I-Discovery/stm32469i_discovery_ts.c#L173-L182).

So I don't know and I cannot check it…

rleh: Then I would prefer to define a B-02 revision which we do not know if it exists at all and which we have not tested.

rleh: Then I would prefer to not define a B-02 revision which we do not know if it exists at all and which we have not tested.

src/modm/board/disco_f469ni/board.xml
```diff
@@ -4,6 +4,7 @@
       <path>../../../../repo.lb</path>
     </repository>
   </repositories>
+  <!-- <versions>b-03,b-02,b-01</versions> -->
```
rleh: Uncomment?

salkinium: Somewhere the revisions need to be declared, but I've changed the mechanism to read from the `module.lb` instead of the `board.xml`.

rleh: Ok for me, but currently that does not happen in `module.lb`.

src/modm/board/disco_f469ni/module.lb
```diff
@@ -14,6 +14,7 @@
 def init(module):
     module.name = ":board:disco-f469ni"
     module.description = FileReader("module.md")
+    # Revisions = [b-03, b-01]
```
salkinium, line 17: here

rleh: But the line is commented out?!

salkinium: The config is just a named (versioned) alias to a `project.xml` file inside modm. It must be declared in `repo.lb` so it can act as a repository option, therefore you need a mechanism to pass the revision to the `repo.lb` without having a giant table there, which is currently done by regexing the `board/*/module.lb` files. It's not great, but I don't know a better way. Ideas?

rleh: Maybe create a `module.revisions` file and add one configuration name per line?

salkinium: Let's fix it when it becomes a problem. I first would like to see more board revision examples to decide a better way.

PR #820, opened 2022-02-05
[architecture] Require setting SPI config in `SpiDevice` based drivers

chris-durand:
PR to move the discussion about a new API for setting the SPI configuration in `SpiDevice` based device drivers to a different place.

As a first step a new `SpiConfiguration` struct is added under `architecture`. Furthermore, `modm::SpiDevice` will take that config struct as a non-type template parameter and provides a default `configureBus()` function that applies the settings to the SPI master.

EDIT: this does not work as expected, the platform-specific `SpiMaster` driver won't accept `modm::Spi::DataMode` as an argument to `SpiMaster::setDataMode()`. `modm::platform::SpiMaster::DataMode` has to be used.

### DIFF
--- a/examples/nucleo_f429zi/adc_ads816x/main.cpp
+++ b/examples/nucleo_f429zi/adc_ads816x/main.cpp
@@ -36,6 +36,7 @@ main()
 
 	SpiMaster::initialize<Board::SystemClock, 10_MHz>();
 	SpiMaster::connect<Sck::Sck, Mosi::Mosi, Miso::Miso>();
+	adc.configureBus();
 
 	Rst::setOutput();
 	Cs::setOutput(true);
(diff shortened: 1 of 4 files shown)


chris-durand:
The `DataMode` type is dependent on the peripheral driver. I have tried to make the config struct a template. Now gcc segfaults:
```
In file included from main.cpp:14:
modm/src/modm/driver/adc/ads816x.hpp:102:65: internal compiler error: Segmentation fault
  102 |     public modm::SpiDevice<SpiMaster, Ads816xSpiConfig<SpiMaster>>,
      |                                                                 ^~
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://gcc.gnu.org/bugs/> for instructions.
scons: *** [/home/chris/modm/build/nucleo_f429zi/adc_ads816x/scons-release/main.o] Error 1
```
:laughing:

chris-durand:
I found the mistake, now without segfault.

PR #821, opened 2022-02-07
[sam] Fix SPI bug in handling of CPHA/CPOL

mcbridejc:
This corrects my earlier mistake in handling the SPI mode flags.


chris-durand (approved):
Thanks! Verified it with the SAMG55 datasheet, the change looks correct.

salkinium (approved):

PR #822, opened 2022-02-07
[sam] GPIO setOutput overrides peripheral control

mcbridejc:
Calling setOutput now overrides peripheral control of the pin, which is
consistent with the STM32 GPIO implementation. This commit also
ensures output value is set before the output drive is enabled to prevent
glitch.

### DIFF
--- a/src/modm/platform/gpio/sam/pin.hpp.in
+++ b/src/modm/platform/gpio/sam/pin.hpp.in
@@ -278,6 +278,7 @@ template<class... PinConfigs>
 struct PinCfgMixin
 {
 	inline static void set(uint8_t){};
+	inline static void set(uint8_t, uint8_t){};
 };
 
 template<class PinConfig, class... PinConfigs>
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Great!

chris-durand (commented):
Thanks, could you also do it for `setInput()`? On STM32 this will also reset the alternate function mode to plain input.

chris-durand:
I will also make the same change for the other SAM GPIO implementation and add it here.

mcbridejc:
>

mcbridejc:
> Thanks, could you also do it for `setInput()`? On STM32 this will also reset the alternate function mode to plain input.

That seems pretty sensible. Done.

chris-durand:
I found another bug in the other GPIO implementation (D2x, D/E5x). Calling `configure(InputMode)` will reset peripheral control as well but this function should only configure the pull-up/down mode. Will fix it.

chris-durand:
I have pushed the fixes for the SAM D21, D/E5x, GPIO implementation. I will test it tomorrow on a E54 dev board.

chris-durand:
> I will test it tomorrow on a E54 dev board.

It works. I cherry-picked the commits onto the SAME5x PR. With the changes from this PR the button on the dev board works, without them it doesn't. That's proof enough for me.

I will squash the fixup commits and rebase.

chris-durand (approved):

PR #823, opened 2022-02-07
[sam] Add driver for SAMG timer channels

mcbridejc:


examples/samg55_xplained_pro/timer/main.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <modm/io/iostream.hpp>
+#include <modm/platform.hpp>
+
+using namespace modm::platform;
+using namespace modm::literals;
+
+MODM_ISR(TC3)
+{
+	// Clear pending interrupts by reading them
+	(void)TimerChannel3::getInterruptFlags();
```
salkinium, line 21:
```suggestion
	TimerChannel3::acknowledgeInterruptFlags();
```
At least this is the informal naming scheme for the interrupt flags in modm.
Can be implemented like that though.

salkinium (changes requested):
Literally LGTM, since I cannot judge the technical correctness in detail ;-P

examples/samg55_xplained_pro/timer/main.cpp
mcbridejc: Yeah I think I had to think about this a bit. On STM32, you clear the flag with a write. On SAMG5x you clear it with a read, so the `acknowledgeInterruptFlags()` function, if I added it, would be the same as the `getInterruptFlags()` function, except I guess not returning the read value. Adding the extra method does allow for more similar API I suppose, and it's likely to be inlined as a single instruction anyway so I guess it doesn't hurt to have both.

mcbridejc: Actually, it won't match the STM32 API anyway, because `acknowledgeInterruptFlags` takes a mask argument, but there's no way to acknowledge interrupt flags individually on this chip.

chris-durand (approved):
Very nice!

examples/samg55_xplained_pro/timer/main.cpp
chris-durand: One thing that I do not really like about the name `getInterruptFlags()` is that functions named `get...` usually perform a read operation without side effects. From its name it's not obvious it clears any flags. That's why the code above needs the comment.

chris-durand: I can't come up with a better name despite some overly verbose alternatives like `readAndClearInterruptFlags()`.

mcbridejc: I'm happy with renaming getInterruptFlags to readAndClearInterruptFlags, adding an acknowledgeInterruptFlags that is basically a wrapper on getInterruptFlags, or leaving it as is.

salkinium: Let's leave it as is and fix it later if it becomes an issue.

salkinium (approved):

PR #831, opened 2022-02-21
Add drawBitmap method to ColorGraphicsDisplay interface

mcbridejc:
Hey want to solicit thoughts on this change, or something like it, @TomSaw, @salkinium.

The [ILI9341](https://github.com/modm-io/modm/blob/develop/src/modm/driver/display/ili9341.hpp#L283) driver offers a drawBitmap method, which can take a an Rgb565 bitmap as input, but this is not in the ColorGraphicsDisplay interface, and that's a problem because the ILI9341 has many template arguments so it's awkward to pass a reference to it.

Should we just add this virtual method to ColorGraphicsDisplay?

### DIFF
--- a/src/modm/ui/display/color_graphic_display.hpp
+++ b/src/modm/ui/display/color_graphic_display.hpp
@@ -51,6 +51,21 @@ public:
 		return backgroundColor;
 	}
 
+	/**
+	 * Draw a bitmap with 16-bit 565 colored pixels.
+	 *
+	 * \p data is series of pixel color values
+	 */
+	virtual void drawBitmap(
+		glcd::Point upperLeft,
+		uint16_t width,
+		uint16_t height,
+		modm::accessor::Flash<uint8_t> data)
+	{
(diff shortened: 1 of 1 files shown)


salkinium:
I guess? I don't really have an overview anymore over the graphics API, so I'm delegating this to our expert @TomSaw…

TomSaw:
Hey @mcbridejc. I'm sitting on a complete rewrite of the graphic APIs since 2021.
The new architecture is quite flexible and should solve your issue:
It allows copying+converting a bitmap in any imaginable colorspace (f.e. RGB565) stored somewhere (f.e. in flash memory) to any other writable memory in any imaginable colorspace, be it the local RAM or the displays CGRAM.

See these these dependent PRs:

- #781
- #782
- #690
- Very generic graphic buffers [expected Q2 2022]
- Rewrite ILI9341 for new graphic APIs [expected Q2 2022]
- Adapt SSD1306 to new graphic APIs [expected Q2 2022]

PR #832, opened 2022-02-22
[docs] Document MODM_DEBUG_BUILD macro behavior

salkinium:
For some reason I completely forgot to document this, bad Niklas!

cc @TomSaw would this have helped you discover this behavior?


TomSaw:
> For some reason I completely forgot to document this, bad Niklas!
> 
> cc @TomSaw would this have helped you discover this behavior?

Yea that's it!

PR #839, opened 2022-03-17
[example] Servo PWM example on Nucleo-G474RE

rleh:
Just found that commit from last year laying around...

### DIFF
--- a/examples/nucleo_g474re/servo_pwm/main.cpp
+++ b/examples/nucleo_g474re/servo_pwm/main.cpp
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2021, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <modm/board.hpp>
+
+using namespace Board;
+
+int
+main()
+{
+	Board::initialize();
+	LedD13::setOutput();
+
+	MODM_LOG_INFO << "modm example: Servo PWM" << modm::endl << modm::endl;
+	MODM_LOG_INFO << "Generates a 50 Hz PWM signal with a pulse width between 1 ms and 2 ms suitable" << modm::endl;
+	MODM_LOG_INFO << "for many servo motors or RC electronic speed controllers on pin B7." << modm::endl;
+	MODM_LOG_INFO << "1.5 ms pulse width is output when the user button is pressed, " << modm::endl;
+	MODM_LOG_INFO << "otherwise 1 ms pulse width." << modm::endl << modm::endl;
+
+	Timer4::connect<GpioB7::Ch2>();
+	Timer4::enable();
+	Timer4::setMode(Timer4::Mode::UpCounter);
+
+	constexpr uint16_t overflow = 0xffff;
+	constexpr uint16_t minPwm = (overflow / 20.f) * 1.f; // 1 ms pulse width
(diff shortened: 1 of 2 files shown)


examples/nucleo_g474re/servo_pwm/main.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+using namespace Board;
+
+int
+main()
+{
+	Board::initialize();
+	LedD13::setOutput();
+
+	MODM_LOG_INFO << "modm example: Serve PWM" << modm::endl << modm::endl;
```
rleh:
```suggestion
	MODM_LOG_INFO << "modm example: Servo PWM" << modm::endl << modm::endl;
```

examples/nucleo_g474re/servo_pwm/main.cpp
```diff
@@ -0,0 +1,78 @@
...
+	Timer4::setMode(Timer4::Mode::UpCounter);
+
+	constexpr uint16_t overflow = 0xffff;
+	constexpr uint16_t minPwm = (overflow / 20) * 1.0; // 1 ms pulse width
+	constexpr uint16_t targetPwm = (overflow / 20) * 1.5; // 1.5 ms pulse width
+	constexpr uint16_t maxPwm = (overflow / 20) * 2.0; // 2 ms pulse width
+
+	// 170 MHz / 52 / 2^16 = approx. 50 Hz
+	Timer4::setPrescaler(52);
+	Timer4::setOverflow(overflow);
```
salkinium, line 38: I think you can make this more generic via the `setPeriod` method, sadly via runtime computation now (only because of the API limitation).

```cpp
overflow = Timer4::setPeriod<SystemClock>(20'000);
// compute minPwm, targetPwm, maxPwm at runtime?
```

rleh: The problem with `setPeriod()` is that it computes the prescaler and overflow with the goal of minimizing the tolerance and not maximizing the overflow value; it does not even guarantees a minimal overflow value.
In the most use-cases it is acceptable to have a small overflow value which can significantly degrade the accuracy of the pulse widths.

rleh: I would like to see something like
```
Timer::setPrescaler<class SystemClock, uint32_t period_us, uint32_t overflow = 0xffff, tolerance_t tolerance = 1_pct>()
```
in the timer API instead.
But improving the timer API is a different topic...

examples/nucleo_g474re/servo_pwm/main.cpp
```diff
@@ -0,0 +1,78 @@
...
+	MODM_LOG_INFO << "otherwise 1 ms pulse width." << modm::endl << modm::endl;
+
+	Timer4::connect<GpioB7::Ch2>();
+	Timer4::enable();
+	Timer4::setMode(Timer4::Mode::UpCounter);
+
+	constexpr uint16_t overflow = 0xffff;
+	constexpr uint16_t minPwm = (overflow / 20) * 1.0; // 1 ms pulse width
+	constexpr uint16_t targetPwm = (overflow / 20) * 1.5; // 1.5 ms pulse width
+	constexpr uint16_t maxPwm = (overflow / 20) * 2.0; // 2 ms pulse width
```
chris-durand: The division is done in integer arithmetic but the multiplication is float. Did you want to do that?

examples/nucleo_g474re/servo_pwm/main.cpp
salkinium: Yeah, good points…

examples/nucleo_g474re/servo_pwm/main.cpp
rleh: Nop, thanks for pointing that out...

salkinium (approved):

chris-durand (approved):

PR #850, opened 2022-04-21
MAX31855 driver

rasmuskleist:
The driver has been tested using a custom board with stm32g474ret6

### DIFF
--- a/examples/nucleo_g474re/max31855/main.cpp
+++ b/examples/nucleo_g474re/max31855/main.cpp
@@ -1,0 +1,99 @@
+// coding: utf-8
+/*
+ * Copyright (c) 2022, Sarah Vilete
+ * Copyright (c) 2022, Rasmus Kleist Hørlyck Sørensen
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <modm/board.hpp>
+#include <modm/processing.hpp>
+#include <modm/driver/temperature/max31855.hpp>
+
+using SpiMaster = modm::platform::SpiMaster2;
+
+using Cs = modm::platform::GpioA10;
+using Miso = modm::platform::GpioB14;
+using Sck = modm::platform::GpioB13;
+
+using namespace Board;
+
+class ThermocoupleThread : public modm::pt::Protothread
+{
+public:
+    ThermocoupleThread() : thermocouple(data) {}
+
+    bool
+    run()
+    {
+        PT_BEGIN();
+
+        thermocouple.initialize();
+
+        MODM_LOG_INFO << "Max38155 initialized succeded" << modm::endl;
+
+        while (true)
+        {
+            PT_CALL(thermocouple.readout());
+
+            switch (data.getFault())
+            {
+            case modm::max31855::Fault::ShortCircuitVcc:
+                MODM_LOG_ERROR << "Thermocouple error short circuit vcc" << modm::endl;
+                break;
+
+            case modm::max31855::Fault::ShortCircuitGnd:
+                MODM_LOG_ERROR << "Thermocouple error short circuit gnd" << modm::endl;
+                break;
+
+            case modm::max31855::Fault::OpenCircuit:
+                MODM_LOG_ERROR << "Thermocouple error open circuit" << modm::endl;
+                break;
+
+            default:
+                MODM_LOG_INFO << "Thermocouple Temperature: " << data.getThermocoupleTemperature() << " degrees Centigrade" << modm::endl;
+                MODM_LOG_INFO << "Internal Temperature: " << data.getReferenceJunctionTemperature() << " degrees Centigrade" << modm::endl;
+                break;
+            }
+
+            timeout.restart(std::chrono::milliseconds(1000));
+            PT_WAIT_UNTIL(timeout.isExpired());
+        }
+
+        PT_END();
+    }
+
+private:
+    modm::max31855::Data data;
+    modm::Max31855<SpiMaster, Cs> thermocouple;
+
+    modm::ShortTimeout timeout;
+} thermocoupleThread;
+
+int
+main()
+{
+    Board::initialize();
+    Cs::setOutput(modm::Gpio::High);
+
+    SpiMaster::connect<Miso::Miso, Sck::Sck>();
+    SpiMaster::initialize<Board::SystemClock, 656250>();
+
+    MODM_LOG_INFO       << "==========MAX 31855 Test==========" << modm::endl;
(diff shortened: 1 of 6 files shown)


src/modm/driver/temperature/max31855.hpp
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @param data pointer to buffer of the internal data of type Data
+     */
+    Max31855(Data &data);
+
+    /// Call this function once before using the device
+    void
+    initialize();
+
+    /// Read the data from the sensor into
```
rleh: ... into what?

rleh (commented):
Nice, thanks!

src/modm/driver/temperature/max31855.hpp
rasmuskleist: Sorry I will remove that comment

rleh:
> The driver has been tested using a custom board with stm32g474ret6

Could you somehow create an example for (e.g.) the Nucleo-G474RE board, so that the CI at least compiles the code?

rasmuskleist:
I have used the pins in the test that we have allocated on our board for the MAX31855

src/modm/driver/temperature/max31855.hpp
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return static_cast<Fault>(data[3] & 0b111);
+        }
+
+        /// @return the thermocouple temperature scaled according to the device documentation
+        float
+        getThermocoupleTemperature()
+        {
+            // convert raw 14 bit readout in 2's complement to a 16 bit signed
+            const int16_t rawTemp = ((static_cast<int16_t>(data[0]) << 8) | (data[1] & 0b11111100)) / 4;
```
chris-durand: This won't work for negative temperatures. For the data value of -1°C from the data sheet this implementation [will give a result of 4095 °C](https://gcc.godbolt.org/z/K7hhcT3h8).

The two's complement conversion is not correct due to [integer type promotion](https://en.cppreference.com/w/cpp/language/implicit_conversion#Integral_promotion). The expression `static_cast<int16_t>(data[0]) << 8` and also the whole value before dividing by 4 has the type `int`. Thus, you'll get a big positive value for negative temperatures.

```suggestion
            const int16_t rawTemp = static_cast<int16_t>((data[0] << 8) | (data[1] & 0b11111100)) / 4;
```

src/modm/driver/temperature/max31855.hpp
```diff
@@ -0,0 +1,104 @@
...
+        ShortCircuitVcc = Bit2,
+        ShortCircuitGnd = Bit1,
+        OpenCircuit     = Bit0
+    };
+
+    struct modm_packed
+    Data
+    {
+        /// @return value associated with the respective fault
+        Fault getFault()
```
chris-durand: This should be callable on a `const Data` object and could also be `constexpr`.
```suggestion
        constexpr Fault getFault() const
```

src/modm/driver/temperature/max31855.hpp
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return static_cast<Fault>(data[3] & 0b111);
+        }
+
+        /// @return the thermocouple temperature scaled according to the device documentation
+        float
+        getThermocoupleTemperature()
```
chris-durand:
```suggestion
        constexpr float
        getThermocoupleTemperature() const
```

src/modm/driver/temperature/max31855.hpp
```diff
@@ -0,0 +1,104 @@
...
+        {
+            // convert raw 14 bit readout in 2's complement to a 16 bit signed
+            const int16_t rawTemp = ((static_cast<int16_t>(data[0]) << 8) | (data[1] & 0b11111100)) / 4;
+            return 0.25f * rawTemp;
+        }
+
+        /// @return the reference junction temperature scaled according to the device documentation
+        float
+        getReferenceJunctionTemperature()
```
chris-durand:
```suggestion
        constexpr float
        getReferenceJunctionTemperature() const
```

src/modm/driver/temperature/max31855.hpp
```diff
@@ -0,0 +1,104 @@
...
+        {
+            // convert raw 12 bit readout in 2's complement to a 16 bit signed
+            const int16_t rawTemp = ((static_cast<int16_t>(data[2]) << 8) | (data[3] & 0b11110000)) / 16;
```
chris-durand: Same as above

```suggestion
            const int16_t rawTemp = static_cast<int16_t>((data[2] << 8) | (data[3] & 0b11110000)) / 16;
```

examples/nucleo_g474re/max31855/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+public:
+    ThermocoupleThread() : thermocouple(data) {}
+
+    bool
+    run()
+    {
+        PT_BEGIN();
+
+       	thermocouple.initialize();
```
chris-durand: indentation

chris-durand (changes requested):
Thank you!

examples/nucleo_g474re/max31855/main.cpp
rasmuskleist: Sorry what do you mean?

chris-durand: Could be my browser but on my end it looks like there is one excess space in front of `thermocouple.initialize(); ` in line 35.

rasmuskleist: To me they are at the same level of indentation

src/modm/driver/temperature/max31855.hpp
rasmuskleist: Why would this be constexpr? The return value value is not constant

examples/nucleo_g474re/max31855/main.cpp
chris-durand: indentation

There is a tab character. Also lines 80 and 81 are indented with tabs instead of spaces like the rest of the file.

src/modm/driver/temperature/max31855.hpp
rasmuskleist: Whoops, yes you are right sorry

src/modm/driver/temperature/max31855.hpp
chris-durand: It does not have to be. It just means it can be used on a `Data` object in a `constexpr` context. C++20 even has `constexpr` support for `std::vector` with dynamic allocation at compile-time (not sure if any compiler has that implemented yet, though).
It is not that beneficial here with the intended usage of the type, but it certainly does not hurt.

examples/nucleo_g474re/max31855/main.cpp
rasmuskleist: Okay I will change it to spaces

chris-durand (approved):
Nice, thank you!

salkinium (approved):

rleh (approved):

PR #851, opened 2022-04-27
Feature ms5611

rasmuskleist:
I have had this lying around for a while now. There are some points I am not quite happy with yet. More precisely;

1. The `reinterpret_cast<uint16_t *>(&data.calibration)` gives a compiler warning
2. There is no guard on the `readProm()` preventing indices not in range 0..7
3. It may be redundant to have `rawTemperatureTouched` and `rawPressureTouched` since it only makes sense to measure both at once

I would like to hear your thoughts?

### DIFF
--- a/src/modm/driver/pressure/ms5611.lb
+++ b/src/modm/driver/pressure/ms5611.lb
@@ -1,0 +1,34 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2022, Rasmus Kleist Hørlyck Sørensen
+#
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+def init(module):
+    module.name = ":driver:ms5611"
+    module.description = """\
+# MS5611 Barometric Pressure Sensor
+This barometric pressure sensor is optimized for altimeters and variometers with an altitude resolution of 10 cm.
+"""
+
+def prepare(module, options):
+    module.depends(
+        ":architecture:gpio",
+        ":architecture:spi.device",
+        ":math:utils",
+        ":processing:resumable",
+        ":processing:timer")
+    return True
+
+def build(env):
+    env.outbasepath = "modm/src/modm/driver/pressure"
+    env.copy("ms5611.hpp")
+    env.copy("ms5611_impl.hpp")
+    env.copy("ms5611_data.hpp")
+    env.copy("ms5611_data_impl.hpp")
--- a/src/modm/driver/pressure/ms5611_data.hpp
+++ b/src/modm/driver/pressure/ms5611_data.hpp
@@ -167,7 +167,7 @@ private:
     int32_t calibratedTemperature;
 };
 
-} // ms5611data
+} // ms5611data namespace
 
 } // modm namespace
 
--- a/src/modm/driver/pressure/ms5611_data.hpp
+++ b/src/modm/driver/pressure/ms5611_data.hpp
@@ -1,0 +1,176 @@
+// coding: utf-8
+// ----------------------------------------------------------------------------
+/*
+ * Copyright (c) 2021, Rasmus Kleist Hørlyck Sørensen
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_MS5611_DATA_HPP
+#define MODM_MS5611_DATA_HPP
+
+namespace modm
+{
+
+// Forward declaration of the driver class
+template <typename SpiMaster, typename Cs>
+class Ms5611;
+
+namespace ms5611data
+{
+
+/**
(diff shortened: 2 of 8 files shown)


src/modm/driver/pressure/ms5611.lb
```diff
@@ -0,0 +1,35 @@
...
+# MS5611 Barometric Pressure Sensor
+This barometric pressure sensor is optimized for altimeters and variometers with an altitude resolution of 10 cm.
+"""
+
+def prepare(module, options):
+    module.depends(
+        ":architecture:spi",
+        ":architecture:spi.device",
+        ":math:utils",
+        ":debug",
```
salkinium:
```suggestion
```

Don't think `MODM_LOG_*` is used in this code and the debug module is quite large.

rleh:
> 1. The reinterpret_cast<uint16_t *>(&data.calibration) gives a compiler warning

There is also the following statement in ms5611_impl.hpp:173: `modm::fromBigEndian(*reinterpret_cast<uint16_t*>(buffer))`.
The compiler is correct, because such a cast might lead to a bus fault on (e.g.) Cortex-M0 if `buffer` (uint8_t array) is not 16-bit aligned in the memory.
There is a helper to solve this: [`modm::asUnaligned<>()`](https://docs.modm.io/develop/api/nucleo-f429zi/structmodm_1_1unaligned__t.html). It should be possible to just replace `reinterpret_cast` with `modm::asUnaligned`.

> 2. There is no guard on the `readProm()` preventing indices not in range 0..7

I see `(address & 0b111)` in the implementation, an input value >7 is harmless. Would be nice to add the allowed range the the documentation (ms5611.hpp:107).

> 3. It may be redundant to have `rawTemperatureTouched` and `rawPressureTouched` since it only makes sense to measure both at once

I don't know the sensor and the possible applications to be able to judge that. I have no problem with how it is currently implemented.

chris-durand:
> The compiler is correct, because such a cast might lead to a bus fault on (e.g.) Cortex-M0 if buffer (uint8_t array) is not 16-bit aligned in the memory.
There is a helper to solve this: modm::asUnaligned<>(). It should be possible to just replace reinterpret_cast with modm::asUnaligned.

The use of `modm::asUnaligned` is still undefined behaviour in terms of the language, independent of any architecture, even in C in that case. In C++ you may not generally cast a pointer of an object to an unrelated type and then dereference it (there are exceptional cases where it is allowed, e.g. for casts to `char` or `std::byte` to access the byte representation of an object). `modm::asUnaligned` is a no-op for anything that is not a Cortex-M0 and just dereferences the pointer.

Apart from the memory alignment issue there is another reason for that "strict aliasing rule". It allows the compiler to perform an optimization where it assumes objects of different types do not overlap in memory.

In a lot of cases the UB code will emit the expected instructions but a compliant C++ compiler may compile it to anything it wants. The rules are explained here: https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing

For GCC we might get away with it putting a `__attribute((__may_alias__))` somewhere in `asUnaligned` but I honestly don't know really what it does and that would not work for clang and other future compilers.

My recommendation in this case would be to replace the whole expression with a boring bit-shift operation like `(buffer[0] << 8) | buffer[1]`. That is basically the same as `fromBigEndian` would do anyway but without `reinterpret_cast` and UB.

rasmuskleist:
Thanks a lot for the comments I really appreciate your nice feedback! I will make the necessary changes during Saturday

rasmuskleist:
Rewriting parts of the driver I have come across some strange behaviour which I have not been able to explain myself. If not either casting the result of `readProm()` or using `modm::fromBigEndian()` the contents of the prom is incorrect. I even experienced the program crashing consistently when not casting. Currently I have not experienced any problems with the implementation but I wonder why casting is so critical?

src/modm/driver/pressure/ms5611_data.hpp
```diff
@@ -0,0 +1,175 @@
...
+class Ms5611;
+
+namespace ms5611data
+{
+
+/**
+ * @brief Holds the factory calibration data from the PROM.
+ * Upon initialization the driver automatically reads the calibration values from the PROM
+ */
+struct modm_packed
```
salkinium, line 32:
```suggestion
/// @ingroup modm_driver_ms5611
struct modm_packed
```
Otherwise it'll get lost somewhere in the depths of Doxypress…

chris-durand:
> If not either casting the result of `readProm()` or using `modm::fromBigEndian()` the contents of the prom is incorrect. I even experienced the program crashing consistently when not casting.

The program crashes in case you remove the `modm::fromBigEndian()` [there](https://github.com/modm-io/modm/blob/a364cd322b14d64a635107c402f0d70547dcb861/src/modm/driver/pressure/ms5611_impl.hpp#L61)?

rasmuskleist:
Yes that is right. Everything works swiftly with the ´modm::fromBigEndian´but removing it causes the program to crash as soon as the protothread is run the first time.

salkinium:
I'm fine with your PR, unless you're willing to debug our resumable functions (which probably suffer from UB anyways), I would just merge it as is. cc @chris-durand

rasmuskleist:
Debugging your resumable functions would be an interesting task for the future. However, as for now I am fine with merging.

salkinium:
Ok, rebase, squash into two commits?  Something like [driver] Add MS5611 driver + [example] Add MS5611 example.

chris-durand:
I'd be fine with merging for now but I somehow have a bad feeling about it. Last time we debugged why adding another overload for `operator delete()` made the delay run faster we found that our linker script put the code into memory with messed up alignment and it skipped instructions ...
Let's hope it is only related to the resumable functions and we did not trigger another hidden bug.

salkinium (approved):

PR #853, opened 2022-04-29
Fix STM32H7 RCC clock output

### LINKED ISSUE #852: STM32H7 RCC clock output configuration not working
The RCC driver has the following code:
```
%% if target.family in ["f2", "f4", "f7", "h7"]
	static inline bool
	enableClockOutput1(ClockOutput1Source src, uint8_t div)
	{
		uint32_t tmp = RCC->CFGR & ~(RCC_CFGR_MCO1 | RCC_CFGR_MCO1PRE);
		if (div > 1) tmp |= (div + 2) << 24;
		RCC->CFGR = tmp | uint32_t(src);
		return true;
	}
```

The prescaler calculation does not match the device registers:
mco

chris-durand:
Fix STM32H7 RCC clock output and add `connect()` method for alternate function configuration of clock outputs to RCC driver. Tested in hardware on a STM32H755ZI-Q Nucleo.

I haven't yet added an example because support for STM32H7 dual-core is still work in progress. It will go into #844.

Fixes #852

### DIFF
--- a/src/modm/platform/clock/stm32/rcc.hpp.in
+++ b/src/modm/platform/clock/stm32/rcc.hpp.in
@@ -318,7 +318,7 @@ public:
 	};
 %% endif
 
-%% if target.family in ["f2", "f4", "f7", "h7"]
+%% if target.family in ["f2", "f4", "f7"]
 	enum class
 	ClockOutput1Source : uint32_t
 	{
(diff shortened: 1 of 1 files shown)


chris-durand:
Does anyone have an idea why the Windows hosted examples fail? I have already restarted the job once. This PR only touches `platform/clock/stm32/rcc.hpp.in` …

salkinium:
> Does anyone have an idea why the Windows hosted examples fail?

I've seen it on other PRs too, I think something changed on their end, maybe we need to upgrade the OS version or something…

chris-durand:
The Windows unit test executable crashes:
```
Linking········ scons-release\test.exe
scons: *** [phony_target] Error 3221225477
```

chris-durand:
The windows CI fails on all PRs. The compiler version is exactly the same as on the last runs where the unit tests succeeded. I could also exclude that any changes we made on modm recently cause this. I have just rerun the Windows pipeline for a state that previously was fine but it also fails now.

salkinium (approved):
Nice!
Do you want this merged now, or is this going into the large PR later?

chris-durand:
> Nice! Do you want this merged now, or is this going into the large PR later?

I'd rather like this to be merged soon. I am working on STM32 camera interface support and need the clock out to supply the camera with a reference clock. I'm currently testing this on a H7 dual-core for which support is also not merged yet. It would certainly make my git workflow easier not having to deal with 3 feature branches at once :wink:

PR #855, opened 2022-05-01
Fix Windows CI and fibers on Windows

chris-durand:
- [x] Update runner OS to Server 2022
- [x] Update GCC to 10.3
- [x] Fix fiber context switch crash

### DIFF
--- a/.github/workflows/windows_armcortexm.yml
+++ b/.github/workflows/windows_armcortexm.yml
@@ -4,7 +4,7 @@ on: [pull_request]
 
 jobs:
   windows_armcortexm:
-    runs-on: windows-2019
+    runs-on: windows-2022
     env:
       PYTHONIOENCODING: "utf-8"
 
--- a/.github/workflows/windows_hosted.yml
+++ b/.github/workflows/windows_hosted.yml
@@ -4,7 +4,7 @@ on: [pull_request]
 
 jobs:
   windows_hosted:
-    runs-on: windows-2019
+    runs-on: windows-2022
     env:
       PYTHONIOENCODING: "utf-8"
 
--- a/.github/workflows/windows_hosted.yml
+++ b/.github/workflows/windows_hosted.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Download GCC for Windows
         run: |
           $ProgressPreference = 'SilentlyContinue'
-          Invoke-WebRequest -OutFile winlibs-gcc.zip https://github.com/brechtsanders/winlibs_mingw/releases/download/10.2.0-11.0.0-8.0.0-r5/winlibs-x86_64-posix-seh-gcc-10.2.0-mingw-w64-8.0.0-r5.zip
+          Invoke-WebRequest -OutFile winlibs-gcc.zip https://github.com/brechtsanders/winlibs_mingw/releases/download/10.3.0-12.0.0-9.0.0-r2/winlibs-x86_64-posix-seh-gcc-10.3.0-mingw-w64-9.0.0-r2.zip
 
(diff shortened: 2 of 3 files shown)


chris-durand:
The same problems are still there after the update.

salkinium:
Is it something about our windows only compile flags? Maybe something got deprecated? Maybe the runtime changed? `scons run` is calling it wrong on Windows?

chris-durand:
> Is it something about our windows only compile flags? Maybe something got deprecated? Maybe the runtime changed? `scons run` is calling it wrong on Windows?

It suddenly started to fail with the same GCC version so I don't think it's related to compiler flags.

I am currently setting up a modm development environment on a real Windows 10 machine to see if that will run the unit tests successfully.

salkinium:
If you're motivated, could you check if `pip install gdbgui` v0.15 is working on Windows? They [removed support in v0.14](https://github.com/cs01/gdbgui/releases/tag/v0.14.0.0), but didn't mention it in the v0.15 release notes.
https://modm.io/guide/installation/#windows

chris-durand:
I can reproduce that the unit test binary crashes for `develop` as well as the `2022q1` release.

There is something else going on with the build system. Compiling the hosted examples with `examples_compile.py` like in the CI seems to do the right thing, but running `lbuild build; scons` from one of the example directories fails with a compile error:

```
Compiling C++·· release\modm\src\modm\io\iostream_printf.o
In file included from modm\src\modm\io\iostream_printf.cpp:27:
modm\ext/printf/printf.source:580:12: error: conflicting declaration of C function 'int _vsnprintf(out_fct_type, char*, size_t, const char*, va_list)'
  580 | static int _vsnprintf(out_fct_type out, char* buffer, const size_t maxlen, const char* format, va_list va)
      |            ^~~~~~~~~~
In file included from modm\src\modm\io\iostream_printf.cpp:12:
c:\winlibs-gcc\mingw64\x86_64-w64-mingw32\include\stdio.h:768:23: note: previous declaration 'int _vsnprintf(char*, size_t, const char*, va_list)'
  768 |   _CRTIMP int __cdecl _vsnprintf(char * __restrict__ _Dest,size_t _Count,const char * __restrict__ _Format,va_list _Args) __MINGW_ATTRIB_DEPRECATED_SEC_WARN;
      |                       ^~~~~~~~~~
```

chris-durand:
`lbuild -D:target=hosted-windows build` fixed the compile error. The examples specify linux as the target in project.xml.

chris-durand:
The crash is caused by the fiber test. If that is disabled everything passes. The fiber example also crashes on Windows.

salkinium:
Well… due to lack of access, I never tested the Windows fiber code on a real machine, only via proxy by running on the CI. So maybe something in GitHub Actions virtualization changed and the instructions work differently now or something stupid like that.

chris-durand:
> If you're motivated, could you check if `pip install gdbgui` v0.15 is working on Windows? They [removed support in v0.14](https://github.com/cs01/gdbgui/releases/tag/v0.14.0.0), but didn't mention it in the v0.15 release notes. https://modm.io/guide/installation/#windows

When you run it there is a message that Windows is not supported right now from `v0.14` on.

chris-durand:
The `movaps` instruction during the context switch fails because of insufficient alignment. I quickly fixed it by replacing it with `movups`. I guess nobody really cares about optimal performance of fibers on Windows. If someone does they can optimize it later.

rleh (approved):
Nice, thanks!

salkinium (approved):
Awesome, thanks!
(rebase?)

PR #856, opened 2022-05-01
[stm32] Uart: Add transmission complete flag

strongly-typed:
Just a quick expose the TC flag in the API.

TXE goes high when the shift register started pouring the bytes to the line and a new byte can be written to the data register.
The TC flag goes high after the stop bit is send. Required for manually enabling and disabling line drivers.

### DIFF
--- a/src/modm/platform/uart/stm32/uart_hal.hpp.in
+++ b/src/modm/platform/uart/stm32/uart_hal.hpp.in
@@ -118,6 +118,10 @@ public:
 	static inline bool
 	isTransmitRegisterEmpty();
 
+	/// Returns true if the transmission of a frame containing data is complete
+	static inline bool
+	isTransmissionComplete();
+
 	static inline void
 	enableInterruptVector(bool enable, uint32_t priority);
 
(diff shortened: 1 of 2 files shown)


salkinium:
Ohai, long time, no PR!
Do you want check if there is something similiar for AVR, just for consistency?

salkinium (approved):

strongly-typed:
Hoi, yes the first public PR for a long time. `modm` is in heavy use here in Switzerland!

e.g. ATMEGA32 `UCSRA Bit 6 – TXC: USART Transmit Complete`
I'll look into it but have no HW to test.

Shall it become part of the common API?

salkinium:
> Shall it become part of the common API?

No, I just thought it would be neat to have feature parity while we're here. It's not a priority though.

strongly-typed:
Just checked, there is e.g. no `isTransmitRegisterEmpty` in current modm's UART API as there is no UART HAL for AVRs (yet). So I would not add this unless we add the HAL concept to AVR. Agree?

salkinium:
Good point, I completely forgot about the AVR HAL…

rleh (approved):

chris-durand (approved):

PR #857, opened 2022-05-02
Add clear() to queue and use isNotEmpty

strongly-typed:


src/modm/architecture/driver/atomic/queue_impl.hpp
```diff
@@ -122,4 +122,12 @@ modm::atomic::Queue<T, N>::pop()
 	this->tail = tmptail;
 }
 
+template<typename T, std::size_t N>
+void
+modm::atomic::Queue<T, N>::clear()
+{
+	this->head = 0;
+	this->tail = 0;
```
salkinium, line 130: This is not interrupt safe though?

salkinium, line 130: This is not interrupt safe though?

strongly-typed: Uhmm, would a `modm::atomic::Lock l;` improve the situation?

salkinium: Yes, but only if clear is called from outside an interrupt. If clear is called inside an interrupt, there's no guarantee that the code outside the interrupt won't write these members back with their `tmp` values.
The atomic queue is not truly atomic, it's merely a single-producer, single-consumer queue that works one sided with non-nested interrupts. It's really very AVR specific, and should be rewritten completely for Cortex-M (or better, imported from an existing project).

chris-durand: The single-producer, single-consumer queue in boost lockfree also has [a non-thread safe `reset()`](https://github.com/boostorg/lockfree/blob/af906f5745e1d6c024db0b4919b56de7f2f9b70b/include/boost/lockfree/spsc_queue.hpp#L350). It is basically equivalent to the proposed `clear()` method. 

The proposed uses in the UART interrupts would not work in that case. Not sure if it is a good idea to add it. It could easily be misused.

salkinium:
I'm closing this PR and instead added to my TODO list a task to find a good queue library with a number of different use-cases on Cortex-M/AVR.

PR #859, opened 2022-05-03
proposal to add SlaveModeTrigger::Internal0 on F4

lukh:


chris-durand (approved):
Thanks. The change looks good. From the STM32F4 reference manual:
trigger
The bits are the same for all timer instances that have an `SMCR` register.

salkinium (approved):
I rebased and reformulated the commit message.

PR #861, opened 2022-05-08
[board] Add Can1 and Can2 to Board::SystemClock

strongly-typed:
Is there a specific reason I do not know why this was left out?

There might be other board examples where the Can is accidentally left out.


salkinium (approved):

salkinium:
It's all just a copy-paste mess, that's why it's missing.

PR #862, opened 2022-05-09
[board] Update information about board revisions of F469NI Disco.Kits

se-bi:
Board Rev-2 has the same touchscreen address as Rev-1.

But let me double-check that today/tomorrow :relaxed:


salkinium (approved):
Yup, can confirm in hardware

PR #865, opened 2022-05-18
[driver] use std::has_single_bit to assert power of 2

TomSaw:
Have discovered `std::has_single_bit` and cause it's exactly, whats required in `modm::BitBangEncoderInput` it must have been fixed for reference reasons.


salkinium (approved):
Cool, didn't know about `std::has_single_bit` either…

PR #866, opened 2022-05-31
fix modm:platform:itm

xgzeng:
Itm not work if buffer.tx option is not set.
Itm::write_itm assumes data is stored in highest byte.


salkinium (approved):
Thanks a lot!

PR #871, opened 2022-06-07
fix: remove default value in setFlashLatency implementation template …

lukh:
Default value in template param Core_mV is both in declaration and implementation, doesn't work :)


salkinium (approved):
Ooops. I wonder why this compiled all this time…

lukh:
Yeah I was wondering was well :)

PR #872, opened 2022-06-07
added filter::MovingAverage::reset(), maintenance of code

TomSaw:
Hope you see the potential.
I also have maintained the code slightly.

### TODO
- [x]  Add test for `filter::MovingAverage::reset()`


salkinium (commented):
I like it!

TomSaw:
Test is in the box - ready to merge if you are.

src/modm/math/filter/moving_average.hpp
```diff
@@ -76,7 +74,7 @@ namespace modm
 			getValue() const;
 
 		private:
-			Index index;
+			least_uint<N> index{0};
```
salkinium: Nice!

src/modm/math/filter/moving_average.hpp
```diff
@@ -85,16 +83,21 @@ namespace modm
...
+}
+
+// ----------------------------------------------------------------------------
+template<typename T, std::size_t N>
+void
+modm::filter::MovingAverage<T, N>::reset(const T& input)
+{
+	std::fill(std::begin(buffer), std::end(buffer), input);
+	sum = N * input;
 }
```
salkinium: Only if you're motivated: Do you think this function and the constructor can be marked as `constexpr`?
Then the sum and buffer may be placed in pure `.data` and copied during startup for static allocation, rather than being constructed at runtime (still required for dynamic allocation of course).

salkinium: I also think it's faster if you copy the argument for this ie. just `T` not `const T&`. On Cortex-M this forces the argument to be pushed onto the stack and then a pointer is passed due to calling convention, which is slower even for a 64-bit double.

src/modm/math/filter/moving_average_float_impl.hpp
```diff
@@ -53,31 +50,30 @@ namespace modm
...
 		index = 0;
-	}
-
-	sum = 0;
-	for (Index i = 0; i < N; ++i) {
-		sum+= buffer[i];
-	}
 
+	sum = std::accumulate(std::begin(buffer), std::end(buffer), 0);
 }
```
salkinium: If I see this correctly, the only difference is the sum computation, but can't this be specialized inside the function with a `if constexpr`? That would save a lot of duplicated code.

src/modm/math/filter/moving_average_float_impl.hpp
```diff
@@ -12,28 +13,24 @@
...
+#include "moving_average.hpp"
 
-#ifndef MODM_MOVING_AVERAGE_HPP
-	#error	"Don't include this file directly, use 'moving_average.hpp' instead!"
-#endif
+#include <numeric>
 
 namespace modm
 {
 	namespace filter{
```
salkinium: While we're touching this file:
```cpp
namespace modm::filter
{
```
and unindent the code?

src/modm/math/filter/moving_average_float_impl.hpp
TomSaw: Yea come on, then let's also inline the method definitions and spare even more lines!

src/modm/math/filter/moving_average.hpp
TomSaw: The **p**ass **b**y **c**opy **pbc** VS  **p**ass **b**y **r**eference **pbr** topic still confuses me

- In templates T may be small or big. thus calls for **pbc** or **pbr**
- STL uses **pbr** all around. I've once benchmarked on ATmega: std::min(T&, T&) VS modm::min(T&, T&) VS modm::min(T, T) and found the STL-**pbr** variant being the fastest with little exceptions. So i concluded to myself, the compiler is smart enough to optimize **pbr**declarations in beneficial cases.

Wish i would be better trained in tools like _compiler explorer_ to control, if and how things optimize :/

TomSaw: In the case of `MovingAverage` 99% T falls in the category `pbc` so i've corrected it.

src/modm/math/filter/moving_average.hpp
```diff
@@ -15,111 +16,102 @@
...
 
-			/// Append new value
-			void
-			update(const T& input);
+#include <modm/math/utils/integer_traits.hpp>
 
-			/// Get filtered value
-			const T
-			getValue() const;
+namespace modm::filter{
```
salkinium:
```suggestion
namespace modm::filter
{
```

src/modm/math/filter/moving_average.hpp
```diff
@@ -15,111 +16,102 @@
...
+	/// Next call of getValue() returns 'input'
+	constexpr void reset(T input)
+	{
+		std::fill(std::begin(buffer), std::end(buffer), input);
+		sum = N * input;
+	}
 
-	buffer[index] = input;
+	/// Append new value
+	void
```
salkinium:
```suggestion
	constexpr void
```
unless the compiler complains…

src/modm/math/filter/moving_average.hpp
```diff
@@ -15,111 +16,102 @@
...
 
-// -----------------------------------------------------------------------------
+	/// Append new value, constrained for float
+	void
+	update(T input)
+	requires std::floating_point<T>
+	{
+		buffer[index] = input;
 
+		sum = std::accumulate(std::begin(buffer), std::end(buffer), 0);
```
salkinium: I was thinking something like this in one function:

```cpp
if constexpr(std::floating_point<T>) {
    buffer[index] = input;
    sum = std::accumulate(std::begin(buffer), std::end(buffer), 0);
} else {
    sum -= buffer[index];
    sum += input;
    buffer[index] = input;
}

File: test/modm/math/filter/moving_average_test.cpp [hunk 8]
```diff
@@ -117,3 +131,17 @@ MovingAverageTest::testFloatAverage()
...
+{
+	modm::filter::MovingAverage<TestDataFloat::Type, 4> filter;
+
+	for (uint_fast8_t i = 0; i < 4; ++i)
+	{
+		filter.update(dataF[i].input);
+	}
+
+	filter.reset(42.23);
+	TEST_ASSERT_EQUALS(filter.getValue(), 42.23);
```
salkinium:
```suggestion
	TEST_ASSERT_EQUALS_DELTA(filter.getValue(), 42.23, double(1e-4));
```
to make sure this works with lower resolution 32-bit floats too, rather than the 80-bit doubles or whatever the x86 CI uses.

TomSaw: When to use `TEST_ASSERT_EQUALS_FLOAT` rather than `TEST_ASSERT_EQUALS_DELTA`?

salkinium: Oh, in this case it should probably be `TEST_ASSERT_EQUALS_FLOAT`, I just copied the line from the check above it, since I forgot the details of the unit test checks again…

src/modm/math/filter/moving_average.hpp
TomSaw: Works

test/modm/math/filter/moving_average_test.cpp
salkinium: The `TEST_ASSERT_EQUALS_FLOAT` has a fixed epsilon, while with `TEST_ASSERT_EQUALS_DELTA` the epsilon can be specified in case your computations are more numerically unstable and yield higher differences.

salkinium (approved):
Awesome!

(the submodules got updated by accident, but the rest is fine)

test/modm/math/filter/moving_average_test.cpp
TomSaw: Got it. Thanks for your always rich explains!

TomSaw: TEST_ASSERT_EQUALS_DELTA is kinda bandpass.

salkinium (approved):

PR #886, opened 2022-07-15
[ext] Upgrade printf implementation

salkinium:
The mpaland version has been stagnant, so the project moved to other people.
The new version is much nicer to work with, in particular, we can make the printf implementation its own module now and people can use it independently from IOStream to also hook global `printf()` calls.
On AVR, we use the faster ASM implementations for formatting integers and floats by default, otherwise the (very large) printf implementation is used.

- [x] Module documentation
- [x] Adapt IOStream documentation
- [x] Example for hooking global `print()` without IOStream
- [x] Example on AVR
- [x] Unit tests
- [x] Fix bug in Makefile outpath relocation and custom flag files

### DIFF
--- a/examples/stm32f469_discovery/printf/main.cpp
+++ b/examples/stm32f469_discovery/printf/main.cpp
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2022, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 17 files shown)


examples/stm32f469_discovery/printf/main.cpp
```diff
@@ -0,0 +1,47 @@
...
+		counter++;
+		printf("Counter %3d\n", counter);
+	}
+	return 0;
+}
+
+
+
+
+
```
chris-durand:
```suggestion
```

Too many newlines at the end of the file.

chris-durand (approved):
Nice!

PR #900, opened 2022-08-18
Add Thing Plus RP2040 Board

cocasema:


src/modm/board/thingplus_rp2040/board.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <modm/platform.hpp>
+
+using namespace modm::platform;
+
+namespace Board
+{
+using namespace modm::literals;
+
+/// RP2040 running at 125MHz generated from the external 12MHz crystal
+/// @ingroup modm_board_feather_rp2040
```
salkinium: Nope 😜

salkinium (commented):
Nice!

src/modm/board/thingplus_rp2040/module.lb
```diff
@@ -0,0 +1,55 @@
...
+with flexible digital interfaces featuring the Raspberry Pi Foundation's RP2040 microcontroller.
+Besides the Thing Plus or Feather footprint (with 18 GPIO pins),
+the board also includes an SD card slot, 16MB (128Mbit) flash memory,
+a JST single cell battery connector (with a charging circuit and fuel gauge sensor),
+an addressable WS2812 RGB LED, JTAG PTH pins, four (4-40 screw) mounting holes,
+and SparkFun's Qwiic connector.
+
+https://www.sparkfun.com/products/17745
+
+See `modm:rp-pico` for programming instructions.
```
rleh, line 29:
```suggestion
See `modm:board:rp-pico` for programming instructions.
```

rleh: It doesn't matter because both [`modm:board:rp-pico`](https://modm.io/reference/module/modm-board-rp-pico/) and [`modm:rp-pico`](https://modm.io/reference/config/modm-rp-pico/) share the same documentation.

src/modm/board/thingplus_rp2040/module.lb
```diff
@@ -0,0 +1,55 @@
...
+    env.substitutions = {
+        "with_logger": False,
+        "with_assert": env.has_module(":architecture:assert")
+    }
+    env.template("../board.cpp.in", "board.cpp")
+    env.copy('.')
+
+    env.outbasepath = "modm/openocd/modm/board/"
+    env.copy(repopath("tools/openocd/modm/rp2040_picoprobe.cfg"), "rp2040_picoprobe.cfg")
+    env.collect(":build:openocd.source", "modm/board/rp2040_picoprobe.cfg")
```
rleh, line 55: What about moving the programmer configuration into its own module, since it is not part of the actual devboard?

The RP2040-Picoprobe module can still be included in the blink example in `project.xml`.

(This should of course also apply to all existing (STM32) Blue-Pill/Black-Pill/... Dev-Board without integrated programmer, but that's out of the scope of this pull request.)

salkinium: Not super convinved, since it's just a single file, so seems weird to have it's own module for it.
Maybe it would make more if we had some "debug" module that's more separate from the build module, but that's kinda weird too.
If you write your own non-BSP based application, then you need to copy this file manually anyways?

src/modm/board/thingplus_rp2040/board.hpp
cocasema: :)

src/modm/board/thingplus_rp2040/module.lb
rleh: I always found it annoying in the past with boards like the Blue-Pill to have a different programmer (e.g. stlink v2 vs. stlink v3) than the one assumed in the BSP for no reason and then have to manually patch around in the generated code each time.

> it's just a single file, so seems weird to have it's own module for it.

I do not see any disadvantage?
Probably the module should also add the logging functionality to the board (if the programmer has a integrated USB VCP channel) and then document the necessary wiring.

>  Maybe it would make more if we had some "debug" module that's more separate from the build module, but that's kinda weird too.

[`modm:debug` module](https://modm.io/reference/module/modm-debug/) exists, and contains `modm::log::Logger`. Why is it called `debug`?

> If you write your own non-BSP based application, then you need to copy this file manually anyways?

Or just include the programmer module, if that is a module?

salkinium: > have to manually patch around in the generated code each time.

Yeah, that's really annoying. But does specifying the programmer in the `openocd.cfg` file  not work as documented in the Blue Pill docs?

> Probably the module should also add the logging functionality to the board (if the programmer has a integrated USB VCP channel) and then document the necessary wiring.

Ok, I like that.

> [modm:debug module](https://modm.io/reference/module/modm-debug/) exists, and contains modm::log::Logger. Why is it called debug?

Good old historic reasons of course. I would move it to `modm:io:logger`.

> Or just include the programmer module, if that is a module?

Sure, but if the BSP already includes the programmer module, then you cannot uninclude it again. So you're back to square one?

salkinium: I'm not opposed to this at all btw, I think it's a really good idea to separate the programmer out, it just needs to be done well as per usual ;-P

PR #901, opened 2022-08-20
[rp2040] Fix infinite loop in SPI due to uint8 overflow

cocasema:
... when trying to send big chunk of data.


rleh (approved):

salkinium (approved):
😱

chris-durand (approved):
:scream_cat:

PR #905, opened 2022-08-27
[driver] Add option for measurement resolution to DS18B20 driver + fix MacOS CI

chris-durand:
Add option to configure measurement resolution to DS18B20 temperature sensor driver. Tested in hardware.

### DIFF
--- a/src/modm/driver/temperature/ds18b20.hpp
+++ b/src/modm/driver/temperature/ds18b20.hpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2010-2011, Fabian Greif
  * Copyright (c) 2012-2014, Niklas Hauser
+ * Copyright (c) 2022, Christopher Durand
  *
  * This file is part of the modm project.
  *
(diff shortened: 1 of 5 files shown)


src/modm/driver/temperature/ds18b20.hpp
```diff
@@ -17,15 +18,25 @@
 
 namespace modm
 {
+	struct ds18b20
```
salkinium, line 32:
```suggestion
	/// \ingroup	modm_driver_ds18b20
	struct ds18b20
```

salkinium (commented):
Nice!

src/modm/driver/temperature/ds18b20.hpp
chris-durand: Fixed. I have also added documentation for the enum and added the `configure()` method to the existing example that it gets instantiated when compiling the driver in the CI.

salkinium (approved):
Thanks!

chris-durand:
@salkinium The make build of Mac OS hosted examples is still broken:
```
Building: linux/build_info/main.cpp with Make
Compiling C···· release/modm/ext/printf/printf.o/bin/sh: gcc-10: command not found
make: *** [build/make-release/modm/ext/printf/printf.o] Error 127
```

The make generator has gcc-10 hard-coded for Mac OS hosted, but the pipeline only has 11:
```
%% if platform == "hosted"
        UNAME_S := $(shell uname -s)
        ifeq ($(UNAME_S),Darwin)
                C_SUFFIX := -10
        # Using homebrew include and lib paths on macOS
        CPPDEFINES += -I/usr/local/include
        LIBPATH += -L/usr/local/lib
        endif
%% endif
```

Can I change this to 11 or do you want to fix this properly?

salkinium:
> Can I change this to 11 or do you want to fix this properly?

You can change it to 11, very few people if any use modm on macOS with make anyways.

PR #909, opened 2022-09-03
[build] Search for GCC >=10 on macOS, including 12

strongly-typed:
Add 12 to the list of possible GCC versions


salkinium:
Ah yes, it's that time again…
I think the Make generator also needs to be [fixed here](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/make/resources/compiler.mk.in#L21).

PR #910, opened 2022-09-05
Add wsl2 docs

daixtrose:
🎉 🎆 

I successfully got modm.io up and running within WSL2 and Ubuntu-22.04 LTS. It seems the USB bridge using [usbipd](https://github.com/dorssel/usbipd-win) finally works on Windows 10.

I would like to share my experience and add another section dedicated to the combination of Windows and WSL2. I hope you like it.

Any comments or requests for improvement are welcome.

### DIFF
--- a/docs/src/guide/installation.md
+++ b/docs/src/guide/installation.md
@@ -317,7 +317,7 @@ avrdude --version
 
 For a pure Windows installation see [above](#windows).
 
-The [Windows Subsystem for Linux 2](https://docs.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2) allows you to run a Linux distribution in parallel to Windows. In order to use this Linux for developing you need to [bridge the USB hub to which the device is attached](https://docs.microsoft.com/en-us/windows/wsl/connect-usb).
+The [Windows Subsystem for Linux 2][wsl2] allows you to run a Linux distribution in parallel to Windows. In order to use this Linux for developing you need to [bridge the USB hub to which the device is attached](https://docs.microsoft.com/en-us/windows/wsl/connect-usb).
 
 - Install Ubuntu-22.04 LTS as WSL2 distribution
   - See the [general WSL doc](https://docs.microsoft.com/en-us/windows/wsl/install), but
@@ -416,4 +416,4 @@ The [Windows Subsystem for Linux 2](https://docs.microsoft.com/en-us/windows/wsl
 [wingit]: https://git-scm.com/download/win
 [winterm]: https://github.com/Microsoft/Terminal
 [winwsl]: https://docs.microsoft.com/en-us/windows/wsl/about
-    
+[wsl2]: https://docs.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2    
--- a/docs/src/guide/installation.md
+++ b/docs/src/guide/installation.md
@@ -285,8 +285,26 @@ The [Windows Subsystem for Linux 2](https://docs.microsoft.com/en-us/windows/wsl
 	``` 
   - In WSL2 Ubuntu-22.04 configure the USB ports as user accessible. This requires the configuration of `udev` rules. 
     - Use an editor to create a new `udev` rules file with a high enough number in the filenam, e.g. `sudo nano /etc/udev/rules.d/100-udev-stm.rules`
-    - Add some rules similar to [the ones used in OpenOCD](https://github.com/openocd-org/openocd/blob/master/contrib/60-openocd.rules#L84-L99)
-    - Either reboot the machine or be lucky that a combination of following commands is sufficient 
+    - Add some rules similar to [the ones used in OpenOCD](https://github.com/openocd-org/openocd/blob/master/contrib/60-openocd.rules#L84-L99), e.g.
+      ```
+	  # STMicroelectronics ST-LINK V1
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="3744", MODE="660", GROUP="plugdev", TAG+="uaccess"
+
+	  # STMicroelectronics ST-LINK/V2
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="3748", MODE="660", GROUP="plugdev", TAG+="uaccess"
+
+	  # STMicroelectronics ST-LINK/V2.1
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="374b", MODE="660", GROUP="plugdev", TAG+="uaccess"
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="3752", MODE="660", GROUP="plugdev", TAG+="uaccess"
+
+	  # STMicroelectronics STLINK-V3
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="374d", MODE="660", GROUP="plugdev", TAG+="uaccess"
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="374e", MODE="660", GROUP="plugdev", TAG+="uaccess"
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="374f", MODE="660", GROUP="plugdev", TAG+="uaccess"
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="3753", MODE="660", GROUP="plugdev", TAG+="uaccess"
+	  ATTRS{idVendor}=="0483", ATTRS{idProduct}=="3754", MODE="660", GROUP="plugdev", TAG+="uaccess"
+      ```
+    - Either reboot the machine or be lucky that a combination of the following commands is sufficient 
       ```bash 
 	  sudo udevadm trigger
 	  sudo udevadm control --reload
--- a/docs/src/guide/installation.md
+++ b/docs/src/guide/installation.md
@@ -421,7 +421,7 @@ The [Windows Subsystem for Linux 2][wsl2] allows you to run a Linux distribution
 [winterm]: https://github.com/Microsoft/Terminal
 [winwsl]: https://docs.microsoft.com/en-us/windows/wsl/about
 [wsl2]: https://docs.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2    
-[wsl-connect-usb]: https://docs.microsoft.com/en-us/windows/wsl/connect-usb)
+[wsl-connect-usb]: https://docs.microsoft.com/en-us/windows/wsl/connect-usb
 [wsl-install]: https://docs.microsoft.com/en-us/windows/wsl/install
 [wsl-vscode]: https://docs.microsoft.com/en-us/windows/wsl/tutorials/wsl-vscode
 [wsl-vscode-integrated-terminal]: https://code.visualstudio.com/docs/remote/wsl-tutorial#_integrated-terminal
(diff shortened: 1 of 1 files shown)


docs/src/guide/installation.md
```diff
@@ -1,3 +1,4 @@
+
```
rleh: Nope.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
 Please use the free and open-source [7-Zip file archiver][7_zip] to extract the
 files in the next steps.
 
+## Windows with WSL2
```
rleh, line 315: This Win+WSL2 section currently interrupts of the Windows section (starting at L208).
The whole added section belongs to the end of this file (of before the windows section, to be discussed).

Also there should be references at both the _Windows_ and _Windows+WSL2_ section to each other.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
...
+
+The [Windows Subsystem for Linux 2](https://docs.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2) allows you to run a Linux distribution in parallel to Windows. In order to use this Linux for developing you need to [bridge the USB hub to which the device is attached](https://docs.microsoft.com/en-us/windows/wsl/connect-usb).
+
+- Install Ubuntu-22.04 LTS as WSL2 distribution
+  - See the [general WSL doc](https://docs.microsoft.com/en-us/windows/wsl/install), but
+  - as of this writing `wsl --list --online` will not list Ubuntu 22.04.1 LTS, but it is [available in the Microsoft Store](https://www.microsoft.com/store/productId/9PN20MSR04DW). Install it from there via mouse clicks
```
rleh: Afaik the minor version of Ubuntu is not important, so please remove this line to redice the maintainance effort to keep this up-to-date.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
...
+- Install [Visual Studio Code](https://code.visualstudio.com/) and the [remote extensions for WSL](https://docs.microsoft.com/en-us/windows/wsl/tutorials/wsl-vscode) and make sure you can log in to your linux distribution
```
rleh: We do not not suggest using a specific IDE, that is a choice of the user.

The user just has to somehow access his WSL2 system.

docs/src/guide/installation.md
rleh, line 315: This Win+WSL2 section currently interrupts of the Windows section (starting at L208).
The whole added section belongs to the end of this file (of before the windows section, to be discussed).

Also there should be references at both the _Windows_ and _Windows+WSL2_ section to each other.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
 Please use the free and open-source [7-Zip file archiver][7_zip] to extract the
 files in the next steps.
 
+## Windows with WSL2
+
+The [Windows Subsystem for Linux 2](https://docs.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2) allows you to run a Linux distribution in parallel to Windows. In order to use this Linux for developing you need to [bridge the USB hub to which the device is attached](https://docs.microsoft.com/en-us/windows/wsl/connect-usb).
+
+- Install Ubuntu-22.04 LTS as WSL2 distribution
```
rleh, line 321: Please use Ubuntu 20.04 since that is the reference linux distribution which is also used in Linux section above.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
...
+  - See the [general WSL doc](https://docs.microsoft.com/en-us/windows/wsl/install), but
+  - as of this writing `wsl --list --online` will not list Ubuntu 22.04.1 LTS, but it is [available in the Microsoft Store](https://www.microsoft.com/store/productId/9PN20MSR04DW). Install it from there via mouse clicks
+- Install [Visual Studio Code](https://code.visualstudio.com/) and the [remote extensions for WSL](https://docs.microsoft.com/en-us/windows/wsl/tutorials/wsl-vscode) and make sure you can log in to your linux distribution
+- Physically attach the microcontroller board to a USB port
```
rleh, line 325: Not part of the installation.
I think the best place is to add a Windows+WSL section [here](https://github.com/modm-io/modm/blob/develop/examples/README.md).

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
...
+	  sudo udevadm trigger
+	  sudo udevadm control --reload
+	  ```
+	- After adding the new `udev` rules the usb device has to be virtually detached and reattached to the Linux system via PowerShell:
+	  ```powershell
+	  usbipd wsl detach --busid 3-3
+	  usbipd wsl attach --busid 3-3
+      ``` 
+	  Alternatively remove the USB connector, reattach it, then issue the command `usbipd wsl attach --busid 3-3`
+  - On WSL2 Linux follow the [Linux installation instructions](#linux). Hint: They are made for 20.04, but work on 22.04 as well.
```
rleh, line 381: Regrading the _Hint_:
See comment above.
The instructions will be upgraded together with the CI to Ubuntu 22.04 probably before the next release.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
...
+  - In addition perform the following installation steps
+    ```bash
+	sudo apt install picocom libncurses5
```
rleh: 1. The choice of the terminal program is up to the user.
2. Please move `libncurses5` to the linux section above. It is always a dependency of GDB, but under Ubuntu Desktop it is probably already installed by default, which is why the problem was never noticed before.

docs/src/guide/installation.md
```diff
@@ -339,3 +413,4 @@ avrdude --version
 [wingit]: https://git-scm.com/download/win
 [winterm]: https://github.com/Microsoft/Terminal
 [winwsl]: https://docs.microsoft.com/en-us/windows/wsl/about
+
```
rleh: Nope.

docs/src/guide/installation.md
```diff
@@ -250,6 +251,79 @@ Download and run the [Installer][doxypress_binaries].
...
+    ```powershell
+	PS C:\Windows\system32> usbipd wsl detach --busid 3-3
+	PS C:\Windows\system32> usbipd wsl attach --busid 3-3
+    usbipd: info: Using default distribution 'Ubuntu-22.04'.
+	``` 
+  - In WSL2 Ubuntu-22.04 configure the USB ports as user accessible. This requires the configuration of `udev` rules. 
+    - Use an editor to create a new `udev` rules file with a high enough number in the filenam, e.g. `sudo nano /etc/udev/rules.d/100-udev-stm.rules`
+    - Add some rules similar to [the ones used in OpenOCD](https://github.com/openocd-org/openocd/blob/master/contrib/60-openocd.rules#L84-L99), e.g.
+      ```
+	  # STMicroelectronics ST-LINK V1
```
rleh: 1. There should be no need to create any additional custom udev rules since [Ubuntus OpenOCD package contains](https://packages.ubuntu.com/jammy/amd64/openocd/filelist) the udev rule file `/lib/udev/rules.d/60-openocd.rules`. 
2. modm supports not just STM32, so (if necessary, see above) just copy the whole file.

rleh (commented):
Nice, thank you!

Is it possible to add some Windows+WSL checks to the CI, to detect potential problems with Windows+WSL2 early enough in the future?

docs/src/guide/installation.md
daixtrose: I moved the section to the end of the file

docs/src/guide/installation.md
daixtrose: Not very sure if this works with Ubuntu 20.04

docs/src/guide/installation.md
daixtrose: Removed

docs/src/guide/installation.md
daixtrose: > We do not not suggest using a specific IDE, that is a choice of the user.

Although I agree with this statement in general it does not apply here IMHO. If you are totally new to developing on Windows then the hint to VSCode will spare you a lot of time. I do not want to cancel VSCode just because it is evil in the eyes of some.

docs/src/guide/installation.md
daixtrose: moved below

docs/src/guide/installation.md
daixtrose: I begin to understand why I have hard times digging through modm docs ... if you have to click into many different places to collect all relevant information just to have things up and running it's no fun and will people turn away for no good reason. 

IMHO it **is** a relevant information because people might not guess that this has to be done before issuing the other commands and fail on the bridge.

docs/src/guide/installation.md
daixtrose: Under Linux a text file is [required to have a newline at the end](https://stackoverflow.com/a/729795). Some tools may fail otherwise

docs/src/guide/installation.md
daixtrose: > 1. There should be no need to create any additional custom udev rules since [Ubuntus OpenOCD package contains](https://packages.ubuntu.com/jammy/amd64/openocd/filelist) the udev rule file `/lib/udev/rules.d/60-openocd.rules`.

Weird enough it only started to work after adding it under `/etc/udev/`. Maybe @joelsa has an explanation?

daixtrose: > 2\. modm supports not just STM32, so (if necessary, see above) just copy the whole file.

The file is rather long. Do you want me to copy the whole content into the text?

docs/src/guide/installation.md
daixtrose: > 2\. Please move `libncurses5` to the linux section above. It is always a dependency of GDB, but under Ubuntu Desktop it is probably already installed by default, which is why the problem was never noticed before.

Moved libncruses5 to the Linux section

daixtrose: > 1. The choice of the terminal program is up to the user.

Yes, but how and where do you want to put the `picocom` or other terminal software hint into your docs?

docs/src/guide/installation.md
daixtrose: I can't. The last time I tried the USB bridge did not work.

docs/src/guide/installation.md
daixtrose: Also see above. I can only confirm it working on 22.04.

docs/src/guide/installation.md
daixtrose: > > 1. There should be no need to create any additional custom udev rules since [Ubuntus OpenOCD package contains](https://packages.ubuntu.com/jammy/amd64/openocd/filelist) the udev rule file `/lib/udev/rules.d/60-openocd.rules`.
> 
> Weird enough it only started to work after adding it under `/etc/udev/`. Maybe @joelsa has an explanation?

I have a gut feeling that the OpenOCD rules get overwritten due to their low number `60-openocd.rules`. Is there a way to automatically find out? 

The file `/etc/udev/rules.d/70-snap.snapd.rules`  also seems to be a concatenation:

```sh
# This file is automatically generated.

# Concatenation of all ModemManager udev rules
# do not edit this file, it will be overwritten on update
```

~~Are files in `/lib/udev/rules.d/` considered in general?~~ (https://askubuntu.com/a/44922)

salkinium:
Reformatted this in https://github.com/modm-io/modm/pull/919.

PR #912, opened 2022-09-10
Use STM32 Timer channel signal in methods

strongly-typed:
Adding on @chris-durand example in #908

The `advanced_extended` property still looks a bit rough to me. That information should come from modm-devices ... I had to disable some device families to get all tests run. I would leave it as it is and add better props later.

### DIFF
--- a/src/modm/platform/timer/stm32/module.lb
+++ b/src/modm/platform/timer/stm32/module.lb
@@ -110,8 +110,8 @@ def prepare(module, options):
     props["target"] = target = device.identifier
     props["types"] = set()
     # extended advanced control timer with 6 channels and trigger out 2
-    props["advanced_extended"] = target.family not in ["f0", "f1", "f2", "f4", "l0", "l1"] or \
-            target.family == "f3" and target.name not in ["73", "78"]
+    props["advanced_extended"] = (target.family not in ["f0", "f1", "f2", "f3", "f4", "l0", "l1"]) or \
+            ((target.family == "f3") and (target.name not in ["73", "78"]))
 
     props["timer_vectors"] = tim_vectors = defaultdict(list)
     vectors = [v["name"] for v in device.get_driver("core")["vector"] if "TIM" in v["name"]]
--- a/examples/blue_pill_f103/servo_pwm/main.cpp
+++ b/examples/blue_pill_f103/servo_pwm/main.cpp
@@ -50,7 +50,7 @@ int main()
 	Timer4::setPrescaler(22);
 	Timer4::setOverflow(overflow);
 
-	Timer4::configureOutputChannel(2, Timer4::OutputCompareMode::Pwm, 0);
+	Timer4::configureOutputChannel<GpioB7::Ch2>(Timer4::OutputCompareMode::Pwm, 0);
 
 	Timer4::applyAndReset();
 
@@ -64,11 +64,11 @@ int main()
 	MODM_LOG_INFO << "1.0 ms pulse width for 1 s" << modm::endl;
 
 	LedGreen::set();
-	Timer4::configureOutputChannel(2, Timer4::OutputCompareMode::Pwm, minPwm);
+	Timer4::configureOutputChannel<GpioB7::Ch2>(Timer4::OutputCompareMode::Pwm, minPwm);
 	modm::delay(2s);
-	Timer4::configureOutputChannel(2, Timer4::OutputCompareMode::Pwm, maxPwm);
+	Timer4::configureOutputChannel<GpioB7::Ch2>(Timer4::OutputCompareMode::Pwm, maxPwm);
 	modm::delay(2s);
-	Timer4::configureOutputChannel(2, Timer4::OutputCompareMode::Pwm, minPwm);
+	Timer4::configureOutputChannel<GpioB7::Ch2>(Timer4::OutputCompareMode::Pwm, minPwm);
 	LedGreen::reset();
 	modm::delay(1s);
 
@@ -76,7 +76,7 @@ int main()
 	MODM_LOG_INFO << modm::endl;
 
 	MODM_LOG_INFO << "Pulse width set to 1.5 ms" << modm::endl;
-	Timer4::setCompareValue(2, targetPwm);
+	Timer4::setCompareValue<GpioB7::Ch2>(targetPwm);
 
 	while (true)
 	{
--- a/examples/nucleo_g474re/servo_pwm/main.cpp
+++ b/examples/nucleo_g474re/servo_pwm/main.cpp
@@ -37,7 +37,7 @@ main()
 	Timer4::setPrescaler(52);
 	Timer4::setOverflow(overflow);
 
-	Timer4::configureOutputChannel(2, Timer4::OutputCompareMode::Pwm, 0);
+	Timer4::configureOutputChannel<GpioB7::Ch2>(Timer4::OutputCompareMode::Pwm, 0);
 
 	Timer4::applyAndReset();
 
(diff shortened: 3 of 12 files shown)


strongly-typed:
@salkinium I would love to set the `ci:hal` label so the CI tests all devices.

src/modm/platform/timer/stm32/module.lb
```diff
@@ -110,8 +110,8 @@ def prepare(module, options):
     props["target"] = target = device.identifier
     props["types"] = set()
     # extended advanced control timer with 6 channels and trigger out 2
-    props["advanced_extended"] = target.family not in ["f0", "f1", "f2", "f4", "l0", "l1"] or \
-            target.family == "f3" and target.name not in ["73", "78"]
+    props["advanced_extended"] = \
+            (target.family == "f3" and target.name not in ["03", "73", "78"])
```
chris-durand: Nope. This breaks my production code :wink:. The advanced control timer implementation with 6 channels is present in all F7, G0, G4, H7, L4, L5, U5, WB and WL devices as well as all F3 except 373 and 378.

If you remove the `advanced extended` property user code using timer channels 5 and 6 will still compile and access wrong places in the peripheral memory when trying to configure those channels. That would be very bad there, for instance:

```
	static inline void
	setCompareValue(uint32_t channel, Value value)
	{
%% if advanced_extended
		if(channel <= 4) {
			*(&TIM{{ id }}->CCR1 + (channel - 1)) = value;
		} else if(channel == 5) {
			TIM{{ id }}->CCR5 = value;
		} else if(channel == 6) {
			TIM{{ id }}->CCR6 = value;
		}
%% else
		*(&TIM{{ id }}->CCR1 + (channel - 1)) = value;
%% endif
	}
```

chris-durand (changes requested):

src/modm/platform/timer/stm32/module.lb
chris-durand: From looking at the ST headers the correct condition should be:
```suggestion
    props["advanced_extended"] = target.family not in ["f0", "f1", "f2", "f3", "f4", "l0", "l1"] and \
            (target.family != "f3" or target.name not in ["73", "78"])
```

If there is a compile error for F303, then all what we call `advanced_extended` is not actually identical.

salkinium:
> I would love to set the ci:hal label so the CI tests all devices.

I've given you triage rights for this repo, so you should be able to set issues/pr labels and milestones yourself.

src/modm/platform/timer/stm32/module.lb
strongly-typed: ... but GPIO `base.hpp` is missing `Ch5` and `Ch6` in many cases:

```
/// @cond
	enum class
	Signal
	{
[...]
		Ch1,
		Ch1n,
		Ch2,
		Ch2n,
		Ch3,
		Ch3n,
		Ch4,
		Ch4n,
		Cha1,
		Cha2,
		Chb1,
[...]
```

e.g. for `examples/nucleo_g474re/servo_pwm`
```
Compiling C++·· release/modm/src/modm/platform/timer/timer_4.o
In file included from modm/src/modm/platform/timer/timer_4.hpp:23,
                 from modm/src/modm/platform/timer/timer_4.cpp:17:
modm/src/modm/platform/timer/general_purpose_base.hpp: In static member function 'static consteval int modm::platform::GeneralPurposeTimer::signalToChannel()':
modm/src/modm/platform/timer/general_purpose_base.hpp:272:56: error: 'Ch5' is not a member of 'modm::platform::Gpio::Signal'; did you mean 'Ch1'?
  272 |   } else if constexpr (Signal::Signal == Gpio::Signal::Ch5) {
```

chris-durand: The CI compile error is caused by the absence of `Gpio::Signal::Ch5` and `Gpio::Signal::Ch6`. Channels 5 and 6 can't be connected to IO pins on any STM32, so there are no GPIO signals for it. That was my mistake. Just remove the following lines in `signalToChannel()` and everything should work:

```
%% if advanced_extended
		} else if constexpr (Signal::Signal == Gpio::Signal::Ch5) {
			return 5;
		} else if constexpr (Signal::Signal == Gpio::Signal::Ch6) {
			return 6;
%% endif
```

strongly-typed: > Channels 5 and 6 can't be connected to IO pins on any STM32, so there are no GPIO signals for it.

I was just studying the datasheets and came to the same conclusion. No worries, just fixed it.

chris-durand:
I am sorry, I messed it up again :facepalm:. The correct condition for "advanced_extended" is:
```
    props["advanced_extended"] = target.family not in ["f0", "f1", "f2", "f3", "f4", "l0", "l1"] or \
            (target.family == "f3" and target.name not in ["73", "78"])
```
Otherwise "advanced_extended" will never be true on F3.

chris-durand (approved):
Thank you!

strongly-typed:
The 14-pin versions of [STM32L011D3P6](https://github.com/modm-io/modm/runs/8287421849?check_suite_focus=true#step:5:39) fail in the CI as they only have 3 channels.

What is the best way to distinguish these parts?

Something like?
```
    props["advanced_extended"] = ((target.family not in ["f0", "f1", "f2", "f3", "f4", "l0", "l1"]) and (target.package != 'd')) or \
            ((target.family == "f3") and (target.name not in ["73", "78"]))
```

chris-durand:
> The 14-pin versions of STM32L011D3P6 fail in the CI as they only have 3 channels.
>
> What is the best way to distinguish these parts?

`advanced_extended` is not the issue here, the condition is false already for `L0`. The devices have only one general purpose and one basic timer.

A hacky but easy solution would be guarding the `Ch4` if-branch with a condition to exclude them:

```
%% if target.family != "l0" or target.package != "d"
		} else if constexpr (Signal::Signal == Gpio::Signal::Ch4) {
			return 4;
%% endif
```

Or you determine from the gpio data which signals exist. You would need to add a query to the gpio module to get the signals from the timer driver and only generate code in `signalToChannel()` for the existing channels.

I am fine with the easy one. If ST released more tiny devices with missing timer channels someone would have to touch that condition again. But that should not be a big deal.

strongly-typed:
> I am fine with the easy one. If ST released more tiny devices with missing timer channels someone would have to touch that condition again. But that should not be a big deal.

Let's go with the easy solution first and see where it takes us.

I would like to ask if you could share any example code that uses `Ch5` and `Ch6` in the CI? Currently, this is all unused and further changes are not checked by the CI.

chris-durand:
> Let's go with the easy solution first and see where it takes us.

After testing it, I know we won't come far :laughing:. It's `target.pin != "d"`, the package part in `STM32L011D3P6` is the `p`.

chris-durand:
> I would like to ask if you could share any example code that uses `Ch5` and `Ch6` in the CI? Currently, this is all unused and further changes are not checked by the CI.

I am actually planning to add an example once I have implemented the ADC external trigger function in the F3/G4/H7/... ADC driver. It currently would not add much value to the CI. The code is not much different to using channels 1-4. You only pass a different number to `configureOutputChannel()` and that is a runtime parameter not checked during compilation:
```
Motor::PwmTimer::configureOutputChannel(5, PwmMode, CurrentSamplingOffset);
```
Then it configures TRGO2 to output the channel 5 compare signal (by calling `setMode()`). That signal triggers the ADCs.

Right now I am writing ADC registers manually, but it should not be too hard to add that API to the modm driver for all devices (if you can parse the list of triggers from the reference manuals, but I have heard @salkinium has a solution to that problem :smiley:).

chris-durand:
I will push the fix with `target.pin != "d"` instead of `target.package != "d"` to your branch and restart the compile all job.

strongly-typed:
The CI ran fine. So if there aren't any objections: LGTM. Thank you all for your good support and great collaboration.

salkinium (approved):
Will merge tonight, currently traveling.

PR #914, opened 2022-09-15
fixed filter moving average for floating type

Artiom9:
std::accumulate<T>(first, last, init) deduces type T based on init argument. Therefore, for floating types, you need to pass the type float.

### DIFF
--- a/src/modm/math/filter/moving_average.hpp
+++ b/src/modm/math/filter/moving_average.hpp
@@ -82,7 +82,7 @@ public:
 	{
 		if constexpr(std::floating_point<T>) {
 			buffer[index] = input;
-			sum = std::accumulate(std::begin(buffer), std::end(buffer), 0);
+			sum = std::accumulate(std::begin(buffer), std::end(buffer), T{0});
 		} else {
 			sum -= buffer[index];
 			sum += input;


src/modm/math/filter/moving_average.hpp
```diff
@@ -82,7 +82,7 @@ class MovingAverage
 	{
 		if constexpr(std::floating_point<T>) {
 			buffer[index] = input;
-			sum = std::accumulate(std::begin(buffer), std::end(buffer), 0);
+			sum = std::accumulate(std::begin(buffer), std::end(buffer), 0.0);
```
salkinium:
```suggestion
			sum = std::accumulate(std::begin(buffer), std::end(buffer), 0.0f);
```
Perhaps even this to prevent double deduction?

chris-durand:
```suggestion
			sum = std::accumulate(std::begin(buffer), std::end(buffer), T{0});
```

On some targets modm compiles with `-fsingle-precision-constant`. That's why 0.0 is interpreted as a `float` constant instead of `double`. On other platforms with hardware 64-bit float support that behaviour is not enabled and the type would be deduced as `double` with `0.0`. `T{0}` should do the right thing in all circumstances.

chris-durand: Hi @Artiom9,

you have marked the conversation as resolved but did not push the review suggestion. Could you do that or should I? Afterwards we would merge the PR.

Artiom9: Hi, @chris-durand! To be honest, I do not really understand the Githab and do not really understand the procedure of the pull request. I just wanted to point out the problem and provide my solution. If some other actions are needed, do it for me. Thank you.

chris-durand: Thank you very much for contributing the fix! I have squashed the commits and we will merge it soon.

chris-durand (approved):

chris-durand:
@salkinium merge?

salkinium (approved):
Thanks!

PR #925, opened 2022-11-06
Fix MacOS CI

chris-durand:
Compilation of MacOS hosted examples fails due to a [toolchain bug in Xcode 14.0](https://stackoverflow.com/questions/73714336/xcode-update-to-version-2395-ld-compile-problem-occurs-computedatomcount-m). Update Xcode to 14.1 to fix the issue.

The linker crashes with:
```
ld: Assertion failed: (_file->_atomsArrayCount == computedAtomCount &&
 "more atoms allocated than expected"), function parse,
 file macho_relocatable_file.cpp, line 2061.ld: Assertion failed: (_file->_atomsArrayCount == computedAtomCount && "more atoms allocated than expected"), function parse, file macho_relocatable_file.cpp, line 2061.
```

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -7,6 +7,11 @@ jobs:
     runs-on: macos-12
 
     steps:
+      # update Xcode version to fix linker bug
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: '14.1'
+
       - name: Setup environment - Brew tap
         run: |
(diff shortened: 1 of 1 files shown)


rleh:
The linux/avr-examples job fail looks like some temporary network issues with the CI runner and/or PyPI.
I restarted the job,

rleh (approved):

salkinium:
Thanks for fixing this!

PR #929, opened 2022-11-24
[ui] Fix uninitialized frame buffer memory in MonochromeGraphicDisplay

chris-durand:
Before:
display1

After:
display2


salkinium (approved):
Yes

PR #933, opened 2022-12-02
[stm32] Only configure BDTR register for timers with complementary PWM

### LINKED ISSUE #932: Stm32G4: Timer2-5 bug
According to reference manual, Timers 2..5 of Stm32G4 family do not have BDTR register. modm generates functions using this register.
Although the .svd and hardware-description files indeed have definition for it, it is not used.

chris-durand:
Only provide the timer HAL functions to configure the BDTR register if it is present in hardware. Previously the functions were generated for all general purpose timers although only instances 15, 16, 17 and the advanced control timers support it.

@ser-plu

### DIFF
--- a/examples/blue_pill_f103/servo_pwm/main.cpp
+++ b/examples/blue_pill_f103/servo_pwm/main.cpp
@@ -55,7 +55,6 @@ int main()
 	Timer4::applyAndReset();
 
 	Timer4::start();
-	Timer4::enableOutput();
 
 	MODM_LOG_INFO << modm::endl;
 	MODM_LOG_INFO << "Teaching sequence for electronic speed controllers:" << modm::endl;
--- a/examples/nucleo_g474re/servo_pwm/main.cpp
+++ b/examples/nucleo_g474re/servo_pwm/main.cpp
@@ -42,7 +42,6 @@ main()
 	Timer4::applyAndReset();
 
 	Timer4::start();
-	Timer4::enableOutput();
 
 	MODM_LOG_INFO << "Teaching sequence for electronic speed controllers:" << modm::endl;
 	MODM_LOG_INFO << "1.0 ms pulse width for 2 s" << modm::endl;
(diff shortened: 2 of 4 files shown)


ser-plu:
I think, more functions are required. Something like this:
```c++
        static inline void
	enableOutput(uint32_t channel)
	{
		TIM2->CCER |= TIM_CCER_CC1E << (4 * (channel-1));
	}


	template<typename Signal>
	static inline void
	enableOutput()
	{
		constexpr auto channel = signalToChannel<Peripheral::Tim2, Signal>();
		enableOutput(channel);
	}

	static inline void
	disableOutput(uint32_t channel)
	{
		TIM2->CCER &= ~(TIM_CCER_CC1E << (4 * (channel-1)));
	}


	template<typename Signal>
	static inline void
	disableOutput()
	{
		constexpr auto channel = signalToChannel<Peripheral::Tim2, Signal>();
		disableOutput(channel);
	}
```

I'm not sure, that the corrected examples in the commit will actually output PWM without enabling  it in CCER

chris-durand:
> I'm not sure, that the corrected examples in the commit will actually output PWM without enabling it in CCER

The respective bits in `CCER` are set inside of `configureOutputChannel()`, see the driver code [here](https://github.com/modm-io/modm/blob/14dbb428bbcc2e461b6a90700f9462e1b88895bf/src/modm/platform/timer/stm32/general_purpose.cpp.in#L196).

The `PinState` parameter controls whether the `CCxE` bit gets set and is `PinState::Enable` by default:
```
template<typename Signal>
static void
configureOutputChannel(OutputCompareMode_t mode,
                       Value compareValue, PinState out = PinState::Enable,
                       bool enableComparePreload = true)
```

chris-durand:
I have tried the G474 `servo_pwm` example on a Nucleo L552ZE-Q (I don't have a G4 Nucleo at hand right now) and it is outputting PWM:

photo_2022-12-03_01-15-58

ser-plu:
Yep, true.
________________________________

ser-plu:
Nice! Thanks for the fast feedback!

________________________________

salkinium (approved):
Thanks!

PR #936, opened 2022-12-26
[ext] Update all submodules to latest versions

salkinium:
This updates all submodules in preparation for the 2022q4 release.

- [x] CMSIS-DSP v1.14.2: was split off from the CMSIS_5 repository into its own repo and this adds that repo and updates it to the latest version. Only very minor changes were required, so I don't expect anything to break.
- [x] FreeRTOS 202212: Only minor config adaptation.
- [x] FatFS v0.15: Only minor config adaptation.
- [x] TinyUSB v0.14.0: Only minor config adaptation.
- [x] Printf v6.1.0: Only minor config adaptation.
- [x] ARM GCC 11.3 and 12.2 sometimes links against Newlib syscalls in big projects (FreeRTOS+TCP pulls in `_gettimeofday` somehow, LVGL all file syscalls). I think this is an issue with the linker garbage collection, since no source code calls any of these functions directly. I've added weak empty stub implementations for now.
- [ ] The upgrade of the SAM header files leads to issues with redefinitions.
- [x] All other submodules updated without issues.

Tested in hardware:

- [x] CMSIS-DSP on NUCLEO-F429ZI
- [x] TinyUSB on RP2040
- [x] FatFs + TinyUSB on BlackPill F411
- [x] LVGL on DISCO-F469I
- [x] ETL on NUCLEO-F429ZI
- [x] NanoPB on NUCLEO-F429ZI


salkinium:
Looks like the macOS CI is broken again??? Seems to fail when calling `which scons`. Ugh.

chris-durand:
> Could @chris-durand or @rleh test the CMSIS-DSP examples on NUCLEO-F429ZI until the 22Q4 release?

I also don't have access to any STM32 hardware right now, but accidentally took a SAMV71 board with me. I will try running the tests on that one.

chris-durand:
The tests have all passed on the SAMV71 board.

salkinium:
> The tests have all passed on the SAMV71 board.

Fantastic, thanks!
I'm currently updating the rest of the submodules, and I'll have my hardware again soon, so I can test the rest too.

salkinium:
I've changed this PR to also update all other submodules. @chris-durand the update of the SAM header files leads to issues with redefinitions etc. I think you already mentioned that to me some time ago, so I would suggest to fix it later if it's too much of a hassle for now.

I'll update the CI to GCC11 next.

salkinium:
> I'll update the CI to GCC11 next.

I'm not going to do this, since I cannot test avr-gcc 12 since it doesn't compile on macOS 13, because of a linker bug?
So I'm not going to update the compiler for this release and instead just continue to use GCC 10.

salkinium:
The CI is compromised: The `examples_compile.py` script uses multiprocessing, which since Python 3.8 uses `spawn` instead of `fork` to create processes, and thus something breaks. I'm experiencing a pretty bad lockup on macOS 13 with Python 3.10 and the CI is just silently failing to build at all. This is very bad, I'm trying to fix it by setting it back to use `fork`. Also wtf.

salkinium:
I don't have a fix for the multiprocessing issue, perhaps updating Python from 3.8 to a newer version will fix it.
I also don't have a fix for the GCC12 linker issue, but that will be a problem for when we upgrade the CI to v12.

PR #937, opened 2023-01-01
2022q4 release

salkinium:
- [x] Update release notes
- [x] Submodules and Hardware testing done in #936
- [ ] mac CI needs to be fixed seperately (no clue yet what exactly breaks)
- [ ] CI images need to be upgraded to GCC 12 separately
- [ ] GCC 12.2 linker failing to link due to `project.elf has a LOAD segment with RWX permissions`, which can be disabled with `-Wl,--no-warn-rwx-segment`, but that fails GCC 11.3 and earlier with `unrecognized option '--no-warn-rwx-segment'`. Not sure how to fix that in a robust way yet.
- [ ] CI is **silently** not building all examples due to issues with Python multiprocessing! Could not fix it yet.


salkinium:
Just a reminder: @rleh Please copy the docs to a new release folder tomorrow, when the CI is finished uploading them.

PR #941, opened 2023-01-12
[sam] Add support for GpioInverted

chris-durand:


salkinium (approved):
Is there a BSP that could be updated with it?

chris-durand:
I could add a trivial hardware unit test that sets a pin and then checks if it got set to an inverted state. Not sure if there would be a good use in any BSP right now.

salkinium:
Nah, I was just wondering. Rebase pls.

PR #944, opened 2023-01-16
[stm32] Add PLL disable function

### LINKED ISSUE #943: Disabling PLL on STM32
If I need to change PLL settings (i need it when switching from a bootloader to the main application), I need to change system clock to Internal source and to disable PLL. There is a function for the first part, but the second required something like this:

```c++
bool
Rcc::disablePll(uint32_t waitCycles) {
	RCC->CR &= RCC_CR_PLLON;
	while (RCC->CR & RCC_CR_PLLRDY) and --waitCycles)
		;
}
```

chris-durand:
Resolves #943

@ser-plu


ser-plu:
Yes, the feature is working for me. Thank you for the quick response!

salkinium (approved):
👍

PR #950, opened 2023-02-01
[driver] IMU LSM6DSO

rleh:
- [x] Driver
- [x] Example
- [x] Test on real hardware (it works ✅)

### DIFF
--- a/examples/nucleo_g474re/imu_lsm6dso/main.cpp
+++ b/examples/nucleo_g474re/imu_lsm6dso/main.cpp
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2023, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/platform/spi/spi_master_1.hpp>
+#include <modm/driver/inertial/lsm6dso.hpp>
+#include <optional>
+
+using namespace Board;
+
+using SpiMaster = SpiMaster1;
+using Mosi = GpioA7;
+using Miso = GpioA6;
+using Sck = GpioA5;
+using Cs = GpioC5;
+using Int1 = GpioC4;
+using ImuVdd = GpioA10;
+
+using Lsm6dso = modm::Lsm6dso<modm::Lsm6TransportSpi<SpiMaster, Cs>>;
+
+Lsm6dso imu{};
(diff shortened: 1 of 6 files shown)


salkinium (approved):
Yes please

examples/nucleo_g474re/imu_lsm6dso/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+	// Give the sensor some time to boot...
+	modm::delay(10ms);
+
+	MODM_LOG_INFO << "Resetting and rebooting imu (BOOT | SW_RESET)..." << modm::endl;
+	// Write bits BOOT | SW_RESET (and IF_INC, because it default to 1) to reboot imu
+	if(!RF_CALL_BLOCKING(imu.writeRegister(modm::lsm6dso::Register::CTRL3_C, 0b1000'0101))) {
+		MODM_LOG_INFO.printf(" ... failed.");
+		Board::Leds::set();
+	}
```
chris-durand, line 62: Shouldn't this be a named function in the driver or done in `initialize()` if the reboot is required?

src/modm/driver/inertial/lsm6dso.hpp
```diff
@@ -0,0 +1,288 @@
...
+class Lsm6dso : public lsm6dso, public Transport
+{
+public:
+	/**
+	 * @brief Constructor
+	 *
+	 * For I2c this also sets the address to 0x6a (alternative: 0x6b with pin
+	 * SA0 high).
+	 */
+	Lsm6dso(uint8_t address=0x6a);
```
chris-durand:
```suggestion
	explicit Lsm6dso(uint8_t address=0x6a);
```

[C.46: By default, declare single-argument constructors explicit](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit)

chris-durand (commented):
Thanks, very nice!

examples/nucleo_g474re/imu_lsm6dso/main.cpp
rleh: There are no intructions in STs documentation about how to use this bit or how to reset/reboot the IMU.
Also removing this code block does not change the behavior of IMU in this example, mainly because the IMU is resetted by removing the power supply just a few lines above.

I kept the lines in this example because in some (future) applications someone might need to reset the IMU configuration...

chris-durand: Ok, thanks for the clarification.

rleh: So you are okay with keeping this as it is?

chris-durand: Yes

chris-durand (approved):

PR #953, opened 2023-02-02
[docs] Fix order of release dropdown on https://docs.modm.io/

rleh:
I just found this commit laying around from last year...


chris-durand (approved):
Thanks!

salkinium (approved):

PR #960, opened 2023-02-14
add openocd config for nucleo h723 (single bank flash)

mat-kie:
The STM32H723 only has a single bank flash so the open-ocd config for the Nucleo-H743 (which has dual bank flash) that is shipped with open-ocd can not be used.

### DIFF
--- a/tools/openocd/modm/st_nucleo_h723zg.cfg
+++ b/tools/openocd/modm/st_nucleo_h723zg.cfg
@@ -1,0 +1,10 @@
+# This is an ST NUCLEO-H723ZG board with single STM32H723ZG chip.
+# http://www.st.com/en/evaluation-tools/nucleo-h723zg.html
+
+source [find interface/stlink.cfg]
+
+transport select hla_swd
+
+source [find target/stm32h7x.cfg]
+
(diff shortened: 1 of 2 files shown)


tools/openocd/modm/st_nucleo_h723zg.cfg
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
```
rleh: Shouldn't this rather be MPL-v2 or no license at all?

rleh (commented):
Thank you for your pull request!

tools/openocd/modm/st_nucleo_h723zg.cfg
mat-kie: yes, this was a copy paste error from my pull request to the open ocd repo.

rleh (approved):
Thanks!

Would you like to squash your commits into a single one (and remove the leading two empty lines from `st_nucleo_h723zg.cfg`)?

salkinium (approved):
Yup, thanks!

PR #965, opened 2023-02-25
[tools] Refactor tooling and add JLink support

salkinium:
Cleaned up `modm_tools` and the Scons wrappers for it a little, and added JLink support.
For our purposes OpenOCD is still better and much more configurable, however, JLink tends to have more device coverage and adds new devices faster. It can also be faster for certain tasks like RTT and uploading.

- [x] Refactored modm_tools:
  - [x] Added missing gdbinit functions for BMP and JLink
  - [x] Moved SCons logging wrapper into openocd wrapper where it belongs
  - [x] Added SCons `env.SubstList()` to listify, aggregate and substitute environment variables
  - [x] Using a contextmanager for the backend to simplify code
  - [x] Simplify SCons with `env.AlwaysBuildAction`
  - [x] Split log into `rtt.py` and `itm.py` and clean up options.
  - [x] Only `compare-sections` when using GDB in a UI mode
  - [x] Properly call the code as a Python module to remove the `sys.path` hacks.
- [x] Added JLink support
  - [x] Programming
  - [x] Debugging via GDB
  - [x] Reset
  - [x] RTT
  - [x] ITM via SWO
  - [x] Documentation
    - [x] Build
    - [x] Scons
    - [x] Make
    - [x] Cmake
  - [x] Build Systems Support
    - [x] SCons
    - [x] Make
    - [x] CMake


salkinium:
Ok, this PR got a little out of hand. But I wanted to clean up `modm_tools` anyways in preparation for pulling it external and making it its own python module… Anyways, JLink is running not as stable as OpenOCD at least on macOS, kinda dissappointing tbh. I only tested the STLink->Jlink replacement and the JLink Mini EDU, perhaps the more expensive probes are better. Meh.

rleh:
When you write `jlink`, you mean the combination of one of the _J-Link (Edu/Base/Plus/Mini/Classic/...)_ debug probes (hardware device with USB connector from Segger) and the _J-Link Tools_ (proprietary software blob from Segger), right?

The J-Link (or compatible, cheaply available as _H-Link_/...) USB dongle works fine with OpenOCD and supports a very large number of devices, e.g. Atmel/Microchip SAM.

examples/blue_pill_f103/itm/main.cpp
```diff
@@ -22,7 +22,7 @@ modm::IOStream stream(itm_device);
 /*
  * Blinks the green user LED with 1 Hz and outputs the loop counter over ITM.
  * SWO pin is PB3. openocd > 0.11.0 is required. Receive data with:
- * scons log-itm fcpu=72000000
+ * scons itm fcpu=72000000
```
rleh: Is it really necessary to (again) rename the scons commands?

salkinium: No, and I just remembered that I named these commands `log-*` on purpose to make clear that they are somewhat one directional (although the `log-rtt` allows you to send back).

salkinium:
> When you write jlink, you mean the combination of one of the J-Link (Edu/Base/Plus/Mini/Classic/...) debug probes (hardware device with USB connector from Segger) and the J-Link Tools (proprietary software blob from Segger), right?

Yes, I use modm_tools for work and added jlink support, and am now contributing this back upstream again. J-Link generally has support for newer devices earlier than OpenOCD, otherwise I very much prefer OpenOCD, since it's way more configurable and also more stable than the J-Link proprietary software, at least on macOS.

I need to check if RTT performance is better as they claim, I vaguely remember that OpenOCD was only able to transfer single-digit kB/s range, because they drive everything from the USB Host, whereas J-Link should do it directly on the probe.

salkinium:
Ok, so I've refactored a little more and the Makefile and CMake generator are now calling modm_tools via the module syntax, ie. `python3 -m modm_tools.openocd [args]`. This is in preparation for moving these tools into the modm package, that we currently only use for dependency grouping. It will then change to `python3 -m modm.openocd [args]`, and there will be actual Python documentation rather than the Markdown hack that we have right now (tools.modm.io?).

PR #967, opened 2023-03-06
[board] [stm32] Nucleo F103RB enable PB4

### LINKED ISSUE #966: STM32F103 PB4 not accessible / remapping not part of modm
Hi there,

ST decided to use `PB4` as `D5` on their [F103RB nucleo](https://github.com/modm-io/modm/blob/develop/src/modm/board/nucleo64_arduino.hpp#L29).

Unfortunately, `PB4` is by default mapped as `NJTRST`, see [Table 37 of RM0008](https://www.st.com/resource/en/reference_manual/cd00171190-stm32f101xx-stm32f102xx-stm32f103xx-stm32f105xx-and-stm32f107xx-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf#page177).

I have not found a way to remap PB4 to GPIO by modm.

In an example with [Blue Pill there is a hint in the code](https://github.com/modm-io/modm/blob/develop/examples/blue_pill_f103/encoder_input/main.cpp#L55) to access `AFIO->MAPR` directly, which works.

What should we do about it? Add new API to modm?

strongly-typed:
Resolves #966

### DIFF
--- a/src/modm/board/nucleo_f103rb/board.hpp
+++ b/src/modm/board/nucleo_f103rb/board.hpp
@@ -120,6 +120,9 @@ initialize()
 	stlink::Uart::initialize<SystemClock, 115200_Bd>();
 
 	Button::setInput();
+
+	// Enable PB4 as GPIO
+	AFIO->MAPR |= AFIO_MAPR_SWJ_CFG_JTAGDISABLE;
 }
 /// @}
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Yes, thanks!

PR #968, opened 2023-03-06
C++23 and C23

rleh:
Do we plan to enable C++23 (and C23) for modm in the (near) future?
This would give access to nice new [lanuage and library features](https://en.cppreference.com/w/cpp/compiler_support) even already with GCC12.

- [x] Testing
- [x] Remove GCC10 vs GCC12 build system flags hack

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 <!--/webignore-->
 
 modm (pronounced like dial-up "modem") is a toolbox for
-building custom C++20 libraries tailored to your embedded device.
+building custom C++23 libraries tailored to your embedded device.
 modm generates startup code, HALs and their implementations, communication
 protocols, drivers for external devices, and BSPs in a modular, customizable
 process that you can fine-tune to your needs.
@@ -49,7 +49,7 @@ git clone --recurse-submodules --jobs 8 https://github.com/modm-io/modm.git
 
 ## Features
 
-- Efficient and fast object-oriented C++20 API.
+- Efficient and fast object-oriented C++23 API.
 - Support for thousands of AVR and ARM Cortex-M microcontrollers from Microchip, STMicroelectronics and Raspberry Pi.
 - Build system agnostic: Choose SCons, CMake, Makefile or use your own.
 - Modular, data-driven, target-specific HAL generation using the lbuild code generator.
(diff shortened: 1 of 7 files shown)


salkinium:
Sure, but C++23 is not final yet?

rleh:
Yeah, but we use (only) GCC12 and it already implements a significant proportion of the standard.

chris-durand:
C++23 is considered feature-complete and only small obscure details will change. I'd consider it safe to turn it on. While some bigger features are still missing in GCC 12 you'll still get a lot of nice things. Unless we want to maintain compatibility with GCC 10/11 for some reason but I don't think we have a strong motivation to do that.

salkinium:
OK, then let's enable it after the 23q1 release, so that we/I can test GCC12 on C++20 and then GCC12 on C++23 separately.

salkinium:
If you are motivated, I would also like to do some C++ upgrade work, ie. making everything constexpr (particularly constructors), spaceship operator (?), container iterators are deprecated, maybe more std functions/GCC builtins instead of inline asm, etc.
Doesn't need to happen in this PR, but we'll probably get more mileage out of that than using the lastest C++ standard ;-)

rleh:
> OK, then let's enable it after the 23q1 release, so that we/I can test GCC12 on C++20 and then GCC12 on C++23 separately.

Good idea!
The release is in 3 weeks already 😱

> If you are motivated, I would also like to do some C++ upgrade work, [...]. Doesn't need to happen in this PR, but we'll probably get more mileage out of that than using the lastest C++ standard ;-)

Absolutely

salkinium:
Reminder to self: Remove the GCC10 vs. GCC12 hacks in the build systems again.

salkinium:
@rleh I'd be more happy to merge this PR though ;-P

salkinium (approved):

rleh (commented):
✅ Ready to merge from my side.

I ran the unittests for Nucleo-F429/F767/U575/F091/F103/G474 and SAMV71 without any problems.
(Except for the _fdcan_test_ and _gpio_sampler_test_ on the Nucleo-G474RE, but they also fail on current develop, see #1013.)

salkinium:
Let's do it, the new features are actually quite useful for embedded!

salkinium:
Was there a reason you chose to remove 43b5f36 and 860bae2 from this PR? They're not critical, just useful.

rleh:
Oups, that was unintentional.
It probably happened during the rebase and I did not notice 😿

PR #974, opened 2023-03-13
create method to enable and disbale scan mode

victorandrehc:


victorandrehc:
Sorry this should be on my own fork

PR #976, opened 2023-03-13
create method to enable and disable scan mode

victorandrehc:
This Pull request create enable and disable methods for STM32's adc. Currently there is support for adding multiple channels on the adc but not for enabling scan mode. 
https://github.com/modm-io/modm/discussions/973


salkinium:
Let's see what the CI says to that 🤞. Maybe you're lucky and ST actually made something consistent for once 😛

salkinium (approved):
Nice!

PR #977, opened 2023-03-14
fix scan mode for STM32F3

victorandrehc:
This PR removes enableScanMode and disableScanMode methods and incorporate them into addChannel.

 - [STM32F3](https://www.st.com/resource/en/reference_manual/rm0316-stm32f303xbcde-stm32f303x68-stm32f328x8-stm32f358xc-stm32f398xe-advanced-armbased-mcus-stmicroelectronics.pdf) family doesn't have a hardware Scan mode. Its scan mode is computed via ADCx->SQR1 L bits only. Methods enableScanMode and disableScanMode make little sense then, since the addChannle method already handles it.

### DIFF
--- a/src/modm/platform/adc/stm32/adc.hpp.in
+++ b/src/modm/platform/adc/stm32/adc.hpp.in
@@ -370,18 +370,6 @@ public:
 	static inline void
 	acknowledgeInterruptFlags(const InterruptFlag_t flags);
 
-	/**
-	 * Enables scan mode
-	 */
-	static inline void
-	enableScanMode();
-
-	/**
-	 * Disables scan mode
-	 */
-	static inline void
-	disableScanMode();
-
 private:
 	/**
 	 * Select the frequency of the clock to the ADC. The clock is common
--- a/src/modm/platform/adc/stm32/adc_impl.hpp.in
+++ b/src/modm/platform/adc/stm32/adc_impl.hpp.in
@@ -127,6 +127,11 @@ modm::platform::Adc{{ id }}::addChannel(const Channel channel,
 	}
 	// update channel count
 	ADC{{ per }}->SQR1 = (ADC{{ per }}->SQR1 & ~ADC_SQR1_L) | (channel_count << 20);
+%% if target["family"] not in ["f3"]
+	// set hardware scan mode
+	ADC{{ per }}->CR1 |= ADC_CR1_SCAN;
+
+%%endif
 
 	setSampleTime(channel, sampleTime);
 	return true;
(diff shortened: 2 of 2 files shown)


victorandrehc:
@salkinium I created this PR improving the scan mode that was merged on monday, but it seems like the CI died on me. It made 13/25 tests and never executed the last 12. How can I make it execute the last 12?

salkinium:
The 25 other CI jobs are guarded behind the `ci:hal` tag, since they run quite long and "steal" CI time from other PRs. It's not the best mechanism (you cannot set the tag, only maintainers).
You can however execute the HAL tests locally, see the readme in the tests folder.

salkinium:
I don't really know anything about the ADCs, so I'm delegating this to @rleh ;-P

victorandrehc:
This Pr basically sets the scan flag on register CR1 and remove the enable and disable scan mode methods, because I figured that if one is adding a channel they are probably meaning scan mode, so would be better to just enable scan mode there silently.

rleh:
I'll look into it, give me some (2-3) days...

chris-durand (commented):
Hi @victorandrehc,

the F3 devices in the linked reference manual use a completely different ADC driver found in [`modm/platform/adc/stm32f3`](https://github.com/modm-io/modm/tree/e5a579ae18d1c15bd2d135171f94acd80517783c/src/modm/platform/adc/stm32f3). The driver you changed is for F1, F2, F4, F7, L1 and F37x devices only. The scan mode flag seems to be present in all of them.

I'd be interested what your actual use-case for scan mode is. In my understanding scan mode is only valid when utilizing DMA which the driver does not support yet.

What it does is automatically starting the next conversion after the previous one has been completed. The DMA hardware would be triggered to fetch the conversion result and the next sampling cycle would commence shortly after. It's not recommended to use that mode with interrupts since the latency could be too high and the data register might be invalidated before it is read.

Thus, the F1 reference manual states:

scan

victorandrehc:
Hi @chris-durand, I will change(eliminate) the reference of the F3 then, I thought it was referring to F303 controllers.
About the scan mode, my objective is exactly to support the DMA. I have a local branch doing that already and it would be my next PR. My strategy was to merge it on smaller easier to review pull requests. If you prefer we can close this PR and I can do it all together on a big PR.    
Since you use scan mode just for dma, maybe it makes sense to keep the enable and disable scan mode methods.

chris-durand:
>  Since you use scan mode just for dma, maybe it makes sense to keep the enable and disable scan mode methods.

I would just leave it as is and recommend to close the PR without merging. The scan mode does no harm for now. We can finally sort it when you implement ADC DMA in your next PR.

victorandrehc:
The DMA PR is not right yet. But I will create a draft so the progress can be followed.

PR #993, opened 2023-04-06
[driver] Add MAX31865 driver with STM32F469 Discovery example

hshose:
- [x] Driver
- [x] Example with F469
- [x] Works with Hardware
- [x] README

### DIFF
--- a/examples/stm32f469_discovery/max31865/main.cpp
+++ b/examples/stm32f469_discovery/max31865/main.cpp
@@ -1,0 +1,88 @@
+// coding: utf-8
+/*
+ * Copyright (c) 2023, Henrik Hose
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <modm/board.hpp>
+#include <modm/driver/temperature/max31865.hpp>
+#include <modm/processing.hpp>
+
+using namespace Board;
+
+using SpiMaster = modm::platform::SpiMaster2;
+
+using Cs = D9;
+using Mosi = D11;
+using Miso = D12;
+using Sck = D13;
+
+class ThermocoupleThread : public modm::pt::Protothread
+{
+public:
+	ThermocoupleThread() : data{}, pt100(data) {}
+
+	bool
+	run()
+	{
+		PT_BEGIN();
+		PT_CALL(pt100.initialize());
+
+		while (true)
+		{
+			MODM_LOG_INFO << "\nNew readout:" << modm::endl;
+			PT_CALL(pt100.readout());
+
+			MODM_LOG_INFO << "          resistance : " << data.getResistance() << " Ohm"
+						  << modm::endl;
+			MODM_LOG_INFO << "     temperature fast: " << data.getTemperatureFast() << " degrees"
+						  << modm::endl;
+			MODM_LOG_INFO << "  temperature precise: " << data.getTemperaturePrecise() << " degrees"
+						  << modm::endl;
+
+			timeout.restart(std::chrono::milliseconds(1000));
+			PT_WAIT_UNTIL(timeout.isExpired());
+		}
+
+		PT_END();
+	}
+
+private:
+	using Max31865 = modm::Max31865<SpiMaster, Cs, modm::max31865::pt100>;
+	Max31865::Data data;
+	Max31865 pt100;
+
+	modm::ShortTimeout timeout;
+};
+
+ThermocoupleThread pt100Thread{};
+
+int
+main()
+{
+	Board::initialize();
+	Cs::setOutput(modm::Gpio::High);
+
+	SpiMaster::connect<Miso::Miso, Mosi::Mosi, Sck::Sck>();
+	SpiMaster::initialize<Board::SystemClock, 351_kHz>();
+
+	MODM_LOG_INFO << "==========MAX 31865 Test==========" << modm::endl;
+	MODM_LOG_DEBUG << "Debug logging here" << modm::endl;
+	MODM_LOG_INFO << "Info logging here" << modm::endl;
+	MODM_LOG_WARNING << "Warning logging here" << modm::endl;
+	MODM_LOG_ERROR << "Error logging here" << modm::endl;
+	MODM_LOG_INFO << "===============================" << modm::endl;
+
+	while (true)
+	{
+		pt100Thread.run();
+		Board::LedOrange::toggle();
+	}
+
+	return 0;
+}
--- a/src/modm/driver/temperature/max31865.hpp
+++ b/src/modm/driver/temperature/max31865.hpp
@@ -1,0 +1,230 @@
+// coding: utf-8
+// ----------------------------------------------------------------------------
+/*
+ * Copyright (c) 2023, Henrik Hose
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_MAX31865_HPP
+#define MODM_MAX31865_HPP
+
+#include <array>
+#include <modm/architecture/interface/register.hpp>
+#include <modm/architecture/interface/spi_device.hpp>
+#include <modm/processing/resumable.hpp>
+#include <modm/processing/timer.hpp>
+
+namespace modm
+{
+
+/// @ingroup modm_driver_max31865
+struct max31865
+{
+	struct Pt
+	{
+		float R0;
+		float Rref;
+		float alpha;
+		double a;
+		double b;
+		double c;
+	};
+
+	/// Standard values for IEC 751 (PT100) with 430 Ohm reference
+	static constexpr Pt pt100{
+		.R0 = 100.f,
+		.Rref = 430.f,
+		.alpha = 3.85055e-3f,
+		.a = double{3.90830e-3},
+		.b = double{-5.77500e-7},
+		.c = double{-4.18301e-12},
+	};
+
+	enum class Fault : uint8_t
+	{
+		RtdHighThreshold = Bit7,
+		RtdLowThreshold = Bit6,
+		RefinMGreater85Percentbias = Bit5,
+		RefinMLess85PercentVbias = Bit4,
+		RtdinMLess85PercentVbias = Bit3,
+		OverUnderVoltage = Bit2
+	};
+	MODM_FLAGS8(Fault)
+
+	enum class Config : uint8_t
+	{
+		VBias = Bit7,
+		ConversionModeAuto = Bit6,
+		OneShot = Bit5,
+		ThreeWire = Bit4,
+		FaultDetectionCycle1 = Bit3,
+		FaultDetectionCycle0 = Bit2,
+		FaultStatusClear = Bit1,
+		Rejection = Bit0
+	};
+	MODM_FLAGS8(Config)
+
+	enum class FaultDetectionWrite : uint8_t
+	{
+		NoAction = 0b00,
+		FaultDetectionAutomaticDelay = 0b01,
+		RunFaultDetectionWithManualDelayCycle1 = 0b10,
+		FinishFaultDetectionWithManualDelayCycle2 = 0b11,
+	};
+	using FaultDetectionWrite_t =
+		Configuration<Config_t, FaultDetectionWrite, 0b11, 2>;  // Bit 8..10
+
+	enum class FaultDetectionRead : uint8_t
+	{
+		FaultDetectionFinished = 0b00,
+		AutomaticDetectionStillRunning = 0b01,
+		ManualCycle1Running =
+			0b10,  // waiting for user to write FinishFaultDetectionWithManualDelayCycle2
+		ManualCycle2StillRunning = 0b11,
+	};
+	using FaultDetectionRead_t = Configuration<Config_t, FaultDetectionRead, 0b11, 2>;  // Bit 8..10
+
+	enum class Rejection : uint8_t
+	{
+		Rejection60Hz = 0b0,
+		Rejection50Hz = 0b1,
+	};
+	using Rejection_t = Configuration<Config_t, Rejection, 0b1>;  // Bit 8..10
+
+	static constexpr uint16_t
+	rtdfaultthresh(const uint16_t res)
+	{
+		return res << 1;
+	};
+
+	enum class Register : uint8_t
+	{
+		ReadConfiguration = 0x00,
+		ReadRtdMsb = 0x01,
+		ReadRtdLsb = 0x02,
+		ReadHighFaultThresholdMsb = 0x03,
+		ReadHighFaultThresholdLsb = 0x04,
+		ReadLowFaultThresholdMsb = 0x05,
+		ReadLowFaultThresholdLsb = 0x06,
+		ReadFaultStatus = 0x07,
+
+		WriteConfiguration = 0x80,
+		WriteHighFaultThresholdMsb = 0x83,
+		WriteHighFaultThresholdLsb = 0x84,
+		WriteLowFaultThresholdMsb = 0x85,
+		WriteLowFaultThresholdLsb = 0x86,
+	};
+
+	template<Pt pt>
+	struct modm_packed Data
+	{
+		/// @return measure resistance in ohm
+		constexpr float
+		getResistance() const
+		{
+			const uint16_t adccode = data >> 1;
+			return adccode / 32768.f * pt.Rref;
+		}
+
+		/// @return fast temperature in degrees celsius, about 0.3 degrees error between 0 and 100
+		/// degrees celsius
+		constexpr float
+		getTemperatureFast() const
+		{
+			return (getResistance() - pt.R0) / pt.alpha / 100.f;
+		}
+
+		/// @return slow but accurate temperature in degrees celsius
+		constexpr double
+		getTemperaturePrecise() const
+		{
+			const double res = getResistance();
+			double T = getTemperatureFast();
+			const double c0 = T <= 0 ? pt.c : 0;
+
+			// Do some fixed number of newton steps for root finding
+			// on Callendar Van Dusen equation:
+			// R = R0*(1+a*T+b*T*T+c*(T-100)*T*T*T)
+			// Newton seems to need double precision to achieve 1.e-10 residual?!
+			for (int i = 0; i < 10; i++)
+			{
+				const double R = double{pt.R0} * (1 + (pt.a * T) + (pt.b * T * T) +
+												  (c0 * (T - 100) * T * T * T)) -
+								 res;
+				const double Rdash =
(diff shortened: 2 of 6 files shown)


rleh (commented):
Nice driver!

Please use [tabs](https://github.com/modm-io/modm/blob/develop/docs/coding_convention.md#layout)  instead of spaces.

src/modm/driver/temperature/max31865.hpp
```diff
@@ -0,0 +1,216 @@
...
+			}
+
+			return T;
+		}
+
+		uint16_t data;
+	};
+};  // struct max31865
+
+using max31865pt100 = max31865<>;
```
rleh: Wouldn't it make sense to put the default template arguments (R0, Rref, alpha, a, b, c) into a struct, pass it as a template argument to the `Max31865` class and forward it to the `max31865` struct?
That way the whole code is reusable and the `Max31865` class does not have to be copied to change the values.

examples/stm32f469_discovery/max31855/main.cpp
```diff
@@ -0,0 +1,86 @@
+// coding: utf-8
+/*
+ * Copyright (c) 2022, Sarah Vilete
```
rleh: Nope

rleh (commented):
The driver is (correctly) named _max31865_, but example _max318**5**5_.

src/modm/driver/temperature/max31865.hpp
rleh:
```c++

struct Max31865Config
{
	float R0;
	float Rref;
	float alpha;
	double a;
	double b;
	double c;
};

/// Standard values for IEC 751 (PT100) with 430 Ohm reference
constexpr Max31865Config max31865ConfigPt100
{
	.R0 = 100.f,
	.Rref = 430.f,
	.alpha = 3.85055e-3f,
	.a = double{3.90830e-3},
	.b = double{-5.77500e-7},
	.c= double{-4.18301e-12},
};

template<Max31865Config max31865Config>
struct max31865
{
	static constexpr float R0 = max31865Config.R0;
	static constexpr float Rref = max31865Config.Rref;
	// ...

	// [...]
};

template<typename SpiMaster, typename Cs, Max31865Config max31865Config = max31865ConfigPt100>
class Max31865 : public max31865<max31865Config>,
				 public modm::SpiDevice<SpiMaster>,
				 protected modm::NestedResumable<3>
{
	// [...]
};
```

examples/stm32f469_discovery/max31865/main.cpp
```diff
@@ -0,0 +1,87 @@
...
+	bool
+	run()
+	{
+		PT_BEGIN();
+		pt100.initialize();
+
+		while (true)
+		{
+			MODM_LOG_INFO << "\nNew readout:" << modm::endl;
+			RF_CALL_BLOCKING(pt100.readout());
```
salkinium:
```suggestion
			PT_CALL(pt100.readout());
```

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+Max31865<SpiMaster, Cs>::initialize()
+{
+
+	this->attachConfigurationHandler([]() {
+		static_assert(std::is_same_v<SpiMaster, modm::platform::SpiMaster2>);
+		SPI2->CR1 &= ~SPI_CR1_SPE;
+		__DSB();
+		SpiMaster::setDataMode(SpiMaster::DataMode::Mode3);
+		SPI2->CR1 |= SPI_CR1_SPE;
+		__DSB();
```
salkinium: Should be moved into the SpiMaster code.

salkinium: You can also move the attach call into the constructor.

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+	});
+
+	Cs::setOutput(modm::Gpio::High);
+
+	config = Config();
+	config.set(Config::FaultStatusClear);
+	config.set(Rejection_t(Rejection::Rejection50Hz));
+	RF_CALL_BLOCKING(writeSingleRegister(Register::WriteConfiguration, config.value));
+	config = Config_t(RF_CALL_BLOCKING(readSingleRegister(Register::ReadConfiguration)));
```
salkinium: You can turn this function into an resumable function too.

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+modm::ResumableResult<uint8_t>
+Max31865<SpiMaster, Cs>::readSingleRegister(Register address)
+{
+	RF_BEGIN();
+	RF_WAIT_UNTIL(this->acquireMaster());
+
+	Cs::reset();
+	buffer[0] = uint8_t(address);
+	RF_CALL_BLOCKING(SpiMaster::transfer(buffer.data(), nullptr, 1));
+	RF_CALL_BLOCKING(SpiMaster::transfer(nullptr, buffer.data(), 1));
```
salkinium:
```suggestion
	RF_CALL(SpiMaster::transfer(nullptr, buffer.data(), 1));
```

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+modm::ResumableResult<uint16_t>
+Max31865<SpiMaster, Cs>::readTwoRegisters(Register address)
+{
+	RF_BEGIN();
+	RF_WAIT_UNTIL(this->acquireMaster());
+
+	Cs::reset();
+	buffer[0] = uint8_t(address);
+	RF_CALL_BLOCKING(SpiMaster::transfer(buffer.data(), nullptr, 1));
+	RF_CALL_BLOCKING(SpiMaster::transfer(nullptr, buffer.data(), 2));
```
salkinium:
```suggestion
	RF_CALL(SpiMaster::transfer(nullptr, buffer.data(), 2));
```

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+	if (this->releaseMaster()) { Cs::set(); }
+	modm::delay(100ms);
```
salkinium: ?

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+	RF_BEGIN();
+	RF_WAIT_UNTIL(this->acquireMaster());
+
+	Cs::reset();
+	buffer[0] = uint8_t(address);
+	RF_CALL_BLOCKING(SpiMaster::transfer(buffer.data(), nullptr, 1));
+	RF_CALL_BLOCKING(SpiMaster::transfer(nullptr, buffer.data(), 1));
+
+	if (this->releaseMaster()) { Cs::set(); }
+	modm::delay(100ms);
```
salkinium: ?

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+template<typename SpiMaster, typename Cs>
+modm::ResumableResult<void>
+Max31865<SpiMaster, Cs>::readout()
+{
+	RF_BEGIN();
+	config.set(Config::VBias);
+	RF_CALL(writeSingleRegister(Register::WriteConfiguration, config.value));
+	timeout.restart(std::chrono::milliseconds(10));
```
salkinium:
```suggestion
	timeout.restart(10ms);
```

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+	RF_WAIT_UNTIL(timeout.isExpired());
+
+	config.set(Config::OneShot);
+	RF_CALL(writeSingleRegister(Register::WriteConfiguration, config.value));
+	timeout.restart(std::chrono::milliseconds(65));
```
salkinium:
```suggestion
	timeout.restart(65ms);
```

src/modm/driver/temperature/max31865_impl.hpp
salkinium: At least I think that should work inside the `modm` namespace.

salkinium (changes requested):

src/modm/driver/temperature/max31865_impl.hpp
hshose: definetly

src/modm/driver/temperature/max31865_impl.hpp
hshose: Whupsss that was for debugging... removed it

src/modm/driver/temperature/max31865_impl.hpp
hshose: Whupsss that was for debugging... removed it

src/modm/driver/temperature/max31865_impl.hpp
hshose: #994, thanks @chris-durand

src/modm/driver/temperature/max31865.hpp
hshose: Should work like this.

Is `struct Pt{...}` is confusing in the modm namespace?
Should I call it something like `struct RtdPt{...}` maybe?

hshose:
Thanks @salkinium und @rleh for the review.

Please don't merge yet - I havent tested the fixes with hardware yet.

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,115 @@
...
+Max31865<SpiMaster, Cs, pt>::initialize()
+{
+
+	config = typename max31865<pt>::Config();
+	config.set(max31865<pt>::Config::FaultStatusClear);
+	config.set(typename max31865<pt>::Rejection_t(max31865<pt>::Rejection::Rejection50Hz));
+	RF_CALL_BLOCKING(writeSingleRegister(max31865<pt>::Register::WriteConfiguration, config.value));
+	config = typename max31865<pt>::Config_t(
+		RF_CALL_BLOCKING(readSingleRegister(max31865<pt>::Register::ReadConfiguration)));
+}
```
rleh: You need to use `RE_BEGIN();`/`RF_END();` in a resumable function, and then replace `RF_CALL_BLOCKING` by `RF_CALL`.

```suggestion
template<typename SpiMaster, typename Cs, Pt pt>
modm::ResumableResult<void>
Max31865<SpiMaster, Cs, pt>::initialize()
{
	RF_BEGIN();
	config = Config();
	config.set(Config::FaultStatusClear);
	config.set(Rejection_t(Rejection::Rejection50Hz));
	RF_CALL(writeSingleRegister(Register::WriteConfiguration, config.value));
	config = Config_t(RF_CALL(readSingleRegister(Register::ReadConfiguration)));
	RF_END();
}
```

src/modm/platform/core/cortex/module.lb
```diff
@@ -369,5 +369,4 @@ def build(env):
         env.collect(":build:archflags", "-mfloat-abi={}".format(env["float-abi"]), fpu_spec)
         single_precision = ("-sp-" in fpu_spec)
     if single_precision:
-        env.collect(":build:ccflags", "-fsingle-precision-constant",
-                                      "-Wdouble-promotion")
+        env.collect(":build:ccflags", "-Wdouble-promotion")
```
salkinium: This might make the calculation even more precise at the cost of using a software math library instead of the FPU, since constants are not forced to be `float` instead of `double`. (The STM32F4 FPU is single precision only).

rleh:
@hshose I pushed some commits to your branch, because it was too confusing to explain this in code comments.
Feel free to delete the commits from your branch if you don't like them.

src/modm/platform/core/cortex/module.lb
rleh: Maybe we should create an option in the `:platform:cortex-m` module with a nice description text so that this is more apparent to the user?

src/modm/driver/temperature/max31865.hpp
```diff
@@ -0,0 +1,249 @@
...
+	struct FaultThreshold
+	{
+		using RtdResistance_t = uint16_t;
+	};
+
+	static constexpr uint16_t
+	rtdfaultthresh(const FaultThreshold::RtdResistance_t res)
+	{
+		return static_cast<uint16_t>(res << 1);
+	};
```
rleh: What is this used for?
Is the `FaultThreshold::RtdResistance_t` typedef really necessary or could one just pass an `uint16_t` to `rtdfaultthresh( )`?

src/modm/driver/temperature/max31865.hpp
```diff
@@ -0,0 +1,249 @@
...
+	using typename max31865<pt>::FaultDetectionWrite;
+	using typename max31865<pt>::FaultDetectionWrite_t;
+	using typename max31865<pt>::FaultDetectionRead;
+	using typename max31865<pt>::FaultDetectionRead_t;
+	using typename max31865<pt>::Rejection;
+	using typename max31865<pt>::Rejection_t;
+	using typename max31865<pt>::FaultThreshold;
+	using typename max31865<pt>::rtdfaultthresh;
+	using typename max31865<pt>::Register;
+	using typename max31865<pt>::Data;
```
rleh: Does anyone know a nicer way to do this?
(I don't understand the reason why members of a templated base class are hidden in the derived class.)

src/modm/platform/core/cortex/module.lb
salkinium: Nah, you should write `float(1.0/3.0)` or `1.0f/3.0f` to get a single precision floating point constant, which is consistent with the C promotion rules. This flag was just a mistake.

src/modm/driver/temperature/max31865.hpp
chris-durand: Isn't the original purpose of this base class pattern that the definitions are not dependent on the template arguments? It does not make much sense that your register definitions are different types if you change the sensor resistance.

chris-durand: Move the template arguments to the inner `Data` class and make `max31865` a regular class.

src/modm/driver/temperature/max31865.hpp
```diff
@@ -0,0 +1,249 @@
...
+	.alpha = 3.85055e-3f,
+	.a = double{3.90830e-3},
+	.b = double{-5.77500e-7},
+	.c = double{-4.18301e-12},
+};
+
+/// @ingroup modm_driver_max31865
+// Standard values for IEC 751 (PT100) with 400 Ohm reference
+template<Pt pt>
+struct max31865
```
chris-durand:
```suggestion
struct max31865
```

src/modm/driver/temperature/max31865.hpp
```diff
@@ -0,0 +1,249 @@
...
+		ReadFaultStatus = 0x07,
+
+		WriteConfiguration = 0x80,
+		WriteHighFaultThresholdMsb = 0x83,
+		WriteHighFaultThresholdLsb = 0x84,
+		WriteLowFaultThresholdMsb = 0x85,
+		WriteLowFaultThresholdLsb = 0x86,
+	};
+
+	struct modm_packed Data
```
chris-durand, line 125:
```suggestion
	template<Pt pt>
	struct modm_packed Data
```

src/modm/driver/temperature/max31865.hpp
rleh: Fair point. I pushed a commit that moves the `[...]::Data` struct completely outside the `max31865` struct.
Not sure if that really better/nicer that just moving the template argument to the `Data` sub-struct...

src/modm/driver/temperature/max31865.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+/// Standard values for IEC 751 (PT100) with 430 Ohm reference
+constexpr Pt pt100{
+	.R0 = 100.f,
+	.Rref = 430.f,
+	.alpha = 3.85055e-3f,
+	.a = double{3.90830e-3},
+	.b = double{-5.77500e-7},
+	.c = double{-4.18301e-12},
+};
```
rleh: I don't like that this pollutes the `modm::` namespace with `Pt` and `pt100`, which are not obvious without the context. They should be moved inside the non-template `max31865` struct.

src/modm/driver/temperature/max31865.hpp
chris-durand: I would have put `Data` into `max31865` for consistency. It's like that in most other drivers. There is no other technical reason.

src/modm/platform/core/cortex/module.lb
salkinium: Ok, I see the problem with this 🤦. I'll move it to it's own PR.

src/modm/driver/temperature/max31865_impl.hpp
```diff
@@ -0,0 +1,115 @@
...
+	RF_CALL(writeSingleRegister(Register::WriteConfiguration, config.value));
+	timeout.restart(10ms);
+	RF_WAIT_UNTIL(timeout.isExpired());
+
+	config.set(Config::OneShot);
+	RF_CALL(writeSingleRegister(Register::WriteConfiguration, config.value));
+	timeout.restart(65ms);
+	RF_WAIT_UNTIL(timeout.isExpired());
+
+	data.data = RF_CALL(readTwoRegisters(Register::ReadRtdMsb));
```
hshose: I'm testing with hardware and it breaks here: `data.data` remains always 0, even though the max31865 register holds the correct value and my spi `readtworegisters` function reads it correctly.

If I have a `uint16_t` member in the `Max31865` class, like so:
```c++
	Data &data;
	uint16_t d;
```
then the return of the `RF_CALL` is assigned correctly, i.e.,
```c++
	data.data = RF_CALL(readTwoRegisters(Register::ReadRtdMsb));
	MODM_LOG_INFO << data.data <<modm::endl;
	d = RF_CALL(readTwoRegisters(Register::ReadRtdMsb));
	MODM_LOG_INFO << d <<modm::endl;
```
where `data.data`  prints value 0 and `d` prints value 16376 (the correct value for my room temperature and rtd). Both times, with logic analyzer, I read the same value from the MAX31865 register.
What is wrong here?
Could this be something broken with the resumeables or am I missing something simple?

When I test with commit 435077357, `data.data` also gets the correct value, and the `readTwoRegisters` and `data.data` haven't changed since then, I think...

chris-durand: This is really weird ...
What happens if you do:
```
d = RF_CALL(readTwoRegisters(Register::ReadRtdMsb));
MODM_LOG_INFO << d << modm::endl;
data.data = d;
MODM_LOG_INFO << data.data << modm::endl;
```

hshose: Then both `d` and `data.data` hold the correct value.
But why did assigning directly to `data.data` break?

hshose: The `d` workaround is not nice but works...

salkinium:
If it's not too annoying, could you try out your changed on top of #1001 and check if these "weird" issues dissappear?

rleh (approved):
Nice, thanks!

hshose:
> If it's not too annoying, could you try out your changed on top of #1001 and check if these "weird" issues dissappear?

If I also add the `:processing:protothread:use_fiber` option it works and fixes #1012.
I assume you wanted to mention this under #1012?

chris-durand (approved):
Thanks!

salkinium (approved):

Issue #907, opened 2022-09-02
RP2040 modm::chrono::micro_clock::now() is imprecise when called from Core1


cocasema:
When I was working on nrf24 code, I tried moving the rf logic to Core1, and got lots timeouts and delays. This initially got me thinking about some hw slowness due to locks etc, but luckily it narrowed down to `micro_clock` being very imprecise when called from Core1. While running on Core0, the delays between nrf24 write and interrupt ack would be  ~350us, and on Core1 I was getting 0 and ~1ms.

RP2040 provides [microsecond time via 64 bit counter](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#page=537). 
And if I replace the code of `modm::chrono::micro_clock::now()` in `platform/clock/systick_timer.cpp` with reading the `timer_hw->timel*` registers, I get reliable time with a microsend resolution on both cores.

I can make a PR with the changes. Q: Should I `%if` the implementation based on the target and keep it in the same systick_timer.cpp
or should `modm::chrono::micro_clock::now()` be moved to some rp2040 related file for this case?

salkinium:
It's a race-condition between the two cores that are accessing the [same `interrupt` variable](https://github.com/modm-io/modm/blob/develop/src/modm/platform/clock/systick/systick_timer.cpp.in#L22).
You could fix it for CM0 by making it an `bool interrupt[2]` array and using CPUID to index the right bool. The interrupt would set both true then.

You can still add the 64-bit counter as a separate module in the clock folder. The systick module should be optional (ie. only the BSP should pull it in), intentionally so that there can be alternate implentations. Not sure how FreeRTOS would then work on a dual-core systems with a single Systick though, probably only single core?

cocasema:
Thanks!

PR #997, opened 2023-04-08
[ext] Update submodules

rleh:
Let's not do this right before the (next) release first this time...

- [x] Hardware test
  - [x] CMSIS DSP
  - [x] SAMx7x
  - [x] TinyUsb
    - [x] F429 MSC, CDC
    - [x] RP2040 CDC
    - [x] F303 DFU
  - [x] pico-sdk 
  - [x] LVGL
    - [x] F469 Disco

### DIFF
--- a/examples/nucleo_f429zi/usbfatfs/main.cpp
+++ b/examples/nucleo_f429zi/usbfatfs/main.cpp
@@ -1,0 +1,156 @@
+/*
+ * Copyright (c) 2020, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <fatfs/ff.h>
+#include <tusb.h>
+#include <modm/board.hpp>
+#include <modm/processing.hpp>
+#include <modm/math/utils.hpp>
+
+/* You can test this by updating this firmware with itself:
+
+If you want to use the sanity check:
+
+	scons bin && \
+	cp ../../../build/nucleo_f429zi/usb_fatfs/scons-release/usbfatfs.bin . && \
+	echo -n "F429" >> usbfatfs.bin && \
+	crc32 usbfatfs.bin | xxd -r -p - >> usbfatfs.bin && \
+	mv usbfatfs.bin /Volumes/MODM_USB
+
+Otherwise just copy the file as is and `#define WITH_SANITY_CHECK 0`
+
+	scons bin && \
+	cp ../../../build/nucleo_f429zi/usb_fatfs/scons-release/usbfatfs.bin /Volumes/MODM_USB
+
+*/
+
+#define WITH_SANITY_CHECK 1
+
+// ----------------------------------------------------------------------------
+modm_ramcode
+void ram_apply(uint8_t pages, const uint8_t *image, uint32_t length)
+{
+	modm::atomic::Lock l;
+	// ONLY RAMCODE FROM HERE ON!!!
+	for (auto page{0u}; page <= pages; page++)
+		Flash::erase(page);
+	for (uintptr_t ptr{0}; ptr < length; ptr += sizeof(Flash::MaxWordType))
+		Flash::program(Flash::OriginAddr + ptr, *(Flash::MaxWordType*)(image + ptr));
+	NVIC_SystemReset();
+}
+
+bool is_valid(FIL *fil, size_t size)
+{
+#if WITH_SANITY_CHECK
+	UINT read;
+	// Validate processor type string:
+	f_lseek(fil, size-8);
+	char type[5];
+	f_read(fil, type, 4, &read);
+	if (memcmp(type, "F429", 4))
+		return false;
+	// Validate CRC32 of the entire image
+	uint32_t file_crc;
+	f_read(fil, &file_crc, 4, &read);
+	file_crc = modm::swap(file_crc);
+	f_rewind(fil);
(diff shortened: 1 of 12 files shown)


salkinium:
Yes, I think we should switch to updating the submodules as soon as they are released, and just bite the bullet and maybe have a little more work with integrating multiple new releases over the quarter (looking at you, ETL).
But then the testing is much more spread out and not such a blocker for our own release.

Would be interesting to have a bot create the PR automatically, but then again it's probably just more work programming that than doing the work ourselves.

rleh:
Absolutely.
I also don't like this PR already because all submodules are updated at once. Rather update and test individually, at least with TinyUSB, LVGL, FreeRTOS, etc...

chris-durand:
Two issues that will come up I have already solved elsewhere. I'll push the commits to here.

chris-durand:
The SAM UART defines have changed again. I'll quickly fix it ...

ext/rp/timer.h
```diff
@@ -0,0 +1,34 @@
...
+/*! \brief Return a 32 bit timestamp value in microseconds
+*  \ingroup hardware_timer
+*
+* Returns the low 32 bits of the hardware timer.
+* \note This value wraps roughly every 1 hour 11 minutes and 35 seconds.
+*
+* \return the 32 bit timestamp
+*/
+static inline uint32_t time_us_32(void) {
+    return timer_hw->timerawl;
```
salkinium: Why not `modm::PreciseClock::now()`? Does TinyUSB configure the timer itself?

chris-durand: I did not feel comfortable to change the code because I don't understand enough about the RP2040 platform to know if the modm timer does the right thing or not. The function is used inside some USB interrupt handler. Would it be safe to call `modm::PreciseClock::now()` from there, especially with the two cores? 

That `timer_hw` is some alternative platform specific microsecond timer. It was also mentioned in #907. TinyUSB does not configure it, but probably the platform initialization code does somehow. I was not able to figure out yet if it works with modm out of the box or not.

salkinium: Oh, then I should really fix that issue!

chris-durand: That timer is actually only used if [TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX](https://github.com/hathach/tinyusb/blob/28817a715024c82b78f9b9d19f37a72771870759/src/portable/raspberrypi/rp2040/rp2040_usb.c#L381) is set. It enables a workaround for a silicon errata which is only relevant if the RP2040 is plugged into a RPi4. The issue does not occur with any other hub or host controller. It is quite expensive and reduces bulk transfer bandwidth by 20%. We could simply decide to not support the workaround for now if it's too much work.

salkinium: Yes, let's fix it later.
I want to use the hardware timer anyways to fix #907, just not right now.

chris-durand: The "timer.h" header is now empty and the workaround is disabled. It has to be there because tinyusb always includes it, even when it is not used.

chris-durand:
Could someone test LVGL, tinyusb and RP2040 in hardware? I don't have access to any suitable hardware right now.

rleh:
> Could someone test LVGL, tinyusb and RP2040 in hardware? I don't have access to any suitable hardware right now.

After April 16th, yess :)

salkinium:
I'll test it next week in the detail it deserves, otherwise it's a rushed job right now.

ext/rp/timer.h
```diff
@@ -0,0 +1 @@
+// Empty header "timer.h" for tinyusb
```
rleh, line 1:
```suggestion
// Empty header "timer.h" for tinyusb

#if defined TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX && (TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX == 1)
#error TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX is not supported in modm currently.
#endif
```
Maybe, just in case someone tries to enable `TUD_OPT_RP2040_USB_DEVICE_UFRAME_FIX`?

salkinium:
Tested all the remaining hardware, it's good to go from my side. I'll rebase and squash.

salkinium:
I also added an action to the `modm-ext/*-partial`repo's workflow that creates some activity if more than 50 days have passed. Before I had to re-enable the cronjobs every 60 days, and I would forget because GitHub sends no email and just silently disables the workflow. You then have to go to actions and click on the workflow to see if it is disabled, utter insanity.

salkinium:
(I squashed the fixes into the submodule update commits, so that they are an atomic update.)

salkinium (approved):

examples/nucleo_f429zi/usbfatfs/main.cpp
```diff
@@ -0,0 +1,156 @@
...
+// ----------------------------------------------------------------------------
+modm_ramcode
+void ram_apply(uint8_t pages, const uint8_t *image, uint32_t length)
+{
+	modm::atomic::Lock l;
+	// ONLY RAMCODE FROM HERE ON!!!
+	for (auto page{0u}; page <= pages; page++)
+		Flash::erase(page);
+	for (uintptr_t ptr{0}; ptr < length; ptr += sizeof(Flash::MaxWordType))
+		Flash::program(Flash::OriginAddr + ptr, *(Flash::MaxWordType*)(image + ptr));
```
chris-durand, line 46: Deferencing `*(Flash::MaxWordType*)(image + ptr)` is UB unless `Flash::MaxWordType` is one of `char`, `unsigned char`, `signed char` or `std::byte`. The standard requires you to use `memcpy` here which would get optimized out. At least until we get `std::start_lifetime_as` from C++23. If it works we can leave it as is for now and fix it later.

chris-durand (approved):
Thanks

PR #999, opened 2023-04-12
[sam] Fix SAMx7x DMA LinkedListTransfer copy construction and assignment

chris-durand:
The compiler generated default copy operations do not update the internal linked list structure to the new addresses. In the example the compiler could do copy elision so nothing was ever copied. Tested in hardware.

### DIFF
--- a/src/modm/platform/dma/samx7x/transfer.hpp
+++ b/src/modm/platform/dma/samx7x/transfer.hpp
@@ -126,6 +126,9 @@ public:
 
 	constexpr LinkedListTransfer();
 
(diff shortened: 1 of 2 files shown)


salkinium (approved):
Yes, I… guess? C++ is scary \*crawls back to Python\*.

chris-durand:
> Yes, I… guess? C++ is scary *crawls back to Python*.

It's actually not that complicated. There is data in the object which contains pointers to itself because it's a linked list. If you move the thing somewhere else you can't just `memcpy` it byte by byte because the pointers inside the data structure will still point to the old location where you copied it from.

PR #1004, opened 2023-04-20
[math] added initializer_list initialization to matrix

hshose:
Added constructor with initializer list to `modm::math::Matrix` class.

The huge diff is from the clang-format script... Not sure if you want to have this?

### DIFF
--- a/src/modm/math/matrix.hpp
+++ b/src/modm/math/matrix.hpp
@@ -19,6 +19,7 @@
 #include <stdint.h>
 
 #include <modm/io/iostream.hpp>
+#include <modm/math/matrix.hpp>
 
 namespace modm
 {
@@ -60,7 +61,7 @@ namespace modm
 		 * Creates a Matrix with uninitialized elements. Use zeroMatrix() to
 		 * create a matrix with all elements set to zero.
 		 */
-		Matrix();
+		Matrix() = default;
 
 		/**
 		 * \brief	Create a matrix from an array
@@ -77,6 +78,17 @@ namespace modm
 		 * \endcode
 		 */
 		Matrix(const T *data);
+
+		/**
+		 * \brief	Create a matrix from an initializer list
+		 *
+		 * Example:
+		 * \code
+		 * modm::Matrix<int16_t, 3, 2> a{1,2,3,4,5,6};
+		 * \endcode
+		 */
+		template<typename... U>
+		explicit constexpr Matrix(U... data) requires (std::is_convertible_v<U, T> && ...);
 
 		/**
 		 * \brief	Get a zero matrix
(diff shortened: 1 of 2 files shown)


salkinium:
> The huge diff is from the clang-format script... Not sure if you want to have this?

No, I think we came to the conclusion that we only change the formatting if we refactor the entire file completely.
Otherwise the changes become simply unreviewable.

hshose:
Sure :face_with_spiral_eyes: undid the `clang-format`ing...

hshose:
Maybe also not so super relevant to merge this, if the modm custom matrix math is abandoned anyways..

salkinium:
> Maybe also not so super relevant to merge this, if the modm custom matrix math is abandoned anyways..

Kinda depends how long you can wait, cos I'm not very certain if eigen can really be intergrated until I try it.

src/modm/math/matrix_impl.hpp
```diff
@@ -30,6 +30,13 @@ modm::Matrix<T, ROWS, COLUMNS>::Matrix(const T *data)
 	}
 }
 
+// ----------------------------------------------------------------------------
+template<typename T, uint8_t ROWS, uint8_t COLUMNS>
+modm::Matrix<T, ROWS, COLUMNS>::Matrix(const std::initializer_list<T> data)
+{
+	std::copy(data.begin(), data.end(), element);
+}
```
chris-durand: The way `initializer_list` is designed there is no way to check the count of elements at compile time because the function argument is not a constant expression. That code is quite dangerous because it could initialize more elements and corrupt the stack or initialize less and leave values undefined.

A better alternative is using variadic templates. That way you can enforce the correct number of values.
```suggestion
template<typename T, uint8_t ROWS, uint8_t COLUMNS>
template<typename... U>
constexpr modm::Matrix<T, ROWS, COLUMNS>::Matrix(U... data)
	requires (std::is_convertible_v<U, T> && ...)
	: element{T(data)...}
{
	static_assert(sizeof...(data) == ROWS * COLUMNS, "Invalid element count");
}

```

src/modm/math/matrix.hpp
```diff
@@ -78,6 +78,16 @@ namespace modm
...
 
+		/**
+		 * \brief	Create a matrix from an initializer list
+		 *
+		 * Example:
+		 * \code
+		 * modm::Matrix<int16_t, 3, 2> a{1,2,3,4,5,6};
+		 * \endcode
+		 */
+		Matrix(const std::initializer_list<T> data);
```
chris-durand:
```suggestion
		template<typename... U>
		constexpr Matrix(U... data) requires (std::is_convertible_v<U, T> && ...);
```

chris-durand:
@hshose I have committed the review suggestion to run the CI. Please test if the code works for you.

salkinium (approved):
Nice!

src/modm/math/matrix_impl.hpp
hshose: Wizardry :magic_wand:
Very nice :+1:  thanks!

hshose:
Yap, works, thanks.

PR #1010, opened 2023-05-01
[clock] Fix multicore us clock reading on RP2040

### LINKED ISSUE #907: RP2040 modm::chrono::micro_clock::now() is imprecise when called from Core1

salkinium:
When using the microsecond clock from both Core0 and Core1, the interrupt semaphore was accessed by both, which broke the overflow detection and caused jitters of up to 1ms.
This gives each core its own flag.

- [x] Implement fix
- [x] Test fix
- ~Add 64-bit timer as alternative~ too much effort, cannot be bothered
- [x] Add examples

Fixes #907. cc @cocasema

### DIFF
--- a/examples/rp_pico/mcblink/main.cpp
+++ b/examples/rp_pico/mcblink/main.cpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2023, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/debug/logger.hpp>
+#include <modm/processing.hpp>
+#include <mutex>
+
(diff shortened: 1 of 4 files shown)


salkinium:
Can you check if this also fixes your Problem, @cocasema?

cocasema:
> Can you check if this also fixes your Problem, @cocasema?

Ran the example till `loop: 13647`, no ERRORs reported. :heavy_check_mark:

PR #1014, opened 2023-05-04
[ci] Fix MacOS workflow

rleh:
- [x] Fix bug with GCC by using `gcc@12` ~~(Apple thinks it's a good idea to link `g++` to their Clang/LLVM binary. Clang 14 seems to have some issues.)~~
- [x] Update to MacOS 13 (Ventura)

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -7,10 +7,6 @@ jobs:
     runs-on: macos-13
 
     steps:
-      # update Xcode version to fix linker bug
-      # - uses: maxim-lobanov/setup-xcode@v1
-      #   with:
-      #     xcode-version: '14.2'
 
       - name: Setup environment - Brew tap
         run: |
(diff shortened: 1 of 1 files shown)


salkinium:
So scons will do the right thing and [choose `gcc-13` (installed through homebrew) instead of `gcc` (the llvm abomination)](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/resources/SConscript.in#L27), you must not symlink gcc and g++, as that will break a lot of stuff.
I'm trying to reproduce this on my setup, but I am failing so far.

salkinium:
I dunno, it explicitly complained about boost linking issues, is this really a compiler issue, or a boost issue?

```
ld: warning: direct access in function '__ZN5boost4asio6detail20posix_tss_ptr_createERm'
from file 'build/scons-release/main.o' to global weak symbol 
'__ZZN5boost16exception_detail27get_static_exception_objectINS0_10bad_alloc_EEENS_13exception_ptrEvE2ep'
from file 'build/scons-release/main.o' means the weak symbol cannot be overridden at runtime.
This was likely caused by different translation units being compiled with different visibility settings.
```

salkinium:
Nothing stands out in the macOS runner image update from last week:
https://github.com/actions/runner-images/commit/816adb335a8f322b844463a50fbbeead1b82501c

rleh:
Looks like GCC13 was the problem, the CI [here now runs fine](https://github.com/modm-io/modm/actions/runs/4887961992/jobs/8725148340?pr=1014) with gcc@12.

salkinium:
You're the macOS whisperer!1!!

salkinium (approved):
Thanks!

PR #1015, opened 2023-05-04
[build] Remove GCC10 vs GCC12 compile flags differences

salkinium:
There was a time we were supporting GCC10, GCC11, and GCC12, unfortunately they required slightly different compile and link flags, which had to be detected at runtime.

Since we're now GCC12 only, we can remove these hacks.


rleh (approved):

rleh:
Merge this PR next (once CI checks pass)?

chris-durand (approved):

PR #1023, opened 2023-05-18
Use isNotEmpty() instead of !isEmpty()

strongly-typed:
Using only one part of #857

### DIFF
--- a/src/modm/communication/rpr/interface_impl.hpp
+++ b/src/modm/communication/rpr/interface_impl.hpp
@@ -467,7 +467,7 @@ template <typename Device, std::size_t N>
 void
 modm::rpr::Interface<Device, N>::popMessage(Queue &queue)
 {
-	if (!queue.isEmpty())
+	if (queue.isNotEmpty())
 	{
 		// deallocate the external buffer
 		Message message = queue.get();
--- a/src/modm/communication/xpcc/backend/can/connector_impl.hpp
+++ b/src/modm/communication/xpcc/backend/can/connector_impl.hpp
@@ -123,7 +123,7 @@ xpcc::CanConnector<Driver>::sendWaitingMessages()
 	}
 	else if (canDriver->getBusState() != Driver::BusState::Connected) {
 		// No connection to the CAN bus, drop all messages which should be send
-		while (!sendList.isEmpty()) {
+		while (sendList.isNotEmpty()) {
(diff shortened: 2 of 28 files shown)


strongly-typed:
The CI hangs. Can someone re-trigger the missing jobs?

salkinium:
CI doesn't hang, it purposely doesn't execute the HAL tests until you add the `ci:hal` label. They take forever and block other PRs from running jobs (we only get 20 for the whole repo).
I think you have the access rights to add the label yourself (try it).

salkinium (approved):
Sure, I guess?

We're going to deprecate the containers in favor of stdlibc++ (or ETL) in #1020, so don't spend too much time polishing them ;-P

chris-durand:
> We're going to deprecate the containers in favor of stdlibc++ (or ETL) in #1020, so don't spend too much time polishing them ;-P

Yes, the reason for the existence of most of the modm containers is that we couldn't rely on any standard library facilities at all before we had added libstdc++ for AVR. The containers are heavily dated and don't follow established practices regarding naming and semantics. That makes it hard to replace them with something else without changing all your code and interoperability with the standard library is impossible. 

The plan is to get rid of everything the standard has a conceptually equivalent replacement for and deprecate the redundant containers in modm (#1020). I'd like to replace the `BoundedDeque` with a more modern implementation with a standard library compatible interface and also add a fixed capacity, variable size `std::vector` like type based on [that proposal](https://github.com/gnzlbg/static_vector). Unfortunately the versions from ETL are not the most modern implementations either because their code base still focuses on C++03 compatibility. For instance, there is no `constexpr` support.

chris-durand (commented):
We can add the `isNotEmpty()` methods but I am sceptical of the global refactoring of `!isEmpty()` to `isNotEmpty()`.

Sure it wouldn't do any harm now, but the plan is to get rid of the non-standard modm container interface in the near future. More or less everyone agrees on how container operations should be named and which the basic operations are, (e.g. standary library, ETL, boost, EASTL).
If we stick to conventions the code will be easier to work with for many people and it will be possible to change containers to some other implementation without big code modifications.

For me personally I don't see a readability improvement in `isNotEmpty()` over `!isEmpty()` because I am used to `!empty()` which is how it would by idiomatically expressed with any other container implementation.

salkinium:
Closing this.

PR #1026, opened 2023-05-20
[fiber] Add more concurrency classes

salkinium:
Fibers need a standard set of efficient constructs.
I modelled these classes after the C++ thread concurrency interfaces. Not sure if useful, but also probably not really much room for improvement anyways.

- [x] https://github.com/modm-io/avr-libstdcpp/pull/35
- [x] `modm::fiber::sleep` -> `modm::this_fiber::sleep_for`
- [x] `modm::this_fiber::sleep_until`
- [x] `modm::this_fiber::get_id`
- [x] Align `modm::fiber::Task` to `std::jthread` interface, incl. stop_token.
- [x] stop_token and stop_source (simplified)
- [x] mutex
- [x] timed_mutex
- [x] recursive_mutex
- [x] recursive_timed_mutex
- [x] counting_semaphore (=binary_semaphore)
- [x] shared_mutex
- [x] shared_timed_mutex
- [x] once_flag
- [x] call_once
- [x] latch
- [x] barrier
- [x] condition_variable
- [x] unit tests
- [x] More documentation
- [x] Hardware testing
  - [x] STM32F723
  - [x] STM32G071
  - [x] ATmega2650

### DIFF
--- a/src/modm/processing/fiber/barrier.hpp
+++ b/src/modm/processing/fiber/barrier.hpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2024, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include "fiber.hpp"
+#include <limits>
+
+namespace modm::fiber
+{
+
+/// @ingroup modm_processing_fiber
+/// @{
+
+/// Implements the `std::barrier` interface for fibers.
+/// @warning This implementation is not interrupt-safe!
+/// @see https://en.cppreference.com/w/cpp/thread/barrier
+template< class CompletionFunction = decltype([]{}) >
+class barrier
+{
+	barrier(const barrier&) = delete;
+	barrier& operator=(const barrier&) = delete;
+	using count_t = uint16_t;
+
+	const CompletionFunction completion;
+	count_t expected;
+	count_t count;
+	count_t sequence{};
+
+public:
+	using arrival_token = count_t;
+
+	constexpr explicit
+	barrier(std::ptrdiff_t expected, CompletionFunction f = CompletionFunction())
+	: completion(std::move(f)), expected(expected), count(expected) {}
+
+	[[nodiscard]]
+	static constexpr std::ptrdiff_t
+	max() { return std::numeric_limits<count_t>::max(); }
+
+	[[nodiscard]]
+	arrival_token
+	arrive(count_t n=1)
+	{
+		count_t last_arrival{sequence};
+		if (n < count) count -= n;
+		else
+		{
+			count = expected;
+			sequence++;
+			completion();
+		}
+		return last_arrival;
+	}
+
+	void
+	wait(arrival_token arrival) const
+	{
+		while (arrival == sequence) modm::this_fiber::yield();
+	}
+
+	void
+	arrive_and_wait()
+	{
+		wait(arrive());
+	}
+
+	void
+	arrive_and_drop()
+	{
+		if (expected) expected--;
+		(void) arrive();
+	}
+};
+
+/// @}
+
+}
--- a/src/modm/processing/fiber/condition_variable.hpp
+++ b/src/modm/processing/fiber/condition_variable.hpp
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) 2024, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#pragma once
+
+#include "fiber.hpp"
+#include "stop_token.hpp"
+#include <atomic>
+
+
+namespace modm::fiber
+{
+
+/// @ingroup modm_processing_fiber
+/// @{
+
+enum class
+cv_status
+{
+	no_timeout,
+	timeout
+};
+
+/// Implements the `std::condition_variable_any` interface for fibers.
+/// @see https://en.cppreference.com/w/cpp/thread/condition_variable
+class condition_variable_any
+{
+	condition_variable_any(const condition_variable_any&) = delete;
+	condition_variable_any& operator=(const condition_variable_any&) = delete;
+
+	std::atomic<uint16_t> sequence{};
+
+	const auto inline wait_on_sequence()
+	{
+		return [this, poll_sequence = sequence.load(std::memory_order_acquire)]
+		{ return poll_sequence != sequence.load(std::memory_order_acquire); };
+	}
+public:
+	constexpr condition_variable_any() = default;
+
+	/// @note This function can be called from an interrupt.
+	void inline
+	notify_one()
+	{
+		sequence.fetch_add(1, std::memory_order_release);
+	}
+
+	/// @note This function can be called from an interrupt.
+	void inline
+	notify_any()
+	{
+		notify_one();
+	}
+
+
+	template< class Lock >
+	void
+	wait(Lock& lock)
+	{
+		lock.unlock();
+		this_fiber::poll(wait_on_sequence());
+		lock.lock();
+	}
+
+	template< class Lock, class Predicate >
+	requires requires { std::is_invocable_r_v<bool, Predicate, void>; }
+	void
+	wait(Lock& lock, Predicate&& pred)
+	{
+		while (not std::forward<Predicate>(pred)()) wait(lock);
+	}
+
+	template< class Lock, class Predicate >
+	requires requires { std::is_invocable_r_v<bool, Predicate, void>; }
+	bool
+	wait(Lock& lock, stop_token stoken, Predicate&& pred)
+	{
+		while (not stoken.stop_requested())
+		{
+			if (std::forward<Predicate>(pred)()) return true;
+			wait(lock);
+		}
+		return std::forward<Predicate>(pred)();
+	}
+
+
+	template< class Lock, class Rep, class Period >
+	cv_status
+	wait_for(Lock& lock, std::chrono::duration<Rep, Period> rel_time)
+	{
+		lock.unlock();
+		const bool result = this_fiber::poll_for(rel_time, wait_on_sequence());
+		lock.lock();
+		return result ? cv_status::no_timeout : cv_status::timeout;
+	}
+
+	template< class Lock, class Rep, class Period, class Predicate >
+	requires requires { std::is_invocable_r_v<bool, Predicate, void>; }
+	bool
+	wait_for(Lock& lock, std::chrono::duration<Rep, Period> rel_time, Predicate&& pred)
+	{
+		while (not std::forward<Predicate>(pred)())
+		{
+			if (wait_for(lock, rel_time) == cv_status::timeout)
+				return std::forward<Predicate>(pred)();
+		}
+		return true;
+	}
+
+	template< class Lock, class Rep, class Period, class Predicate >
+	requires requires { std::is_invocable_r_v<bool, Predicate, void>; }
+	bool
+	wait_for(Lock& lock, stop_token stoken,
+			 std::chrono::duration<Rep, Period> rel_time, Predicate&& pred)
+	{
+		while (not stoken.stop_requested())
+		{
+			if (std::forward<Predicate>(pred)()) return true;
+			if (wait_for(lock, rel_time) == cv_status::timeout)
+				return std::forward<Predicate>(pred)();
+		}
+		return std::forward<Predicate>(pred)();
+	}
+
+
+	template< class Lock, class Clock, class Duration >
+	cv_status
+	wait_until(Lock& lock, std::chrono::time_point<Clock, Duration> abs_time)
+	{
+		lock.unlock();
+		const bool result = this_fiber::poll_until(abs_time, wait_on_sequence());
+		lock.lock();
+		return result ? cv_status::no_timeout : cv_status::timeout;
+	}
+
+	template< class Lock, class Clock, class Duration, class Predicate >
+	requires requires { std::is_invocable_r_v<bool, Predicate, void>; }
+	bool
+	wait_until(Lock& lock, std::chrono::time_point<Clock, Duration> abs_time, Predicate&& pred)
+	{
+		while (not std::forward<Predicate>(pred)())
+		{
+			if (wait_until(lock, abs_time) == cv_status::timeout)
+				return std::forward<Predicate>(pred)();
+		}
+		return true;
+	}
+
+	template< class Lock, class Clock, class Duration, class Predicate >
+	requires requires { std::is_invocable_r_v<bool, Predicate, void>; }
+	bool
+	wait_until(Lock& lock, stop_token stoken,
+			   std::chrono::time_point<Clock, Duration> abs_time, Predicate&& pred)
+	{
+		while (not stoken.stop_requested())
+		{
+			if (std::forward<Predicate>(pred)()) return true;
+			if (wait_until(lock, abs_time) == cv_status::timeout)
+				return std::forward<Predicate>(pred)();
+		}
+		return std::forward<Predicate>(pred)();
+	}
+};
+
+/// There is no specialization for `std::unique_lock<fiber::mutex>`.
+using condition_variable = condition_variable_any;
+
+/// @}
+
+}
--- a/src/modm/processing/fiber/functions.hpp
+++ b/src/modm/processing/fiber/functions.hpp
@@ -42,6 +42,13 @@ yield()
 	modm::fiber::Scheduler::instance().yield();
 }
 
+/// Returns the id of the current fiber
+inline modm::fiber::id
+get_id()
+{
+	return modm::fiber::Scheduler::instance().get_id();
+}
+
 /**
  * Yields the current fiber until the time interval has elapsed.
  * This functionality is a convenience wrapper around `modm::Timeout`
--- a/src/modm/processing/fiber/functions.hpp
+++ b/src/modm/processing/fiber/functions.hpp
@@ -16,7 +16,7 @@
 #include "scheduler.hpp"
 #include <modm/processing/timer.hpp>
 
-namespace modm::fiber
+namespace modm::this_fiber
 {
 
 /// @ingroup modm_processing_fiber
@@ -39,7 +39,7 @@ namespace modm::fiber
 inline void
 yield()
 {
-	Scheduler::instance().yield();
+	modm::fiber::Scheduler::instance().yield();
 }
 
 /**
@@ -53,7 +53,7 @@ yield()
  */
 template< typename Rep, typename Period >
 void
-sleep(std::chrono::duration<Rep, Period> interval)
+sleep_for(std::chrono::duration<Rep, Period> interval)
 {
 	// Only choose the microsecond clock if necessary
 	using TimeoutType = std::conditional_t<
@@ -70,4 +70,21 @@ sleep(std::chrono::duration<Rep, Period> interval)
 
 /// @}
 
-} // namespace modm::fiber
+} // namespace modm::this_fiber
+
+/// @cond
+// DEPRECATE: 2025q2
+namespace modm::fiber
+{
+
+[[deprecated("Use `modm::this_fiber::yield()` instead!")]]
+void inline yield()
+{ this_fiber::yield(); }
+
+template< class Rep, class Period >
+[[deprecated("Use `modm::this_fiber::sleep_for()` instead!")]]
+void sleep(std::chrono::duration<Rep, Period> sleep_duration)
+{ this_fiber::sleep_for(sleep_duration); }
+
+}
+/// @endcond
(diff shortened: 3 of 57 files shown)


salkinium:
It's still missing unit tests for latch, barrier and condition_variable and a bunch of documentation, however, it would be amazing if this could get reviewed with a special focus on C++ "conformity".
This implementation should be interrupt-safe (when it makes sense for the functions), so that we can use then in the HAL as signalling primitives for operation completions in the future.

I also noticed that avrlibstd++ does not have `<atomic>`, I assume it was really annoying to port, @chris-durand? If it's an issue with the builtin gcc atomics not being implemented for avr-gcc, we can reuse the Cortex-M0 ones now.

chris-durand:
> I also noticed that avrlibstd++ does not have `<atomic>`, I assume it was really annoying to port, @chris-durand? If it's an issue with the builtin gcc atomics not being implemented for avr-gcc, we can reuse the Cortex-M0 ones now.

I haven't tried to be honest but I expect that the same solution as for Cortex-M0 will work. The header is mostly implemented using compiler builtins. I'd expect it to fall back to calling a library function in case the operation is not atomic on AVR. If you are lucky just dropping in the headers and compiling `atomics_c11_cortex.cpp.in` for AVR could be sufficient.

C++20 atomic wait which requires OS support shouldn't be an issue because it is only enabled when `_GLIBCXX_HAS_GTHREADS` is set.

chris-durand:
I'll take a look at the PR over the weekend.

salkinium:
> If you are lucky just dropping in the headers and compiling `atomics_c11_cortex.cpp.in` for AVR could be sufficient.

Yup that just works, thanks!

src/modm/processing/fiber/functions.hpp
```diff
@@ -70,4 +70,21 @@ sleep(std::chrono::duration<Rep, Period> interval)
...
+// DEPRECATE: 2025q2
+namespace modm::fiber
+{
+
+modm_deprecated("Use `modm::this_fiber::yield()` instead!")
+void inline yield()
+{ this_fiber::yield(); }
+
+template< class Rep, class Period >
+modm_deprecated("Use `modm::this_fiber::sleep_for()` instead!")
```
chris-durand: Is there any advantage of using that macro over a standard C++ attribute (`[[deprecated("...")]]`)?

src/modm/processing/fiber/functions.hpp
```diff
@@ -50,29 +51,85 @@ get_id()
...
+ *       sleep duration may be longer without any guarantee of an upper limit.
+ */
+template< class Clock, class Duration, class Function  >
+[[nodiscard]] bool
+poll_until(std::chrono::time_point<Clock, Duration> sleep_time, Function &&condition)
+{
+	const auto start = Clock::now();
+	const auto sleep_duration = sleep_time - start;
+	do {
+		if (condition()) return true;
```
chris-durand:
```suggestion
		if (std::forward<Function>(condition)()) return true;
```

to be 100% generic. For instance, if `condition` binds to an r-value and its type has an `operator()` with a `&&` ref-qualifier it won't get called without the `std::forward`. Probably never an issue in practice but still more correct :wink:.

src/modm/processing/fiber/functions.hpp
```diff
@@ -50,29 +51,85 @@ get_id()
...
+ * @note For nanosecond delays, use `modm::delay(ns)`.
+ * @note Due to the overhead of `yield()` and the scheduling other fibers, the
+ *       sleep duration may be longer without any guarantee of an upper limit.
+ * @see https://en.cppreference.com/w/cpp/thread/sleep_for
+ */
+template< class Rep, class Period >
+void
+sleep_for(std::chrono::duration<Rep, Period> sleep_duration)
+{
+	(void) poll_for(sleep_duration, [](){ return false; });
```
chris-durand: Just fyi, you can write `[]{ }` without the `()` now for a lambda that doesn't take any arguments.

src/modm/processing/fiber/barrier.hpp
```diff
@@ -0,0 +1,84 @@
...
+class barrier
+{
+	barrier(const barrier&) = delete;
+	barrier& operator=(const barrier&) = delete;
+	using count_t = uint16_t;
+
+	const CompletionFunction completion;
+	volatile count_t expected;
+	volatile count_t count;
+	volatile count_t sequence{};
```
chris-durand: This doesn't look interrupt-safe on 8-bit AVR with `++` and `-=` on volatile 16 bit variables. Shouldn't we use `std::atomic`?

src/modm/processing/fiber/mutex.hpp
```diff
@@ -0,0 +1,178 @@
...
+	std::atomic_bool locked{false};
+public:
+	constexpr mutex() = default;
+
+	[[nodiscard]]
+	bool inline
+	try_lock()
+	{
+		bool expected{false};
+		return locked.compare_exchange_strong(expected, true);
```
chris-durand:
```suggestion
		return locked.compare_exchange_strong(expected, true, std::memory_order_acquire, std::memory_order_relaxed);
```

A mutex only needs acquire-release ordering, acquire on lock and release on unlock. Acquire means no access can be reordered before locking the mutex and release says no access can be moved behind the unlock.
In case the compare-and-swap operation fails, no other operations are synchronized, so relaxed is fine.

src/modm/processing/fiber/mutex.hpp
```diff
@@ -0,0 +1,178 @@
...
+	void inline
+	lock()
+	{
+		while(not try_lock()) modm::this_fiber::yield();
+	}
+
+	void inline
+	unlock()
+	{
+		locked = false;
```
chris-durand:
```suggestion
		locked.store(false, std::memory_order_release);
```

src/modm/processing/fiber/shared_mutex.hpp
```diff
@@ -0,0 +1,125 @@
...
+public:
+	constexpr shared_mutex() = default;
+
+	[[nodiscard]]
+	bool inline
+	try_lock()
+	{
+		const fiber::id new_owner = modm::this_fiber::get_id();
+		fiber::id expected{NoOwner};
+		return owner.compare_exchange_strong(expected, new_owner);
```
chris-durand:
```suggestion
		return owner.compare_exchange_strong(expected, new_owner, std::memory_order_acquire, std::memory_order_relaxed);
```
Same here

src/modm/processing/fiber/shared_mutex.hpp
```diff
@@ -0,0 +1,125 @@
...
+	unlock()
+	{
+		owner = NoOwner;
+	}
+
+	[[nodiscard]]
+	bool inline
+	try_lock_shared()
+	{
+		fiber::id current = owner.load();
```
chris-durand:
```suggestion
		fiber::id current = owner.load(std::memory_order_relaxed);
```

src/modm/processing/fiber/shared_mutex.hpp
```diff
@@ -0,0 +1,125 @@
...
+		do if (current < SharedOwner) return false;
+		while(owner.compare_exchange_weak(current, SharedOwner) == false);
```
chris-durand:
```suggestion
		while(owner.compare_exchange_weak(current, SharedOwner, std::memory_order_acquire, std::memory_order_relaxed) == false);
```

src/modm/processing/fiber/shared_mutex.hpp
```diff
@@ -0,0 +1,125 @@
...
+	void inline
+	lock_shared()
+	{
+		while(not try_lock_shared()) modm::this_fiber::yield();
+	}
+
+	void inline
+	unlock_shared()
+	{
+		owner = NoOwner;
```
chris-durand:
```suggestion
		owner.store(NoOwner, std::memory_order_release);
```

src/modm/processing/fiber/shared_mutex.hpp
```diff
@@ -0,0 +1,125 @@
...
+	void inline
+	lock()
+	{
+		while(not try_lock()) modm::this_fiber::yield();
+	}
+
+	void inline
+	unlock()
+	{
+		owner = NoOwner;
```
chris-durand:
```suggestion
		owner.store(NoOwner, std::memory_order_release);
```

chris-durand (commented):
Very nice!

I haven't managed to look at all the synchronization primitives, but will do tomorrow.

src/modm/processing/fiber/barrier.hpp
salkinium: No, because barrier is only useful if called from different fibers, not also from interrupts. That's also the only reason I'm using atomics in the first place, to be able to call some of these functions from interrupts directly.

src/modm/processing/fiber/mutex.hpp
salkinium: Ok, still not entirely sure my choices were correct, but I've now added a memory order to all operations. I'm mostly unsure about my choices in condition variable.

chris-durand: I'll try to double check it, most likely on the weekend. I'll be quite busy this week.

src/modm/processing/fiber/barrier.hpp
```diff
@@ -0,0 +1,85 @@
...
+class barrier
+{
+	barrier(const barrier&) = delete;
+	barrier& operator=(const barrier&) = delete;
+	using count_t = uint16_t;
+
+	const CompletionFunction completion;
+	count_t expected;
+	count_t count;
+	volatile count_t sequence{};
```
chris-durand: Why is this volatile if the implementation isn't meant to be used with interrupts?
Is the loop in `wait()` optimized out otherwise? I wouldn't expect the compiler to do that.

src/modm/processing/fiber/barrier.hpp
```diff
@@ -0,0 +1,85 @@
...
+public:
+	using arrival_token = count_t;
+
+	constexpr explicit
+	barrier(std::ptrdiff_t expected, CompletionFunction f = CompletionFunction())
+	: completion(std::move(f)), expected(expected), count(expected) {}
+
+	[[nodiscard]]
+	static constexpr std::ptrdiff_t
+	max() { return count_t(-1); }
```
chris-durand:
```suggestion
	max() { return std::numeric_limits<count_t>::max(); }
```
A bit cleaner than the cast to get the maximum.

src/modm/processing/fiber/condition_variable.hpp
```diff
@@ -0,0 +1,178 @@
...
+		{ return poll_sequence != sequence.load(std::memory_order_relaxed); };
+	}
+public:
+	constexpr condition_variable_any() = default;
+
+	/// @note This function can be called from an interrupt.
+	void inline
+	notify_one()
+	{
+		sequence.fetch_add(1, std::memory_order_acquire);
```
chris-durand: Shouldn't this be a release operation? Some store happens before the notify and it shouldn't be reordered behind it. On the consumer side, which is the wait, we need an acquire load or barrier such that no memory access is reordered before the wait is complete.

An acquire operation always has to be paired with a matching release. Otherwise it doesn't provide any ordering guarantees.

Actually, is there any use case where this code would ever have to deal with more than one thread on a multi-core CPU or interrupts running on different cores? I assume not, because the reason for fibers is to have a single-threaded execution model.
In that case we don't even need the actual memory barrier instructions to be emitted. You could use relaxed order on the atomic operation plus a compiler fence ([std::atomic_signal_fence](https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence)). The only thing that can reorder operations within a single thread is the compiler. Hardware barriers on atomics are only ever needed if you have multithreading and multiple cores with their own caches and view on memory.

I don't really know how expensive a few additional `dmb` instructions are on Cortex-M and if it would be worth caring about it. Anyway I'm not sure if any of fibers would work in the presence of multithreading.

salkinium: > Actually, is there any use case where this code would ever have to deal with more than one thread on a multi-core CPU or interrupts running on different cores? I assume not, because the reason for fibers is to have a single-threaded execution model.

Well… my secret plan is to allow running one fiber scheduler per (FreeRTOS) thread and then do some kind of message passing between threads like Qt does with their event loops. I also want to support the dual-core H745 in the future. 

According to the [Programming Guide to Memory Barrier Instructions](https://developer.arm.com/documentation/dai0321/latest/) on Cortex-M0 to Cortex-M4, all (internal) memory accesses are always *in-order*, thus DMBs are not required. However, this is only an implementation choice, the architecture explicitly allows for reordering of memory accesses!

And indeed, there seem to be several caveats for single-core devices:

1. There is no guarantee of ordering on external memories, e.g. SDRAM connected over FMC. It's up to the implementation to correctly implement DMB, they even go as far as saying that you should do a dummy read on something you wrote to enforce the write order. For example, on [STM32F4, the FMC has several FIFOs](https://www.st.com/resource/en/reference_manual/dm00031020-stm32f405-415-stm32f407-417-stm32f427-437-and-stm32f429-439-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf#page=1605):
- Write Data FIFO with 16 x33-bit depth
- Write Address FIFO with 16x30-bit depth
- Cacheable Read FIFO with 6 x32-bit depth (6 x14-bit address tag) for SDRAM controller.

  However, it's unclear to me if these FIFOs are in-order relative to each other. (I would expect it, but who knows?)

2. The document does not talk about Cortex-M7, which has this 64-bit internal AXI bus, that seems to allow for reordering of memory accesses. Specifically, it can do speculative reads at any time. So I think you still need DMBs on single-core Cortex-M7.

I would just leave the DMBs in the code for now. For Cortex-M0(+), I implemented atomics without DMBs anyways, so the optimization of leaving out the DMBs would only apply to CM3/CM4 without external memories.

> I don't really know how expensive a few additional dmb instructions are on Cortex-M and if it would be worth caring about it

I can do a comparison of the DWT->LSUCNT (cycles spent waiting for loads and stores to complete) with and without DMBs once I have a representative example later on to get an idea of the overhead of having "unnecessary" DMBs in the code.

chris-durand: > And indeed, there seem to be several caveats for single-core devices:
> 1. There is no guarantee of ordering on external memories
> 2. The document does not talk about Cortex-M7, which has this 64-bit internal AXI bus, that seems to allow for reordering of memory accesses

And none of those should matter when there is only one core. The problem with multi-core CPUs is that you can observe modifications to multiple memory locations in a different order across cores.

Let's say you have a single-producer single-consumer atomic queue, you write some data to the end and increment the last element pointer. What can happen is that the pointer modification is visible before the actual data change. So you would read invalid values on the consumer thread because they are not visible yet. Acquire-release ordering on the pointer solves that issue.

If you are on a single-core machine this is impossible by definition. Writing a variable and reading back the value must give the same result. There is nothing to be out of order with. This is just single-threaded programming, I can't see what the `dmb`s could accomplish unless you are doing DMA or other similar things where peripherals interact with the memory directly. But that's the peripheral driver's job and out of scope for atomics.

The only things we need from the atomic is the atomicity and the compiler optimization barrier stopping it from doing the same reorderings as the multi-core system could do.

As a side note that is why the `volatile`-based modm interrupt "atomic" queue is broken. The compiler is free to reorder non-volatile accesses around volatile ones and gcc is known to do that (https://blog.regehr.org/archives/28, item 5). We should fix it.

> For Cortex-M0(+), I implemented atomics without DMBs anyways

For plain atomic loads and stores the compiler will still emit `ldr`/`str` plus `dmb` (https://gcc.godbolt.org/z/Enb6179Pb) instead of turning off interrupts. The fallback implementation is only used for read-modify-write operations the M0 can't do atomically due to the lack of `ldrex` and `strex`.

> I can do a comparison of the DWT->LSUCNT

Nice, I didn't even know that existed. That would be a good benchmark. Most likely we can just accept having the `dmb` instructions.

> I also want to support the dual-core H745 in the future

On the H7 dual-core you have to be careful with the caches if you are sharing data between both cores. If a value in SRAM is cached in the M7 DCACHE and has not been written back, an access from the M4 will read the old value from RAM. There is no automatic cache coherency.
ST's own programming model is to write data to a shared SRAM, trigger a cross-core interrupt and memcpy the data over. Not sure what to expect if we share atomics across both CPUs, but it should somehow work if there is no caching involved.

salkinium: > I can't see what the `dmb`s could accomplish

Yes that makes sense to me. The CM7 AXI is not free to reorder memory access if it modifies the single-threaded view of memory (ie. swapping write and read of the same memory address). And I assume that ST didn't screw up the access order for external memory either.

> The only things we need from the atomic is the atomicity and the compiler optimization barrier stopping it from doing the same reorderings as the multi-core system could do.

I thought about defining a bunch of `modm::memory_order_acquire = std::memory_order_relaxed` aliases for single-core devices and then using those everywhere, however, I looked into how GCC actually implements the builtins, and it seems to me like no barrier is generated for `std::memory_order_relaxed`, not even a compiler barrier:

1. `std::atomic<T>::load()` calls [the `__atomic_load()` builtins](https://github.com/gcc-mirror/gcc/blob/master/gcc/sync-builtins.def#L294).
2. I guess this then somehow calls [`‎expand_builtin_atomic_load‎()`](https://github.com/gcc-mirror/gcc/blob/master/gcc/builtins.cc#L6844) which calls
3. [`expand_atomic_load()`](https://github.com/gcc-mirror/gcc/blob/master/gcc/optabs.cc#L7385) and this then generates the actual instructions.
4. this generates a memory barrier via [`expand_memory_blockage()`](https://github.com/gcc-mirror/gcc/blob/master/gcc/optabs.cc#L7296) either as `dmb` or `asm volatile("" ::: "memory")`.
5. it only emits a (compiler) memory barrier if **NOT** relaxed. 

So I think we should still just use it the way it is intended even if it costs a few cycles more.

> Not sure what to expect if we share atomics across both CPUs, but it should somehow work if there is no caching involved.

We should add a MPU driver and mark that SRAM as non-cachable. We need to do this also for memories that are DMA-able on CM7, then we could enable the D-Cache by default.

chris-durand: > it seems to me like no barrier is generated for std::memory_order_relaxed, not even a compiler barrier

Yes, if you want the compiler barrier only you'll have to do a relaxed access on the atomic plus [`std::atomic_signal_fence`](https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence).

salkinium: I think the correct way to approach this is to modify the atomic builtin implementation and not fix it from the outside. Since I'm not motivated to shave that yak, I'll just keep using DMBs for now.

salkinium:
@chris-durand, I added ARMv8-M stack protection, could you check if the example/generic/fiber still works? I don't have a Cortex-M33 in my collection.

chris-durand:
> @chris-durand, I added ARMv8-M stack protection, could you check if the example/generic/fiber still works? I don't have a Cortex-M33 in my collection.

When running this on an STM32L5 I get a hardfault. The stack overflow bit (STKOF, bit 4) in UFSR is set. The fault happens inside `modm_context_jump`, triggered from the call to `yield()` inside the lambda of `fiber_y1`.

chris-durand:
Let me know if there is something specific I can check.

chris-durand:
Screenshot_20240514_005023

Listing: https://gist.github.com/chris-durand/4f3deae41aacbfd0cad2c9d067df0b5a

salkinium:
Hm, so setting MSPLIM is no problem, setting PSPLIM once on context_start is no problem either, but then on the first context switch it fails? ugh.

Could you try to change the assembly in context_jump?
1. 
```diff
 %% elif with_psplim
-		"ldm r1, {r1-r2}	\n\t"
-		"mov sp,  r1		\n\t"	// Set PSP to ctx->sp
-		"msr psplim, r2		\n\t"	// Set PSPLIM to ctx->bottom
+		"ldm r1, {r2-r3}	\n\t"
+		"mov sp,  r2		\n\t"	// Set PSP to ctx->sp
+		"msr psplim, r3		\n\t"	// Set PSPLIM to ctx->bottom
 %% else
```

2.
```diff
 %% elif with_psplim
-		"ldm r1, {r1-r2}	\n\t"
-		"mov sp,  r1		\n\t"	// Set PSP to ctx->sp
-		"msr psplim, r2		\n\t"	// Set PSPLIM to ctx->bottom
+		"ldr sp, [r1]		\n\t"	// Set PSP to ctx->sp
+		"ldr r2, [r1, #4]	\n\t"
+		"msr psplim, r2		\n\t"	// Set PSPLIM to ctx->bottom
 %% else
```

3. 
```diff
 %% elif with_psplim
-		"ldm r1, {r1-r2}	\n\t"
-		"mov sp,  r1		\n\t"	// Set PSP to ctx->sp
-		"msr psplim, r2		\n\t"	// Set PSPLIM to ctx->bottom
+		"ldm r1, {r2-r3}	\n\t"
+		"msr psplim, r3		\n\t"	// Set PSPLIM to ctx->bottom
+		"mov sp,  r2		\n\t"	// Set PSP to ctx->sp
 %% else
```

4. 
```diff
 %% elif with_psplim
-		"ldm r1, {r1-r2}	\n\t"
-		"mov sp,  r1		\n\t"	// Set PSP to ctx->sp
-		"msr psplim, r2		\n\t"	// Set PSPLIM to ctx->bottom
+		"ldr r2, [r1, #4]	\n\t"
+		"msr psplim, r2		\n\t"	// Set PSPLIM to ctx->bottom
+		"ldr sp, [r1]		\n\t"	// Set PSP to ctx->sp
 %% else
```

chris-durand:
No success with any of the options. I've also tried replacing the code in both context_start and context_jump.

salkinium:
Ok, then I'll descope it for now and will order a bunch of new development boards and debug it some more. Thanks for your help!

PR #1027, opened 2023-05-23
[stm32] Timer fix: avoid arithmetic overflow in setPeriod

ser-plu:
For timers 2 and 5 `Value` has `uint32` type, which results in an overflow and `prescaler=0` in `setPeriod` function.


chris-durand (approved):
Thanks!

PR #1032, opened 2023-05-30
[board] Add Adafruit Feather M4 Express BSP and examples

ghost:
Adafruit Feather M4 Express BSP and examples


examples/feather_m4/blink/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+	MODM_LOG_ERROR   << "error"   << modm::endl;
+
+	uint32_t counter(0);
+	
+    while (true)
+    {
+        Led::toggle();
+        modm::delay(500ms);
+	
+		MODM_LOG_INFO << "loop: " << counter++ << modm::endl;
```
chris-durand: The indentation here is a mix of tabs and spaces. Could you make sure all added files are consistent?

Our CI also rejects all files with lines containing trailing white space. It needs to be removed for the unit test CI job to pass.

src/modm/board/feather_m4/board.hpp
```diff
@@ -0,0 +1,171 @@
...
+	enable()
+	{
+		// start the 32kHz crystal oscillator and connect to generator
+		GenericClockController::enableExternalCrystal32k(Xosc32StartupTime::Start_500ms);
+		GenericClockController::enableGenerator<Generator32k, ClockSource::Xosc32k>();
+
+		// set up a 1MHz reference for DPLL0
+		GenericClockController::enableGenerator<Generator1MHz, 
+							GeneratorConfiguration{ClockSource::Dfll, 48U}>();
+		GenericClockController::connect<ClockPeripheral::Fdpll0>(Generator1MHz);
```
chris-durand: Couldn't we configure the DFLL48 in closed-loop mode with the 32k crystal as the reference?

chris-durand (commented):
Thanks!

The board needs to be added [here](https://github.com/modm-io/modm/blob/eda224ec0c7a3f72cc0be25fa89a346a3eccd31c/.github/workflows/linux.yml#L95) that the examples are compiled in the CI run.

Also, please run `python3 tools/scripts/synchronize_docs.py` in the modm top level folder and commit the updated `README.md` file.

src/modm/board/feather_m4/board.hpp
ghost: I had written an option into the board module.lb to choose between open and closed-loop on the DFLL48 (open-loop was the default), and then removed it. Should I put the option back, or just go with the closed-loop with 32k ref?

examples/feather_m4/blink/main.cpp
ghost: I have done the linux.yml, README.md, and whitespace cleanup. Waiting to hear back on closed-loop question before pushing files to my remote.
Once I push everything, do I have to do another PR? Not sure how this all ties together at this point.

ghost:
I ran "synchronize_docs.py", committed and pushed README.md. Then I spotted a typo in that file, and changed "you specific needs" to "your specific needs", committed and pushed it. Big mistake, I guess. Now it appears it's failing the test at that point. I tried to undo it from my local repo but have run into a "divergent branch" situation that I'm unsure how to fix. I don't believe there's really anything wrong with the file.

src/modm/board/feather_m4/board.hpp
chris-durand: If the board always has a 32k oscillator I'd go with the closed-loop mode. The USB spec has quite strict clock requirements (+- 0.25 %). The default open-loop mode won't meet the spec. Therefore, there is also the option to recover the clock from USB SOF packets and tune the DFLL48 to it, but our driver doesn't support it yet and it can only be enabled while connected to a USB host. The closed-loop mode with the 32k crystal should be the easiest option.

examples/feather_m4/blink/main.cpp
chris-durand: Thanks. There is no need to open another PR. We can continue here.

src/modm/board/feather_m4/module.md
```diff
@@ -0,0 +1,76 @@
...
+## Programming
+
+The Feather M4 has a UF2 bootloader. By using a utility to convert the '.elf'
+files (generated by scons or make) to '.uf2' files, it is then possible to 
+copy the uf2 file directly to the board.
+
+Adafruit has such a utility, called 'uf2conv.py', that works for this purpose
+(although the elf file needs to be converted to a binary or hex file first).
+The converter that is supplied with modm, 'elf2uf2.py', is ONLY for the rp2040,
+and will not work here.
```
salkinium: Why not? Is this a lack of configuration or are there different UF2 file formats?

ghost: I did not spend too much time analyzing it. I had been using uf2conv.py for my own (non-modm) projects. Not being aware of the BOSSAC option, I tried the same strategy of creating a uf2 with modm; but at first it failed because the default 'float-abi' option is 'hard', and elf2uf2 kicked it out:

    if (e32h[3] & EF_ARM_ABI_FLOAT_HARD) != 0 :
        raise Exception("HARD-FLOAT not supported")

So I changed float-abi to 'soft', and it failed again (I don't recall the message). I found this near the start of elf2uf2:

uf2_config = {
    "rp2040": {
        "MAGIC_START0": 0x0A324655,
        "MAGIC_START1": 0x9E5D5157,
        "MAGIC_END": 0x0AB16F30,
        "FAMILY_ID": 0xe48bff56,
        "PAGE_SIZE": (1 << 8),
    }
}

And that's the only entry.

To me, that would seem to indicate that this was specifically written for the rp2040. There is a command line argument for 'target' and it appears that other chips could be added, but there's nothing there right now.

At that point, I rewrote the makefile (was using make at the time) to use uf2conv.py, and my problem was solved. I later changed to using BOSSAC, but my goal was to submit a board package to modm, and elf2uf2 was not helping.

The comments in my documentation were only intended to help people trying to use this board with modm, and if the uf2 converter was changed to handle more targets, they could easily be updated.

src/modm/board/feather_m4/board.hpp
ghost: That's fine with me. Here are some diffs for board.hpp and module.md to get your feedback before I commit anything else. In particular, note (in board.hpp) the definition of 'Clock1MHz' and its explanation; the value comes out to 999424Hz. I don't suppose it's really necessary to do it that way. The other section involves the initialization of DFLL48 in closed-loop; the CPU clock is temporarily changed to XOSC32K for the changeover. (This git editor won't display the diffs the way I want, but hopefully you get the idea.)

board.hpp:
90c90,93
<     static constexpr uint32_t Clock1MHz = 1_MHz;
---
> 	// when DFLL is in closed loop mode, the "1_MHz" DPLL reference is not
> 	// exactly 1MHz; calculating the actual value allows 'enableDpll()' to
> 	// check the tolerance
> 	static constexpr uint32_t Clock1MHz = ((48_MHz / 32768) * 32768) / 48;
> 103a107,118
> 
> 		// set up Xosc32k as DFLL reference
> 		GenericClockController::connect<ClockPeripheral::Dfll48>(Generator32k);
> 
> 		// while restarting DFLL, run system temporarily at 32kHz
> 		GenericClockController::enableGenerator<ClockGenerator::Main, ClockSource::Xosc32k>(); > 
> 
> 		// restart DFLL in closed loop
> 		GenericClockController::enableDfll48mClosedLoop<Clock32k>();
> 
> 		// get system back up to speed
> 		GenericClockController::enableGenerator<ClockGenerator::Main, ClockSource::Dfll>();

module.md:
58,59c58,60
< (e.g. 1-2MHz). This is the method used by Adafruit and TinyUSB in their
< FeatherM4 libraries, and that is what is done here.

> (e.g. 1-2MHz). It can be done with the DFLL in open- or closed-loop mode, but
> in order to meet the USB timing specs, closed-loop with XOSC32K as reference is
> used here.

src/modm/board/feather_m4/module.md
salkinium: Yes, good find! I didn't realize how much our uf2 converter was hardcoded. Let's fix it sometime later… the workaround with BOSSAc works well.

salkinium (approved):
I'm happy with this, thanks!

chris-durand (approved):
Thanks!

ghost:
Thanks to your both for your patience! I now understand the process somewhat better, although I am still not friends with git just yet :-)
This is a great project, hope to do more in the future!

salkinium:
I've squashed and rebased your commits and will merge after the CI passes. Thanks!

PR #942, opened 2023-01-13
Add MS5837 high-resolution pressure and temperature sensor driver

lukh:
Implementation of the MS5837 high-resolution pressure and temperature sensor

heavily based on the MS5711 by Rasmus Kleist Hørlyck Sørensen

PR #1034, opened 2023-06-04
Fix MS5611 calibrated value calculation

rasmuskleist:
I realised that I have made a mistake when implementing the ms5611 driver. Thanks @lukh for pointing that out in your [PR](#942)!


chris-durand (approved):
Thanks! I have compared this to the datasheet and the calculation seems to be correct now.

PR #1039, opened 2023-06-22
[driver] Compare calculated crc with MS5611 prom

rasmuskleist:
It occured to me that in application code we only validate that the CRC is nonzero. I have updated the driver so it now compares the calculated CRC with the prom. I have tested this implementation with three barometers. Hopefully this will be the final update to this driver!


salkinium (approved):
Thanks!

chris-durand (approved):

PR #1040, opened 2023-06-23
[driver] Adding InvenSense 6-Axis IMU driver

rasmuskleist:
I have created this driver for IIM-42652, but it also works for ICM-42688-P and possibly many other InvenSense IMUs. The only difference between IIM-42652 and ICM-42688-P, that I have noticed from a software perspective is that IIM-42652 has Register Bank 3. Therefore I do not know if Iim42652 is the right name for the device?

In addition I have a couple of points that I would like your take on:
 1. I cannot quite figure out what the standard is for naming enum class members, when mapping out registers. I have written e.g. BankSel0 = Bit0 as in Adis16470. However, other implementations e.g lis3dsh the convention is BANK_SEL0 = Bit0. I will happily change the defintion to conform to an enforced standard.
 2. Currently I have only added the registers in user bank 0. I presume this is fine?
 3. I have only tested the driver with SPI and I am debating if I should only implement the SPI version and remove the transport class?
 4. I have not implemented FIFO nor APEX yet. However, I plan to implement FIFO within the nearest future. For me it is fine if this feature will be added in a later PR.

I have some different design ideas for the FIFO interface that I will like your view.
 1. Implement a template FifoData class that takes the FIFO buffer size as template argument and inherits from Data. The Data class should then implement functions for retrieving the buffer size and a pointer to the buffer. When constructing the Iim42562 the FifoData class should then be passed instead of the Data class.
 2. Implement a template FifoData class that takes the FIFO buffer size as template argument but which does NOT inherit from Data. This class should then be passed to a readFifoData function on Iim42652. However, I think this is might be an awkward solution as the data is not handled in one class.
 3. Add a FIFO buffer to the existing Data class, and set the FIFO buffer size using an argument passed to lbuild. Thereby the FIFO buffer and size can be retrieved similar to the sensorData currently is retrieved.
 4. Provide the Iim42562 class with an additional template argument to provide the FIFO buffer size. However, I am not in favor of this idea as I find it a bit awkward.

In all circumstances I will add an iterator to the class that handles FIFO data for parsing the data and return parsed FIFO frames. I hope my considerations make sense and thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -751,68 +751,69 @@ your specific needs.
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-is31fl3733">IS31FL3733</a></td>
 </tr><tr>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-itg3200">ITG3200</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ixm42xxx">IXM42XXX</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-l3gd20">L3GD20</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lan8720a">LAN8720A</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lawicel">LAWICEL</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lis302dl">LIS302DL</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-lis3dsh">LIS3DSH</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-lis3dsh">LIS3DSH</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lis3mdl">LIS3MDL</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lm75">LM75</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lp503x">LP503x</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lsm303a">LSM303A</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-lsm6ds33">LSM6DS33</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-lsm6dso">LSM6DSO</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-lsm6dso">LSM6DSO</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-ltc2984">LTC2984</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-max31855">MAX31855</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-max31865">MAX31865</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-max6966">MAX6966</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-max7219">MAX7219</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-mcp23x17">MCP23x17</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-mcp23x17">MCP23x17</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-mcp2515">MCP2515</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-mcp3008">MCP3008</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-mcp7941x">MCP7941x</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-mcp990x">MCP990X</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-mmc5603">MMC5603</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-ms5611">MS5611</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ms5611">MS5611</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-ms5837">MS5837</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-nokia5110">NOKIA5110</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-nrf24">NRF24</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-parallel_tft_display">TFT-DISPLAY</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pat9125el">PAT9125EL</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-pca8574">PCA8574</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-pca8574">PCA8574</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pca9535">PCA9535</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pca9548a">PCA9548A</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pca9685">PCA9685</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sh1106">SH1106</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-siemens_s65">SIEMENS-S65</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-siemens_s75">SIEMENS-S75</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-siemens_s75">SIEMENS-S75</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sk6812">SK6812</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sk9822">SK9822</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-ssd1306">SSD1306</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-st7586s">ST7586S</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-st7789">ST7789</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-stts22h">STTS22H</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-stts22h">STTS22H</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-stusb4500">STUSB4500</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sx1276">SX1276</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-tcs3414">TCS3414</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-tcs3472">TCS3472</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-tlc594x">TLC594x</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-tmp102">TMP102</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-tmp102">TMP102</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-tmp12x">TMP12x</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-tmp175">TMP175</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-touch2046">TOUCH2046</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-vl53l0">VL53L0</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-vl6180">VL6180</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-ws2812">WS2812</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ws2812">WS2812</a></td>
 </tr>
 </table>
 <!--/drivertable-->
--- a/examples/nucleo_g474re/ixm42xxx/main.cpp
+++ b/examples/nucleo_g474re/ixm42xxx/main.cpp
@@ -1,0 +1,100 @@
+// coding: utf-8
+/*
+ * Copyright (c) 2023, Rasmus Kleist Hørlyck Sørensen
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/driver/inertial/ixm42xxx.hpp>
+#include <modm/math.hpp>
+#include <modm/platform.hpp>
+#include <modm/processing.hpp>
+
+using SpiMaster = modm::platform::SpiMaster1;
+using Mosi = GpioA7;
+using Miso = GpioA6;
+using Sck = GpioA5;
+using Cs = GpioC5;
+
+class ImuThread : public modm::pt::Protothread, public modm::ixm42xxx
+{
+    using Transport = modm::Ixm42xxxTransportSpi< SpiMaster, Cs >;
+
+public:
+    ImuThread() : imu(data), timer(std::chrono::milliseconds(500)) {}
+
+    bool
+    run()
+    {
+        PT_BEGIN();
+
+        /// Initialize the IMU and verify that it is connected
+        PT_CALL(imu.initialize());
+        while (not PT_CALL(imu.ping()))
+        {
+            MODM_LOG_ERROR << "Cannot ping IXM42xxx" << modm::endl;
+            PT_WAIT_UNTIL(timer.execute());
+        }
+
+        /// Configure data sensors
+        PT_CALL(imu.updateRegister(Register::GYRO_CONFIG0, GyroFs_t(GyroFs::dps2000) | GyroOdr_t(GyroOdr::kHz1)));
+        PT_CALL(imu.updateRegister(Register::ACCEL_CONFIG0, AccelFs_t(AccelFs::g16) | AccelOdr_t(AccelOdr::kHz1)));
+        PT_CALL(imu.updateRegister(Register::PWR_MGMT0, GyroMode_t(GyroMode::LowNoise) | AccelMode_t(AccelMode::LowNoise)));
+
+        while (true)
+        {
+            PT_WAIT_UNTIL(timer.execute());
+            PT_CALL(imu.readSensorData());
+
+            data.getTemp(&temp);
+            data.getAccel(&accel);
+            data.getGyro(&gyro);
+
+            MODM_LOG_INFO.printf("Temp: %.3f\n", temp);
+            MODM_LOG_INFO.printf("Accel: (%.3f, %.3f, %.3f)\n", accel.x, accel.y, accel.z);
+            MODM_LOG_INFO.printf("Gyro: (%.3f, %.3f, %.3f)\n", gyro.x, gyro.y, gyro.z);
+        }
+        PT_END();
+    }
+
+private:
+    float temp;
+    modm::Vector3f accel;
+    modm::Vector3f gyro;
+
+    modm::ixm42xxxdata::Data data;
+    modm::Ixm42xxx< Transport > imu;
+
+    modm::PeriodicTimer timer;
+
+} imuThread;
+
+int
+main()
+{
+    Board::initialize();
+
+    Cs::setOutput(modm::Gpio::High);
+    SpiMaster::connect<Mosi::Mosi, Miso::Miso, Sck::Sck>();
+    SpiMaster::initialize<Board::SystemClock, 21.5_MHz>();
+
+    MODM_LOG_INFO       << "==========IXM-42xxx Test==========" << modm::endl;
+    MODM_LOG_DEBUG      << "Debug logging here" << modm::endl;
+    MODM_LOG_INFO       << "Info logging here" << modm::endl;
+    MODM_LOG_WARNING    << "Warning logging here" << modm::endl;
+    MODM_LOG_ERROR      << "Error logging here" << modm::endl;
+    MODM_LOG_INFO       << "==================================" << modm::endl;
+
+    while (true)
+    {
+        imuThread.run();
+    }
+
+    return 0;
+}
--- a/examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
+++ b/examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
@@ -1,0 +1,148 @@
+// coding: utf-8
+/*
+ * Copyright (c) 2023, Rasmus Kleist Hørlyck Sørensen
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/driver/inertial/ixm42xxx.hpp>
+#include <modm/platform.hpp>
+#include <modm/processing.hpp>
+
+namespace
+{
+    volatile bool interrupt = false;
+}
+
+using SpiMaster = modm::platform::SpiMaster1;
+using Mosi = GpioA7;
+using Miso = GpioA6;
+using Sck = GpioA5;
+using Cs = GpioC5;
+
+using Int1 = GpioC3;
+
+class ImuThread : public modm::pt::Protothread, public modm::ixm42xxx
+{
+    using Transport = modm::Ixm42xxxTransportSpi< SpiMaster, Cs >;
+
+public:
+    ImuThread() : imu(fifoData), timer(std::chrono::milliseconds(500)) {}
+
+    bool
+    run()
+    {
+        PT_BEGIN();
+
+        Int1::setInput(modm::platform::Gpio::InputType::PullDown);
+        Exti::connect<Int1>(Exti::Trigger::RisingEdge, [](uint8_t)
+        {
+            interrupt = true;
+        });
+
+        /// Initialize the IMU and verify that it is connected
+        PT_CALL(imu.initialize());
+        while (not PT_CALL(imu.ping()))
+        {
+            MODM_LOG_ERROR << "Cannot ping IXM-42xxx" << modm::endl;
+            PT_WAIT_UNTIL(timer.execute());
+        }
+
+        MODM_LOG_INFO << "IXM-42xxx Initialized" << modm::endl;
+        MODM_LOG_INFO << "Fifo Buffer Size: " << fifoData.getFifoBufferSize() << modm::endl;
+
+        /// Configure FIFO
+        PT_CALL(imu.updateRegister(Register::FIFO_CONFIG, FifoMode_t(FifoMode::StopOnFull)));
+        PT_CALL(imu.updateRegister(Register::FIFO_CONFIG1, FifoConfig1::FIFO_HIRES_EN | FifoConfig1::FIFO_TEMP_EN | FifoConfig1::FIFO_GYRO_EN | FifoConfig1::FIFO_ACCEL_EN));
+        PT_CALL(imu.writeFifoWatermark(1024));
+
+        /// Configure interrupt
+        PT_CALL(imu.updateRegister(Register::INT_CONFIG, IntConfig::INT1_MODE | IntConfig::INT1_DRIVE_CIRCUIT | IntConfig::INT1_POLARITY));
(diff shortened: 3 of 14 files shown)


salkinium:
Thanks for the driver!

> Therefore I do not know if Iim42652 is the right name for the device?

You can add the compatible devices into the module description then it is also findable via the homepage.

1. There is no real standard. For a while we thought that CamelCase would be great, but when converting ABBREV register names, it sometimes becomes unclear which letters to capitalize and which not.  I would recommend to use the convention in the datasheet, at least this is canonical.
2. Yeah, sure. You could add a [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) to the lbuild docs to make it clear that this intentional. That makes it clearer for other developers.
3. No, leave it in and put a warning in the lbuild docs via admonition that it's untested.
4. I'd probably prefer option 1, however, you can also allocate the Fifo buffer externally and pass it to the Data class, which can deduct the size and store it as a `std::span`. So effectively a runtime mechanism.

I'll review in detail before the end of the week this still gets into the 23Q2 release.

rasmuskleist:
Thanks for the nice comments! I have addressed point 1, 2 and 3 in the latest commits. I plan to address point 4 tomorrow with my first suggested design. Also I was reading the documentation for ICM-42688-P in greater detail and found that some registers are not identically the same as for IIM-42652. For example the APEX_CONFIG0 is a little different between the two devices. Therefore I have not added an admonition that the driver is compatible with e.g. ICM-42688-P. Is this still relevant or should I leave it as is?

rasmuskleist:
I have started the FIFO implementation and have had some problems with the endianess of the imu which I will have to look into. Specifically, I configure the device to use little endian for both sensor data and fifo count. However, setting the fifo count to use little endian, seem to have the oppostive effect. Every time that I have configured the imu to use little endian and have used (FIFO_COUNTH << 8) | FIFO_COUNTL get a byte count in excess of 4000. That is despite the FIFO only having 2K bytes. I will have to look into this tomorrow.

rasmuskleist:
I have had some trouble getting the FIFO to work. The trouble was caused by setting the FIFO_WM_GT_TH bit in FIFO_CONFIG1 or setting TMST_EN in TMST_CONFIG. Altough I would not have expected setting these bits high would have caused the FIFO_DATA register to contain zero readings. However, currently I am getting readouts through FIFO so now I only need to parse the data, which I have already started.

rasmuskleist:
I have been a bit busy the last week and hence the slow progress. I have now implmented a parser for parsing the fifo data and an iterator which uses the parser to iterate through the fifo data. I will now have to test that the iterator works as intended. Also I will have to figure out if there is a nice way to pass the scale onto the FifoPacket class or if this should be up to the user to provide the scale.

examples/nucleo_g474re/ixm42xxx/main.cpp
```diff
@@ -0,0 +1,104 @@
...
+} imuThread;
+
+int
+main()
+{
+    Board::initialize();
+    Cs::setOutput(modm::Gpio::High);
+
+    SpiMaster::connect<Miso::Miso, Sck::Sck>();
+    SpiMaster::initialize<Board::SystemClock, 21.5_MHz>();
```
salkinium, line 85: Already done in the thread.

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+    run()
+    {
+        PT_BEGIN();
+
+        Cs::setOutput(modm::Gpio::High);
+        SpiMaster::connect<Mosi::Mosi, Miso::Miso, Sck::Sck>();
+        SpiMaster::initialize<Board::SystemClock, 21.5_MHz>();
+
+        Int1::setInput(modm::platform::Gpio::InputType::PullDown);
+        Exti::connect<Int1>(Exti::Trigger::RisingEdge, [count=uint32_t(0)](uint8_t) mutable
```
salkinium:
```suggestion
        Exti::connect<Int1>(Exti::Trigger::RisingEdge, [](uint8_t)
```

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+#include <modm/driver/inertial/ixm42xxx.hpp>
+#include <modm/platform.hpp>
+#include <modm/processing.hpp>
+
+namespace
+{
+    bool interrupt = false;
```
salkinium:
```suggestion
    volatile bool interrupt = false;
```

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+                }
+                intStatus.reset(IntStatus::FIFO_FULL_INT | IntStatus::FIFO_THS_INT);
+            }
+        }
+        PT_END();
+    }
+
+private:
+
+    /// Due to the non-deterministic nature of system operation, driver memory allocation should always be the largest size of 2080 bytes.
```
salkinium, line 117: why is it configurable then?

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+} imuThread;
+
+int
+main()
+{
+    Board::initialize();
+    Cs::setOutput(modm::Gpio::High);
+
+    SpiMaster::connect<Miso::Miso, Sck::Sck>();
+    SpiMaster::initialize<Board::SystemClock, 21.5_MHz>();
```
salkinium, line 133: Already done

src/modm/driver/inertial/ixm42xxx.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * The transport must be initialized by the user!
+     * @warning The SPI frequency must not exceed 2 MHz for this chip,
+     * or 1 MHz with burst mode.
+     */
+    modm::ResumableResult<void>
+    initialize();
+
+    /**
+     * @brief Read the temperature data from the device
+     * @return False in case of any error, e.g. if some register acces is not
```
salkinium:
```suggestion
     * @return False in case of any error, e.g. if some register access is not
```

salkinium: And everywhere below ;-p

src/modm/driver/inertial/ixm42xxx.hpp
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * @brief Read a single register.
+     *
+     * @param reg The register to be read
+     * @param value The placeholder for the read value to be stored
+     * @return False in case of any error, e.g. if some register acces is not
+     * permitted.
+     */
+    modm::ResumableResult<bool>
+    readRegister(Register reg, uint8_t &value);
```
salkinium:
```suggestion
    readRegister(Register reg, uint8_t *value);
```
Makes it clearer at the call site that value will be modified.

src/modm/driver/inertial/ixm42xxx.lb
```diff
@@ -0,0 +1,46 @@
...
+def init(module):
+    module.name = ":driver:ixm42xxx"
+    module.description = """\
+# InvenSense 6-Axis IMU
+
+High Precision 6-Axis MEMS MotionTracking Device
+
+!!! note ""
+
+    The registers in user bank 1, 2, 3 and 4 are currently not mapped out
```
salkinium:
```suggestion
!!! note "The registers in user bank 1, 2, 3 and 4 are currently not mapped out"
```

salkinium: And below

README.md
```diff
@@ -751,68 +751,70 @@ your specific needs.
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-is31fl3733">IS31FL3733</a></td>
 </tr><tr>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-itg3200">ITG3200</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ixm42xxx">IXM42XXX</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ixm42xxx-transport">IXM42XXX-TRANSPORT</a></td>
```
salkinium: Please filter this [out here](https://github.com/modm-io/modm/blob/develop/tools/scripts/synchronize_docs.py#L79).

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,251 @@
...
+    ///@{
+    int16_t
+    getTemp()
+    { return temp; }
+
+    void
+    getTemp(int16_t &temp)
+    { temp = getTemp(); }
+
+    Vector<int32_t, 3>
```
salkinium:
```suggestion
    Vector3i
```
?

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,251 @@
...
+    float accelScale;
+    float gyroScale;
+};
+
+/// @ingroup modm_driver_ixm42xxx
+struct
+FifoPacket
+{
+
+    FifoPacket() : header(0), accel(0), gyro(0), temp(0), timestamp(0), extension(0) {}
```
salkinium, line 123:
```suggestion
    constexpr FifoPacket() : header(0), accel(0), gyro(0), temp(0), timestamp(0), extension(0) {}
```
You can probably mark all functions constexpr.

src/modm/driver/inertial/ixm42xxx_data_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+    else
+    {
+        // FIFO is empty
+        fifoPacket = FifoPacket();
+        fifoIndex = fifoData.size();
+    }
+
+    return fifoIndex;
+}
```
salkinium, line 121: Fancy fancy :-)

src/modm/driver/inertial/ixm42xxx_definitions.hpp
```diff
@@ -0,0 +1,934 @@
...
+    };
+    MODM_FLAGS8(DeviceConfig);
+
+    enum class
+    DataMode : uint8_t
+    {
+        Mode0 = 0,
+        Mode1 = int(DeviceConfig::SPI_MODE),
+        Mode2 = int(DeviceConfig::SPI_MODE),
+        Mode3 = 0,
```
salkinium, line 241: This is correct?

src/modm/driver/inertial/ixm42xxx_definitions.hpp
```diff
@@ -0,0 +1,934 @@
...
+
+        /// USER BANK 3 REGISTER MAP START
+        /// *** ORed with 0x300 to indicate user bank 3 ***
+        PU_PD_CONFIG1 = 0x306,
+        PU_PD_CONFIG2 = 0x30E,
+        /// USER BANK 3 REGISTER MAP END
+
+        /// USER BANK 4 REGISTER MAP START
+        /// *** ORed with 0x400 to indicate user bank 4 ***
+        /// Apex configuration registers
```
salkinium: These will all end up in the doxygen somewhere, just fyi, you probably only want two slashes here.

src/modm/driver/inertial/ixm42xxx_impl.hpp
```diff
@@ -0,0 +1,265 @@
...
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_IXM42XXX_HPP
+#   error "Don't include this file directly, use 'ixm42xxx.hpp' instead!"
+#endif
+
+#include "ixm42xxx.hpp"
+
+#include <modm/math/utils/bit_constants.hpp>
+#include <modm/debug.hpp>
```
salkinium:
```suggestion
```

salkinium (commented):
Very thorough driver, impressive! Just some minor nitpicks from me.

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
rasmuskleist: Good point! This comment is taken from the fifo documentation of the device. I believe it is for use with the FIFO_FULL_INT, when 2080 bytes of memory can be expected. However, for my own use case I have three imus so having three 2080 byte arrays may be a bit too much, especially if there is a lot of application code. What I may do is to set the template argument default 2080?

examples/nucleo_g474re/ixm42xxx/main.cpp
rasmuskleist: True! I believe the correct place to do this is in main so that a SpiMaster is only initialized once if multiple threads are used?

src/modm/driver/inertial/ixm42xxx_definitions.hpp
rasmuskleist: Screenshot 2023-07-11 at 08 22 44
I believe this should be correct!

src/modm/driver/inertial/ixm42xxx_definitions.hpp
rasmuskleist: I should have removed these doxygen comments now

src/modm/driver/inertial/ixm42xxx_data_impl.hpp
rasmuskleist: Thanks!

rasmuskleist:
Thanks for the nice comments! It has taken a lot of energy getting the driver to this state. I choose to rename the driver as in this [repo](https://github.com/InvenSenseInc/public.mcu.iim42652/tree/master/Ixm42xxx). I believe this is fine? I might ad an admonition that the driver is based on iim42562 and that some registers might be missing or not applicable to other InvenSense devices?

What is missing now is to somehow pass the accel and gyro scale onto the FifoPacket or in another way implement code that can convert the raw readings to some scale with physical meaning. The obvious choice is to have FifoData pass it onto the iterator which passes it onto the FifoPacket. However, I think that is a bit clumpsy. Do you have any suggestions?

In addition, I have yet to implement support for APEX data. However, this outside of my personal use case, so I prefer to add it at later.

Finally, I was debating if I should remove the ixm42xxx.transport module and let the ixm42xxx module copy the transport files? I believe the intention behind lis3.transport was to make it available to multiple drivers for example lis3dsh and lis302dl, which have different implementations.

salkinium:
The renaming is fine, especially with documentation about possible limitations and tested versions.

You can remove the separate transport module, we can add it back if needed, just leave the classes in separate files.

I'm not sure about the how to pass the scale, I don't think you have much choice unless you convert it before passing it to the packet.

rasmuskleist:
I now consider this ready for merge unless you have any additional changes? Before merging I suppose that I will change up the git history to have:

[math] Adding long vector typedefs
[driver] Adding InvenSense 6-Axis IMU driver
[example] Adding InvenSense 6-Axis IMU example
[driver] Adding InvenSense FIFO support
[example] Adding InvenSense FIFO example

I have choosen to implement the FifoPacket scaling methods by passing the scale to the method. This servers two purposes. First of all I avoid having to pass the scales twice. More importantly the scales are stored only in one place, so there are not multiple sources of truth. What do you prefer?

Also I should add that the when compiling I get -Wdouble-promotion warning from the printf statements. I suppose that I should simply ignore this?

chris-durand:
> Also I should add that the when compiling I get -Wdouble-promotion warning from the printf statements. I suppose that I should simply ignore this?

If you pass parameters to a C-style variadic function like `printf` they undergo "default argument promotions" and `float` is converted to `double`. We might have to turn off that warning. There would be ways to explicitly suppress the warning but it would be way too verbose and noisy to put that around every `printf` call. Just ignore it for now.

examples/nucleo_g474re/ixm42xxx/main.cpp
salkinium: Yes, then the one in the protothread should be removed.

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
salkinium: Yes, that's a good option.

README.md
rasmuskleist: Is this still relevant with the removal of the transport module?

salkinium: No

src/modm/driver/inertial/ixm42xxx_data.hpp
rasmuskleist: For FIFO high resolution I need 18 bits for accel and 19 bits for gyro. Hence int16_t in Vector3i is insufficent. Therefore I have added Vector3li which uses int32_t. It is possibly not the right PR to do it?

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
rasmuskleist: I tried to do this, but when creating an instance of FifoData, I would then have to write FifoData<> to take the default template argument. Otherwise I get the error: invalid use of template-name 'modm::ixm42xxxdata::FifoData' without an argument list. Should I just leave it as is?

src/modm/driver/inertial/ixm42xxx.hpp
rasmuskleist: I completely overlooked that one...

src/modm/driver/inertial/ixm42xxx.hpp
rasmuskleist: I have changed this, but I believe it is a bit inconsistent now since many other "getter" methods use the reference e.g. readFifoCount.

examples/nucleo_g474re/ixm42xxx_fifo/main.cpp
salkinium: Yes.

salkinium (approved):
Excellent driver! Please rebase/squash as you described.

src/modm/driver/inertial/ixm42xxx.hpp
rasmuskleist: Should we use pointers in this one instance?

salkinium: Sorry, yes, ideally we should always use pointers for arguments that get written to make it clear at the callsite that the variable gets written.
Unfortunately we messed it up with our Uart::read() interface function and it's been getting copied ever since. But let's do it right now, so I don't have to refactor it later.

salkinium: Are you ok with changing all the &args to *args also in the data class?

rasmuskleist: I have done this now. Should Data still be passed to Ixm42xxx with reference?

salkinium: You can make the Data object a member variable of the driver. The separate passing of the data object was only for packing a bunch of data objects into a packed struct and sending that over the wire as an attempt to serialize very cheaply. However, this only worked for a specific use case and is otherwise kinda dumb and annoying.

rasmuskleist: The only problem with making it a member variable is that one should specify if it is Data or FifoData and provide the fifo buffer size if it is FifoData. In that case the driver should take the fifo buffer size as template argument

salkinium: Yeah right. Let's just leave it as it is, and I'll survey all drivers in the future and come up with a proper solution for this.

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+        fifoBuffer(fifoBuffer),
+        accelScale(16.f),
+        gyroScale(2000.f)
+    {
+    }
+
+    // DATA ACCESS
+    ///@{
+    int16_t inline
+    getTemp()
```
chris-durand: Shouldn't all these acessors be `const`?
```suggestion
    getTemp() const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { return sensorData.temp; }
+
+    void inline
+    getTemp(int16_t *temp)
```
chris-durand:
```suggestion
    getTemp(int16_t *temp) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { *temp = getTemp(); }
+
+    void inline
+    getTemp(float *temp)
```
chris-durand:
```suggestion
    getTemp(float *temp) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { float t = getTemp(); *temp = t / 132.48f + 25.f; }
+
+    Vector3i inline
+    getAccel()
```
chris-durand:
```suggestion
    getAccel() const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { return Vector3i(sensorData.accel); }
+
+    void inline
+    getAccel(Vector3i *accel)
```
chris-durand:
```suggestion
    getAccel(Vector3i *accel) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { Vector3li a = getAccel(); *accel = a; }
+
+    void inline
+    getAccel(Vector3f *accel)
```
chris-durand:
```suggestion
    getAccel(Vector3f *accel) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { Vector3f a = getAccel(); a *= accelScale / INT16_MAX; *accel = a; }
+
+    Vector3i inline
+    getGyro()
```
chris-durand:
```suggestion
    getGyro() const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { return Vector3i(sensorData.gyro); }
+
+    void inline
+    getGyro(Vector3i *gyro)
```
chris-durand:
```suggestion
    getGyro(Vector3i *gyro) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { *gyro = getGyro(); }
+
+    void inline
+    getGyro(Vector3f *gyro)
```
chris-durand:
```suggestion
    getGyro(Vector3f *gyro) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+/// @ingroup modm_driver_ixm42xxx
+struct
+FifoPacket
+{
+    FifoPacket() : header(0), accel(0), gyro(0), temp(0), timestamp(0), extension(0) {}
+
+    // DATA ACCESS
+    ///@{
+    int16_t
+    getTemp()
```
chris-durand:
```suggestion
    getTemp() const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { return temp; }
+
+    void
+    getTemp(int16_t *temp)
```
chris-durand:
```suggestion
    getTemp(int16_t *temp) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { *temp = getTemp(); }
+
+    void
+    getTemp(float *temp)
```
chris-durand:
```suggestion
    getTemp(float *temp) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { float t = getTemp(); *temp = (header & HEADER_20) ? t / 132.48f + 25.f : t / 2.07f + 25.f; }
+
+    Vector3li
+    getAccel();
```
chris-durand:
```suggestion
    getAccel() const;
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+    void
+    getAccel(Vector3li *accel)
```
chris-durand:
```suggestion
    getAccel(Vector3li *accel) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { Vector3li a = getAccel(); *accel = a; }
+
+    void
+    getAccel(Vector3f *accel, float scale = 16.f)
```
chris-durand:
```suggestion
    getAccel(Vector3f *accel, float scale = 16.f) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { Vector3f a = getAccel(); *accel = (header & HEADER_20) ? a / 8192.f : a * scale / INT16_MAX; }
+
+    Vector3li
+    getGyro();
```
chris-durand:
```suggestion
    getGyro() const;
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+    void
+    getGyro(Vector3li *gyro)
```
chris-durand:
```suggestion
    getGyro(Vector3li *gyro) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { Vector3li g = getGyro(); *gyro = g; }
+
+    void
+    getGyro(Vector3f *gyro, float scale = 2000.f)
```
chris-durand:
```suggestion
    getGyro(Vector3f *gyro, float scale = 2000.f) const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+    { Vector3f g = getGyro(); *gyro = (header & HEADER_20) ? g / 131.f : g * scale / INT16_MAX; }
+
+    uint16_t
+    getTimestamp()
```
chris-durand:
```suggestion
    getTimestamp() const
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+        HEADER_20 = Bit4,
+        HEADER_TIMESTAMP_ODR = Bit3,
+        HEADER_TIMESTAMP_FSYNC = Bit2,
+        HEADER_ODR_ACCEL = Bit1,
+        HEADER_ODR_GYRO = Bit0,
+    };
+
+public:
+    constexpr bool operator==(const FifoPacket& rhs) const;
+    constexpr bool operator!=(const FifoPacket& rhs) const;
```
chris-durand: Since C++20 there is no need to define `operator!=` anymore. The compiler will automatically define it in terms of `operator==`.
```suggestion
```

src/modm/driver/inertial/ixm42xxx_data_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+FifoPacket::operator!=(const FifoPacket& rhs) const
+{
+    return ((this->header    != rhs.header)             ||
+            (this->temp      != rhs.temp)               ||
+            (this->timestamp != rhs.timestamp)          ||
+            (!std::equal(accel, accel + 3, rhs.accel))  ||
+            (!std::equal(gyro, gyro + 3, rhs.gyro))     ||
+            (!std::equal(extension, extension + 3, rhs.extension)));
+}
+
```
chris-durand:
```suggestion
```

src/modm/driver/inertial/ixm42xxx_data.hpp
```diff
@@ -0,0 +1,269 @@
...
+        constexpr iterator(std::span<const uint8_t> data, uint16_t index) : fifoData(data), fifoIndex(index)
+        {
+            fifoIndex = FifoPacket::parse(fifoData, fifoPacket, fifoIndex);
+        }
+
+        constexpr iterator operator++() { fifoIndex = FifoPacket::parse(fifoData, fifoPacket, fifoIndex); return *this; }
+        constexpr reference operator*() { return fifoPacket; }
+        constexpr pointer operator->() { return &fifoPacket; }
+        constexpr bool operator==(const iterator& rhs) { return fifoIndex == rhs.fifoIndex && fifoPacket == rhs.fifoPacket; }
+        constexpr bool operator!=(const iterator& rhs) { return fifoIndex != rhs.fifoIndex || fifoPacket != rhs.fifoPacket; }
```
chris-durand:
```suggestion
```

chris-durand (commented):
Thank you!

src/modm/driver/inertial/ixm42xxx_data.hpp
rasmuskleist: Yes I agree!

src/modm/driver/inertial/ixm42xxx_data.hpp
rasmuskleist: I have removed the operator!= now

src/modm/driver/inertial/ixm42xxx.hpp
```diff
@@ -0,0 +1,196 @@
...
+    Ixm42xxx(Data &data, uint8_t address = 0b11010000);
+
+    /**
+     * @brief Initialize
+     *
+     * Sets used pins as output.
+     *
+     * The transport must be initialized by the user!
+     * @warning The SPI frequency must not exceed 2 MHz for this chip,
+     * or 1 MHz with burst mode.
```
chris-durand: This is copy-paste from the ADIS16470 driver, isn't it :smiley:? None of the things in the comment applies for this device and driver.

rasmuskleist: Ups yeah, this is unintentional! I have changed the comment for the initialize function.

chris-durand (approved):
Thanks!

PR #1054, opened 2023-08-03
[driver] Adding SpiStack Flash

rasmuskleist:
I have changed the implementation of the BdSpiFlash to use the W25M512JV instruction set as discussed in #890. In addition, I have added a wrapper that makes the stacked die seem like a contiguous piece of memory for homogenous storage devices. This was also discussed in #890, although my implementation is somewhat different. 

Currently, the dieSelect function does more than I probably want it to, so I would like your opinion on the following alternative implementation.

1. Set/reset ADP bit before device reset inside the initialize function. I believe that this should cause all dies to enter the desired addressing mode upon device reset. However, not all devices implement WSR3 command, but in those cases this will probably be a NOP. Alternatively, the ADP bit can be set only for devices larger than 16MB. 
2. Do not reset the device inside the initialize function and implement a seperate reset function. With the stacked dies a reset casues all dies to enter the addressing mode specified by the ADP bit. In the initialize function the device should then enter 4B addressing, and inside BdSpiStack, a loop should then initialize all dies. Here En4BAM is only called when the device size is larger than 16MB
3. Enter/Exit 4B addressing on every die select depending on the device size as currently implemented. This does add some overhead on each dieSelect.
4. Use 4B commands e.g. PP4BA for devices larger than 16MB. This makes it very explicit to the device that 4B addressing is being used. However, logic will have to be implemented in both the spiOperations function and the read, program and erase functions.
5. Use extended address register for 4B addressing. This is only for backward compatability with devices prior to 2011, so I think it is not worth considering.

Finally, the current implementation waits for the BUSY bit to clear after issuing a command. Although I do find it nice that the operation finishes before returning, it does block the driver from issuing commands that can be done while busy. For example with the SpiStack concurrent operations can be performed. Therefore I like your opinion on waiting for the BUSY bit to clear in the program and erase function before issuing the command? Thereby I can select another die and double the troughput. However, I should add that the performance gain is probably very small in practice unless I change the BdSpiStack addressing to be modulo erase block. This could possibly double the throughput, but since I do not know my own throughput at the moment, this is possibly better suited for a later PR!

### DIFF
--- a/src/modm/driver/storage/block_device_spistack_flash.hpp
+++ b/src/modm/driver/storage/block_device_spistack_flash.hpp
@@ -1,0 +1,125 @@
+// coding: utf-8
+/*
+ * Copyright (c) 2018, Raphael Lehmann
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_BLOCK_DEVICE_SPISTACK_FLASH_HPP
+#define MODM_BLOCK_DEVICE_SPISTACK_FLASH_HPP
+
+#include <algorithm>
+#include <cstdlib>
+
+#include <modm/architecture/interface/block_device.hpp>
+#include <modm/processing/resumable.hpp>
+
+namespace modm
+{
+
+/**
+ * \brief	SpiStack homogenoues memory
+ *
+ * The `read()`, `erase()`,`program()` and `write()` methodes wait for
+ * the chip to finish writing to the flash.
+ *
+ * \tparam SpiBlockDevice		Base SPI block device of the homogenous stack
+ *
+ * \ingroup	modm_driver_block_device_spi_stack_flash
+ * \author	Rasmus Kleist Hørlyck Sørensen
+ */
+template <typename SpiBlockDevice, uint8_t DieCount>
+class BdSpiStackFlash : public modm::BlockDevice, protected NestedResumable<3>
+{
+public:
+	/// Initializes the storage hardware
+	modm::ResumableResult<bool>
+	initialize();
+
+	/// Deinitializes the storage hardware
+	modm::ResumableResult<bool>
+	deinitialize();
+
+	/** Read data from one or more blocks
+	 *
+	 *  @param buffer	Buffer to read data into
+	 *  @param address	Address to begin reading from
+	 *  @param size		Size to read in bytes (multiple of read block size)
+	 *  @return			True on success
+	 */
+	modm::ResumableResult<bool>
+	read(uint8_t* buffer, bd_address_t address, bd_size_t size);
+
+	/** Program blocks with data
+	 *
+	 *  Any block has to be erased prior to being programmed
+	 *
+	 *  @param buffer	Buffer of data to write to blocks
+	 *  @param address	Address of first block to begin writing to
+	 *  @param size		Size to write in bytes (multiple of read block size)
+	 *  @return			True on success
+	 */
+	modm::ResumableResult<bool>
+	program(const uint8_t* buffer, bd_address_t address, bd_size_t size);
+
+	/** Erase blocks
+	 *
+	 *  The state of an erased block is undefined until it has been programmed
+	 *
+	 *  @param address	Address of block to begin erasing
+	 *  @param size		Size to erase in bytes (multiple of read block size)
+	 *  @return			True on success
+	 */
+	modm::ResumableResult<bool>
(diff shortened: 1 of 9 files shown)


src/modm/driver/storage/block_device_spistack_flash.hpp
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * The `read()`, `erase()`,`program()` and `write()` methodes wait for
+ * the chip to finish writing to the flash.
+ *
+ * \tparam SpiBlockDevice		Base SPI block device of the homogenous stack
+ *
+ * \ingroup	modm_driver_block_device_spistack_flash
+ * \author	Rasmus Kleist Hørlyck Sørensen
+ */
+template <typename SpiBlockDevice, uint8_t dieCount>
```
salkinium:
```suggestion
template <typename SpiBlockDevice, uint8_t DieCount>
```

src/modm/driver/storage/block_device_spistack_flash.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * \brief	SpiStack homogenoues memory
+ *
+ * The `read()`, `erase()`,`program()` and `write()` methodes wait for
+ * the chip to finish writing to the flash.
+ *
+ * \tparam SpiBlockDevice		Base SPI block device of the homogenous stack
+ *
+ * \ingroup	modm_driver_block_device_spistack_flash
```
salkinium:
```suggestion
 * \ingroup	modm_driver_block_device_spi_stack_flash
```
I think the . in the module name is translated to _ by the doc generator. Please check if this class is in the right doxypress module in the example.

src/modm/driver/storage/block_device_spistack_flash_impl.hpp
```diff
@@ -0,0 +1,148 @@
...
+{
+	RF_BEGIN();
+
+	if((size == 0) || (size % BlockSizeRead != 0) || (address + size > DeviceSize)) {
+		RF_RETURN(false);
+	}
+
+	index = 0;
+	while (index < size) {
+		dv = std::ldiv(index + address, DieSize); // dv.quot = die #ID, dv.rem = die address
```
salkinium, line 57: elegant

salkinium (approved):
>  I would like your opinion on the following alternative implementation.

I don't have enough understanding of flash chips to make an informed decision. All alternatives read like gibberish to me… gotta pass that potato back to you 😛 

>  I like your opinion on waiting for the BUSY bit to clear in the program and erase function before issuing the command?

Could you flip it? Do an operation, then continue doing things until you *really must* wait again? That's how we typically do it in other drivers, basically moves the last wait in the function to the front if that makes sense here.

src/modm/driver/storage/block_device_spistack_flash.hpp
rasmuskleist: Ups yeah. This is also how its done in BdSpiFlash...

src/modm/driver/storage/block_device_spistack_flash_impl.hpp
rasmuskleist: Thanks! I definitely prefer this over integer division.

rasmuskleist:
I'm no expert on flash chips either! For the moment I will leave the selectDie function as is, unless you have any comments to it. Afterall it is working and the overhead is probably not that significant, especially compared to the erase/program durations.

I have flipped the waiting as suggested, and will confirm that it works with my W25M512VJ flash chip tonight. Then I will convert it to a PR. I should add that flipping it will add some overhead for example in the read fuction, but I believe it is worth it since the driver only waits when it has to. This will also make the flow of the tests seem strange as it can enter the memory test while still busy erasing the chip. I can ignore this as it is only aesthetic or I can make the busy function public and wait the threads manually. What do you think?

salkinium:
Yeah, you should probably expose the wait method, otherwise it'll be confusing.
You may also explicitly wait for completion sonetimes like when rebooting or going to sleep.

rasmuskleist:
I have made the `isBusy` and `waitWhileBusy` function available in `BdSpiFlash` and implemented similar functions in `BdSpiStackFlash`. Here `isBusy` returns `true` if any of the stacked dies is busy which i presume is the desired behaviour. I also noticed whilst doing some further testing of integrating `BdSpiFlash` with FatFs that the nesting level should be increased in order to use `write`.

Currently, I do not see any operation in the `BdSpiFlash` implementation that can wait for completion, which is not already waiting. After all, the intention is not to wait for erase and program to finish, so concurrent operations can take place inside `BdSpiStackFlash`.

salkinium (approved):
Very nice driver! Please squash and then I'll merge it.

rasmuskleist:
Thanks! Should I squash more than this?

It occured to me as I was testing the driver with fibers, that it is problematic if a void functions returns a result and vice versa. I have now fixed this issue.

PR #1057, opened 2023-08-13
[example] Adding SpiFlash FatFs compatibility test

rasmuskleist:
I have experimented with integrating FatFs with the `BdSpiBlockDevice` using fibers as discussed in #893 and #814. The example is for testing the compatibility of the physical drive with FatFs and is adapted from [elm-chan](http://elm-chan.org/fsw/ff/res/app4.c). I have tested the example on the STM32G491MET6 on my local setup, but have added the example in the same location as the remaining flash examples.

Currently, the example blocks at `disk_read` in the "Single sector write test (unaligned buffer address)". I have tried to debug this issue and have found that `pbuff+3` in `dr = disk_write(pdrv, pbuff+3, lba, 1);` causes the subsequent call to `disk_read` to block inside `spiOperation()` and specifically at `RF_CALL(Spi::transfer(const_cast<uint8_t*>(tx), rx, length));`. I must admit that I am quite dumbfounded by this bug and do not really know where to start. Do you have any suggestions to why this bug is occring? At this point I am a little bit lost.

### DIFF
--- a/examples/nucleo_f429zi/spi_flash_fatfs/main.cpp
+++ b/examples/nucleo_f429zi/spi_flash_fatfs/main.cpp
@@ -1,0 +1,340 @@
+/*
+ * Copyright (c) 2023, Rasmus Kleist Hørlyck Sørensen
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <fatfs/ff.h>
+#include <fatfs/diskio.h>
+
+#include <modm/board.hpp>
+#include <modm/platform.hpp>
+#include <modm/processing.hpp>
+
+using namespace Board;
+using SpiMaster = SpiMaster1;
+
+// Spi flash chip wiring:
+using Cs = GpioA4;
+using Mosi = GpioB5;
+using Miso = GpioB4;
+using Sck = GpioB3;
+// Connect WP and HOLD pins to +3V3
+// and of course Vdd to +3V3 and Vss to GND
+
+/* The example is adapted from http://elm-chan.org/fsw/ff/res/app4.c */
+
+DWORD buff[FF_MAX_SS];              /* Working buffer (4 sector in size) */
+BYTE pdrv = 0;                      /* Physical drive number to be checked (all data on the drive will be lost) */
+UINT ncyc = 3;                      /* Number of test cycles */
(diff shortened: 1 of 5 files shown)


rasmuskleist:
I am not quite sure what the problem was, but my suggestion is that my local cmake setup did not add the `ffconf_local.hpp` to overwrite the `FF_MAX_SS`. As a consequence the buffer size was only 512b although I expected 4096. Why this first caused problems when adding to the buffer pointer, is probably because I had allocated a 4096 byte stack for the fiber, but this is just guessing. At least the example is now working and should demonstrate that FatFs can be used resumably with fibers!

Also, It occured to me that the flipping of the `waitWhileBusy()` in #1054 is wrong. It does not wait for program/erase to finish if multiple erase/program are performed in the loops. The test I wrote does not catch this mistake because it programs the flash in page size blocks... Sorry for this!

examples/nucleo_f429zi/spi_flash_fatfs/main.cpp
```diff
@@ -0,0 +1,328 @@
...
+        }
+        pns++;
+
+        MODM_LOG_INFO.printf("**** Test cycle %u of %u completed ****\n\n", cc, ncyc);
+    }
+
+    return 0;
+}
+
+modm::Fiber<> testFiber(test_diskio);
```
salkinium: Do you want to add another Fiber that schedules a blinking LED (or whatever you feel like). Otherwise the scheduler doesn't actually switch fibers (cos there's nowhere to switch to) and thus this is basically equivalent to `RF_CALL_BLOCKING`, as the scheduler just waits in place.

salkinium (commented):
Thanks for the super awesome example!!! It's fantastic that this actually works so well! (Sorry that I wasn't helpful here, full-time work keeps me quite busy)

> The test I wrote does not catch this mistake because it programs the flash in page size blocks... Sorry for this!

Don't sweat it, I made/make many mistakes too. You're PRs are always very high quality!

examples/nucleo_f429zi/spi_flash_fatfs/main.cpp
rasmuskleist: I have now added a blinky fiber, and everything works just fine! I should mention that the test finishes almost instantaneously so the test could finish before the blinking starts, and I would not notice. I presume this is not the case.

rasmuskleist:
Thanks for the nice comments! Yeah I am quite happy it was so smoothless. Next step will be to mount the drive and read/write files. I can possibly contribute a rudimentary example of this in a later PR if you wish?

salkinium (approved):

salkinium:
> I can possibly contribute a rudimentary example of this in a later PR if you wish?

Sure, that would be great!

PR #1059, opened 2023-08-23
Fix incorrect register indices in HMC58x3-based drivers

mbait:
Fixed according to the datasheet which can be found at https://cdn-shop.adafruit.com/datasheets/HMC5883L_3-Axis_Digital_Compass_IC.pdf


salkinium (approved):
Thanks! A typical hexadecimal vs decimal bug when copying.

PR #1062, opened 2023-08-29
Fix incorrect register names

mbait:
The fix does not really affect the driver behaviour, but the incorrect register names make a confusion regarding the need to convert from big-endian byte order as as-is the code implies little-endian order.


salkinium (commented):
I think you need to change this line too: https://github.com/modm-io/modm/blob/ee6dee0db9875b975a7ede0d066267a422019eca/src/modm/driver/inertial/hmc58x3_impl.hpp#L33

salkinium (approved):
Thanks!

PR #1063, opened 2023-08-29
[driver] Implement QMC5883L driver

mbait:
QMC5883L is a cheap Chinese implementation of HMC5883L which sought its popularity in DIY Arduino-based projects. Despite the very similar name QMC5883L share almost no common bits with HMC5883L - the register map and the logic vary a lot, so it would be not feasible to re-use the existing driver. The datasheet can be found at https://datasheet.lcsc.com/szlcsc/QST-QMC5883L-TR_C192585.pdf

Even though the chip can barely be considered for use in a serious project (in my humble opinion), people seeking to leave Arduino foundation for something more serious might benefit from this driver. Like I was looking for a way to squize maximal throughput out of that chip for which a non-blocking i2c implementation must be present.

### DIFF
--- a/src/modm/driver/inertial/qmc5883l.lb
+++ b/src/modm/driver/inertial/qmc5883l.lb
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2023, Alexander Solovets
+#
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+
+def init(module):
+    module.name = ":driver:qmc5883l"
+    module.description = """\
+# QMC5883L 3-Axis Digital Magnetometer
+
+The QMC5883L is a multi-chip three-axis magnetic sensor. This surface-mount,
+small sized chip has integrated magnetic sensors with signal condition ASIC,
+targeted for high precision applications such as compassing, navigation and
+gaming in drone, robot, mobile and personal hand-held devices.
+
+The QMC5883L is based on state-of-the-art, high resolution, magneto-resistive
+technology licensed from Honeywell AMR technology.  Along with custom-designed
+16-bit ADC ASIC, it offers the advantages of low noise, high accuracy, low
+power consumption, offset cancellation and temperature compensation. QMC5883L
+enables 1° to 2° compass heading accuracy. The I²C serial bus allows for easy
+interface.
+"""
+
+def prepare(module, options):
+    module.depends(
+        ":architecture:i2c.device",
+        ":architecture:register",
+        ":math:utils")
+    return True
+
+def build(env):
+    env.outbasepath = "modm/src/modm/driver/inertial"
+    env.copy("qmc5883l.hpp")
--- a/README.md
+++ b/README.md
@@ -789,30 +789,31 @@ your specific needs.
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pca9535">PCA9535</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pca9548a">PCA9548A</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-pca9685">PCA9685</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-qmc5883l">QMC5883L</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sh1106">SH1106</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-siemens_s65">SIEMENS-S65</a></td>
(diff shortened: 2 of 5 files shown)


src/modm/driver/inertial/qmc5883l.lb
```diff
@@ -0,0 +1,41 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+
+def init(module):
+    module.name = ":driver:qmc5883l"
+    module.description = """\
+# QMC5883L 3-Axis Digital Magnetometer
+
+The QMC5883L is a multi-chip three-axis magnetic sensor. This surface -mount,
```
salkinium:
```suggestion
The QMC5883L is a multi-chip three-axis magnetic sensor. This surface-mount,
```

src/modm/driver/inertial/qmc5883l.hh
salkinium, line 1: There's a mix of spaces and tabs in this file, please convert to tabs. (There's also a top-level `.clang-format` if you don't want to do it manually)

examples/avr/qmc5883l/main.cc
salkinium, line 1: Also contains a mix of tabs and spaces. => tabs only

src/modm/driver/inertial/qmc5883l.hh
```diff
@@ -0,0 +1,231 @@
...
+
+template <class I2cMaster>
+class Qmc5883l : public Qmc5883lRegisters, public modm::I2cDevice<I2cMaster>
+{
+public:
+  // This flag is supposed to be set from within an ISR. By the time when the
+  // first call to readData() occurs in the program, the DRDY bit will have
+  // been already set, so need to initialize this flag to "true" to enable
+  // first data read.
+  volatile bool dataReady = true;
```
salkinium:
```suggestion
  std::atomic_bool dataReady = true;
```

Also [see this comment as to why](https://github.com/modm-io/modm/pull/1050#discussion_r1264513015).

However, we typically let the application implement the synchronization mechanic to give the users the option to also use polling. So this flag would not be part of the driver.

src/modm/driver/inertial/qmc5883l.hh
```diff
@@ -0,0 +1,231 @@
...
+  /// @cond
+  Data &data;
+  /// @endcond
+	uint8_t buffer[sizeof data.data];
+
+public:
+	Qmc5883l(Data &data, uint8_t address=0x0d) :
+    modm::I2cDevice<I2cMaster> (address), data(data) {}
+
+  auto inline
```
salkinium: (inline is not necessary for template classes.)

src/modm/driver/inertial/qmc5883l.hh
```diff
@@ -0,0 +1,231 @@
...
+		RF_END_RETURN_CALL( this->runTransaction() );
+	}
+
+  modm::ResumableResult<bool>
+  readDataBlocking()
+  {
+    RF_BEGIN();
+    RF_WAIT_UNTIL(dataReady);
+    RF_END_RETURN_CALL(readData());
+  }
```
salkinium: Nah, let the user implement that externally, since there may be other ways of doing it.

salkinium (changes requested):

src/modm/driver/inertial/qmc5883l.hh
mbait: The idea behind this flag was to provide an automatic storage which, one way or another, has to be allocated somewhere. Without it a typical use case would involve one variable declaration for the class instance and one variable declaration for the flag. So even if the "ready" statuc is polled, the mechanic stays the same - set `dataReady` to _true_, call `readBlocking`. If, with this information in mind, you still don't want to see it - I'll remove it. Please, let me know.

salkinium: I appreciate the idea, however, you're mixing drivers with scheduling without knowing exactly how the application schedules things. Maybe a delay is required because multiple readings need to be synchronized, maybe its easier to poll the pin input trigger flag, rather than to use an interrupt to prevent an interrupt storm. It's much easier to just delegate this to the application layer and let the driver be "dumb". In addition, `Blocking` has been used in modm to wrap `RF_CALL_BLOCKING`, rather than to model scheduler timings. So just implement the synchronization in the example with something like `if (flag.exchange(false)) { driver.read(); }`.

mbait:
Addressed all the issues.

examples/avr/qmc5883l/main.cc
salkinium, line 1: main.cc -> main.cpp

src/modm/driver/inertial/qmc5883l.hh
salkinium, line 1: `.hh` -> `.hpp`

salkinium (changes requested):
Please change the file endings for consistency and since I'm not entirely sure if all our tooling works correctly with `.cc/.hh` file endings.

salkinium (approved):
Thank you very much!

salkinium:
Could you synchronize the documentation in the README and rebase on the latest develop branch?

salkinium (approved):

PR #1066, opened 2023-09-02
[stm32] EXTI fixes 

twast92:
Fix spurious EXTI interrupts with multiple enabled lines sharing the same IRQ
Fix compilation of Exti::disconnect()

These commits were cherry-picked from #1052


salkinium (approved):
Yes, thanks!

salkinium:
Unrelated, but I need to split up the STM32G0 compile tests, they run 92mins, which is way too long.

PR #1071, opened 2023-09-04
Improve file names processing in clang-format.sh

mbait:
- Process .cc and .hh files.
- Ignore empty file list.


salkinium (approved):
Thanks!

PR #1082, opened 2023-09-19
[docs] Update macOS installation of toolchain

salkinium:
Always be breaking things, we lovin' it.
I mean, how hard can backward compatibility of a package manager API really be?

cc @rleh

### DIFF
--- a/docs/src/guide/installation.md
+++ b/docs/src/guide/installation.md
@@ -152,6 +152,15 @@ brew install python3 scons git doxygen
 pip3 install modm
 ```
 
+!!! warning "Missing pyelftools"
+    If you get errors about missing `pyelftools` when calling `scons`, you may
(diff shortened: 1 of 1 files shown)


docs/src/guide/installation.md
```diff
@@ -172,8 +172,8 @@ Install the [pre-built ARM toolchain](https://github.com/osx-cross/homebrew-arm)
 
 ```sh
 brew tap osx-cross/arm
-brew install arm-gcc-bin@12
-brew install openocd --HEAD
+brew install --force arm-gcc-bin@12
```
rleh: Does that really fix the problem?

salkinium: No, it should be install then link --force like for AVR.

salkinium: fixed

PR #1090, opened 2023-10-09
[driver] SSD1306 SPI display

rleh:
- Rename `modm::Ssd1306` to `Ssd1306I2c`
- Add `modm::Ssd1306Spi`
  -  Initialize sequence for 128x32px, 0.91inch, blue, *HP12832-02* display module
- Convert all `enum`s in `struct modm::ssd1306` to `enum class`
- Fix mistake in SH1106 driver
- Add example for (existing) SH1106 driver


src/modm/driver/display/ssd1306_i2c.lb
```diff
@@ -12,19 +12,20 @@
 
 
 def init(module):
-    module.name = ":driver:ssd1306"
-    module.description = "SSD1306 Display"
+    module.name = ":driver:ssd1306.i2c"
```
rleh, line 15: Should we add an alias for the old module name?
If so, how does it work?

rleh:
All drivers (`ssd1306.i2c`, `ssd1306.spi` and `sh1106.i2c`) are now tested successfully on Hardware with a *HP12832-02* module for SPI and respectively a *GME12864-41* module for I²C.
image

src/modm/driver/display/sh1106_i2c.hpp
```diff
@@ -39,13 +39,10 @@ class Sh1106 : public Ssd1306<I2cMaster, Height>
 
 		this->transaction_success = true;
 
-		this->commandBuffer[0] = ssd1306::AdressingCommands::HigherColumnStartAddress;
-		this->commandBuffer[1] = 0x02;
-
 		for (page = 0; page < Height / 8; page++)
 		{
-			this->commandBuffer[2] = 0xB0 | page;
-			this->transaction_success &= RF_CALL(this->writeCommands(3));
```
rleh, line 48: The command sequence {`0x10`, `0x02`, `0xB0 | page`} gets interpreted (at least on my SSD1306 _GME12864_ module) as three command:
1.`0x10`:  _Set Higher Column Start Address for Page Addressing Mode (10h~1Fh)_ **to value 0**
2.`0x02` _Set Lower Column Start Address for Page Addressing Mode (00h~0Fh)_ **to value 2**
3. `0xB0` _Set Page Start Address for Page Addressing Mode (B0h~B7h)_ **to value `page`**

The first two commands result in the first two columns of display pixels displaying garbage and the last two columns of the frame not getting displayed, so I removed them.
SSD1106 should behave the same way like SH1106 in _Page Addressing Mode_, but I don't have SH1106 hardware laying arround, so I can't test it. Maybe @TomSaw could check?

src/modm/driver/display/ssd1306_i2c.lb
salkinium: yes, with `module.add_alias` in the `:driver` module. [See example here](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/module.lb#L43-L59). Also add a DEPRECATED: 2024q4 so we can remove it and not forget it like in my own example \*cough\*😇.

examples/nucleo_g474re/sh1106_i2c/main.cpp
```diff
@@ -0,0 +1,70 @@
...
+	Board::initialize();
+
+	modm::delay(100ms);
+
+	MyI2cMaster::connect<Sda::Sda, Scl::Scl>();
+	// Use 20kHz I2C and internal pull-ups, works without external pull-ups
+	// if the jumper wires are short enought
+	MyI2cMaster::initialize<Board::SystemClock, 20_kHz>();
+	Sda::configure(Sda::InputType::PullUp);
+	Scl::configure(Sda::InputType::PullUp);
```
chris-durand: The I2C master connect function takes an argument configuring the pull-ups. It defaults to `PullUps::External`. It would probably be cleaner to use that instead of calling `configure()` on the Gpios.

```suggestion
	// Use 20kHz I2C and internal pull-ups, works without external pull-ups
	// if the jumper wires are short enought
	MyI2cMaster::connect<Sda::Sda, Scl::Scl>(MyI2cMaster::PullUps::Internal);
	MyI2cMaster::initialize<Board::SystemClock, 20_kHz>();
```

src/modm/driver/display/sh1106_i2c.hpp
```diff
@@ -62,14 +59,14 @@ class Sh1106 : public Ssd1306<I2cMaster, Height>
 	{
 		RF_BEGIN();
 		// Default on Power-up - can be omitted
-		this->commandBuffer[0] = ssd1306::AdressingCommands::MemoryMode;
-		this->commandBuffer[1] = ssd1306::MemoryMode::PAGE;
+		this->commandBuffer[0] = uint8_t(ssd1306::AdressingCommands::MemoryMode);
```
chris-durand: Since C++23 the new preferred way to convert an enum to its underlying integer type is [`std::to_underlying`](https://en.cppreference.com/w/cpp/utility/to_underlying). It avoids the potential mistake of getting the type wrong. Not so important here and it's more characters. Just for information, I wouldn't really object to `uint8_t`.

src/modm/driver/display/ssd1306_i2c.hpp
```diff
@@ -0,0 +1,142 @@
...
+	{
+	public:
+		Ssd1306_I2cWriteTransaction(uint8_t address);
+
+		bool
+		configureDisplayWrite(const uint8_t *buffer, std::size_t size);
+
+	protected:
+		virtual Writing
+		writing() override;
```
chris-durand: It's normally recommended to use exactly one of `virtual`, `override` or `final` on a virtual function: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override

src/modm/driver/display/ssd1306_i2c.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+		virtual void
+		detaching(modm::I2c::DetachCause cause) override;
+
+		inline bool
+		isWritable()
```
chris-durand:
```suggestion
		inline bool
		isWritable() const
```

Any member function that doesn't mutate object state should be declared as `const`.

src/modm/driver/display/ssd1306_i2c.hpp
```diff
@@ -0,0 +1,142 @@
...
+	update()
+	{
+		RF_CALL_BLOCKING(startWriteDisplay());
+	}
+
+	/// Use this method to synchronize writing to the displays buffer
+	/// to avoid tearing.
+	/// @return	`true` if the frame buffer is not being copied to the display
+	bool
+	isWritable()
```
chris-durand:
```suggestion
	isWritable() const
```

src/modm/driver/display/ssd1306_i2c_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+// ----------------------------------------------------------------------------
+template<class I2cMaster, uint8_t Height>
+modm::ResumableResult<void>
+modm::Ssd1306I2c<I2cMaster, Height>::startWriteDisplay()
+{
+	RF_BEGIN();
+
+	RF_WAIT_UNTIL(
+		this->transaction.configureDisplayWrite((uint8_t*)(&this->buffer), sizeof(this->buffer)) and
```
chris-durand: `buffer` is a `uint8_t[][]`. This could be written without the C-style cast as `&this->buffer[0][0]`.

chris-durand (commented):
Nice!

src/modm/driver/display/ssd1306_i2c.lb
rleh: I added the following code to `src/modm/driver/module.lb`:
```python
# DEPRECATED: 2024q4
module.add_alias(
    Alias(name="ssd1306",
          destination=":ssd1306.i2c",
          description="DEPRECATED: Please use :driver:ssd1306.i2c instead."))
```
lbuild recognizes the alias and does not complain if I use it in `project.xml`, but does not generate any files...

examples/nucleo_g474re/sh1106_i2c/main.cpp
rleh: Ohh yeah, I remember...
Thanks!

src/modm/driver/display/sh1106_i2c.hpp
rleh: I wasn't aware of this feature, thanks!
The disadvantage is unfortunately that this makes the code lines even longer...

src/modm/driver/display/ssd1306_i2c.hpp
rleh: The code was only moved/copy-pasted, but I'll fix it anyway.

src/modm/driver/display/ssd1306_i2c.lb
salkinium: needs to be `destination=":driver:ssd1306.i2c",` I'll check why it doesn't complain.

README.md
```diff
@@ -792,31 +792,34 @@ your specific needs.
...
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-sh1106-i2c">SH1106-I2C</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-siemens_s65">SIEMENS-S65</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-siemens_s75">SIEMENS-S75</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sk6812">SK6812</a></td>
 <td align="center"><a href="https://modm.io/reference/module/modm-driver-sk9822">SK9822</a></td>
-<td align="center"><a href="https://modm.io/reference/module/modm-driver-ssd1306">SSD1306</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ssd1306-common">SSD1306-COMMON</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ssd1306-i2c">SSD1306-I2C</a></td>
+<td align="center"><a href="https://modm.io/reference/module/modm-driver-ssd1306-spi">SSD1306</a></td>
```
salkinium: let's [filter this noise out here](https://github.com/modm-io/modm/blob/develop/tools/scripts/synchronize_docs.py#L78).

src/modm/driver/display/ssd1306_i2c.lb
```diff
@@ -12,19 +12,20 @@
...
         ":processing:timer",
         ":ui:display")
     return True
 
 def build(env):
     env.outbasepath = "modm/src/modm/driver/display"
-    env.copy("ssd1306.hpp")
-    env.copy("ssd1306_impl.hpp")
-    env.copy("ssd1306_register.hpp")
+    env.copy("ssd1306.hpp") # DEPRECATE 2023q4
```
salkinium:
```suggestion
    env.copy("ssd1306.hpp") # DEPRECATE 2024q4
```

src/modm/driver/display/ssd1306_i2c.hpp
```diff
@@ -0,0 +1,142 @@
...
+	virtual modm::ResumableResult<void>
+	initializeMemoryMode();
+
+	virtual modm::ResumableResult<void>
+	startWriteDisplay();
+
+	bool transaction_success;
+};
+
+template<class I2cMaster, uint8_t Height = 64>
```
salkinium, line 136:
```suggestion
// DEPRECATE 2024q4
[[deprecated("Use `modm::Ssd1306I2c` instead!")]]
template<class I2cMaster, uint8_t Height = 64>
```

src/modm/driver/display/ssd1306.hpp
```diff
@@ -11,181 +9,6 @@
  */
 // ----------------------------------------------------------------------------
 
-#ifndef MODM_SSD1306_HPP
-#define MODM_SSD1306_HPP
+#warning "DEPRECATE 2023q4: Use 'ssd1306_i2c.hpp' instead!"
```
salkinium:
```suggestion
// DEPRECATE 2024q4
#warning "Use 'ssd1306_i2c.hpp' instead!"
```

TomSaw:
As I understood, you could test with sh1106 @rleh ? I've got these around. Let me know if I can help with hardware tests.

src/modm/driver/display/ssd1306_i2c.hpp
rleh: Actually:
```suggestion
// DEPRECATE 2024q4
template<class I2cMaster, uint8_t Height = 64>
using Ssd1306 [[deprecated("Use `modm::Ssd1306I2c` instead!")]] = Ssd1306I2c<I2cMaster, Height>;
```

src/modm/driver/display/ssd1306_i2c.lb
rleh: Nope, still does not work 😞

rleh:
> As I understood, you could test with sh1106 @rleh ? I've got these around. Let me know if I can help with hardware tests.

@TomSaw I tested the `examples/nucleo_g474re/sh1106_i2c/` example only with an (genuine) SSD1306 display module, because I don't have any SH1106 modules.
It would be very cool if you could test the example with your SH1106 display.
I assume you know how to get the example running on a dev board other than the Nucleo-G474RE, otherwise let me know if you need help.

TomSaw:
> > As I understood, you could test with sh1106 @rleh ? I've got these around. Let me know if I can help with hardware tests.
> 
> @TomSaw I tested the `examples/nucleo_g474re/sh1106_i2c/` example only with an (genuine) SSD1306 display module, because I don't have any SH1106 modules.
> It would be very cool if you could test the example with your SH1106 display.
> I assume you know how to get the example running on a dev board other than the Nucleo-G474RE, otherwise let me know if you need help.

ok, but gimme 1,2 days pls.

src/modm/driver/display/ssd1306_i2c.lb
salkinium: Oh ffs, I'll check it over the weekend. I need to rewrite lbuild, it's so unpythonic…

salkinium: It does build the alias destination for me though.

The reason it doesn't give a warning is that the `:driver:ssd1306.i2c` module is a dependency of the `:driver:sh1106.i2c` module, and is thus already included by the time the alias is resolved. If you change the dependency back to the alias:

```
 $ lbuild build
[WARNING] lbuild.node: Node 'modm:driver:ssd1306' has been moved to ':driver:ssd1306.i2c'!

>> modm:driver:ssd1306  [Alias]

DEPRECATED: Please use `:driver:ssd1306.i2c` instead.

Destination: :driver:ssd1306.i2c

>> modm:driver:ssd1306.i2c  [Module]

SSD1306 Display in I2C mode
```

salkinium:
Is there an update for this?

src/modm/driver/display/sh1106_i2c.hpp
TomSaw: And the hardware test shows:
These lines have a purpose. Whitout, the displays pages (columns of pixels) get ugly shifted.
I remember having read about this necessary micro-management in an adafruit or arduino library comment, not in the datasheet or elsewhere.

TomSaw:
Tested SH1106 and you actually have to keep the ssd1306::AdressingCommands::HigherColumnStartAddress` assignments in the SH1106 driver if you don't have a preference for fuzzy diagonally shifted pages of pixels.

The memory mapping of sh1106 is different (and worse) than SSD1306

rleh:
Ok, thanks @TomSaw for testing!
I'll have a look (again) at the SH1106 behavior. I just ordered two SH1106 displays to be able to test it myself, they should arrive in 2-3 weeks...

TomSaw:
If there's more to test, i'll respond much faster now cause my "embedded suite" is re-assembled now @rleh .
Also, it could have been an option to send you an SH1106 instance from my stock, since we're sitting in the same country as i know.

rleh:
> Also, it could have been an option to send you an SH1106 instance from my stock, since we're sitting in the same country as i know.

Thanks for the offer. I'm afraid I have enough (or even too much) to do until Easter, so I won't take care of it until after anyway.

salkinium:
Can I merge this? What's the status of this?

rleh:
I have to test/debug and maybe fix the problem with the SH1106.

PR #1096, opened 2023-11-03
Add isCountingDown/Up functions to STM32 Timers

ser-plu:
Adding functions to read the current counting direction of the timers


chris-durand (approved):
Thanks!

PR #1097, opened 2023-11-05
Fix example main.cpp in custom-project documentation.

jensboe:
The example main.cpp in the custom-project documentation use an old(?) delay function call.


rleh (approved):
Thanks!

PR #1100, opened 2023-11-16
[stm32] Place stack in SRAM by default on H7, add lbuild option

chris-durand:
The DTCM is not accessible by regular DMA on H7 devices, which leads to DMA transfer errors when DMA buffers are allocated on the stack. The default stack location is changed to the D1 SRAM. An lbuild option is added to select the stack location between SRAM and DTCM.

### DIFF
--- a/src/modm/platform/core/stm32/module.lb
+++ b/src/modm/platform/core/stm32/module.lb
@@ -30,6 +30,15 @@ def prepare(module, options):
                 default="rom")
             )
 
+    if options[":target"].identifier.family == "h7":
+        module.add_option(
+            EnumerationOption(
+                name="main_stack_location",
+                description="SRAM (default) or DTCM (faster, but not DMA-capable)",
+                enumeration=["sram", "dtcm"],
+                default="sram")
+            )
+
     module.depends(":platform:cortex-m")
     return True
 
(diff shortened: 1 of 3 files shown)


chris-durand:
@salkinium Do you have an idea how to fix the MacOS CI? There is some Python problem:
```
    import jinja2		# template engine
    ^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'jinja2'
scons: *** [generated/xpcc/identifier.hpp] Error 1
```

jinja2 is not found, but only when building the XPCC examples.

src/modm/platform/core/stm32/module.lb
```diff
@@ -30,6 +30,15 @@ def prepare(module, options):
                 default="rom")
             )
 
+    if options[":target"].identifier.family == "h7":
+        module.add_option(
+            EnumerationOption(
+                name="stack_location",
```
salkinium:
```suggestion
                name="main_stack_location",
```

To align it with the `main_stack_size` naming.

salkinium:
The macOS Image Release was upgraded from [20231018.14](https://github.com/actions/runner-images/releases/tag/macos-13%2F20231018.14) -> [20231025.2](https://github.com/actions/runner-images/releases/tag/macos-13%2F20231025.2). In the process the pre-installed Python version is 3.12.0, while we install 3.11.6.
The issue is that the XPCC SCons integration calls the builds via a `python3` command line, which points to 3.12.0, while jinja2 is only installed for 3.11.6.

The solution is to fix the XPCC SCons integration to use the same Python interpreter that it was called with and not delegate to the shell environment for it. 🙄

salkinium:
Gentle boop @chris-durand

salkinium (approved):
Thanks!

PR #1109, opened 2023-12-22
[ext] Update submodules

salkinium:
In preparation of 2023q4 release, lets update some submodules and see what breaks!

Tested on various hardware:

- [x] ETL v20.38.2 -> v20.38.10.
- ~TinyUSB v0.15.0 -> v0.16.0.~
- [x] LVGL v8.3.10 -> v8.3.11.
- [x] modm-devices
- [x] Nanopb v0.4.7 -> v0.4.8.
- [x] STM32 headers
  - [x] STM32G0 v1.4.2 -> v1.4.3.
  - [x] STM32F4 v2.6.8 -> v2.6.9.
  - [x] STM32U5 v1.3.0 -> v1.3.1.

PR #1112, opened 2023-12-28
2023q4 release

salkinium:
- [x] Update changelog
- [x] Update submodules (depends on #1109)
- [x] Hardware testing


salkinium:
There's not much more to be done, I tested a bunch of examples that the submodules were updated for, which worked fine. I'll do the release and then continue debugging why TinyUSB v0.16.0 doesn't work.

PR #1114, opened 2024-01-02
Miscellaneous fixes and improvements

chris-durand:
- [x] Fix `Rcc::enable` for STM32H7 comparator
- [x] Add STM32 advanced timer `isOutputEnabled()`
- [x] Suppress GCC ABI nuisance warning ("note: parameter passing for argument of type
    '...' when C++17 is enabled changed to match C++14 in GCC 10.1")
- [x] Fix RTT logger if data structure is not in RAM with lowest address
- [x] Allow unit test case names in snake_case (`test_foo()`)
- [x] Treat missing return in function as error (this is always undefined behaviour)
- [x] Fix compilation of STM32G0 DMA. The CI didn't catch it because the template function is never instantiated in the test suite.
- [x] Fix reconfiguring DMAMUX requests
- [x] Fix STM32H7 32-bit timer counter size
- [x] Fix STM32H7 ADC asynchronous clock
- [x] Fix sampling time configuration in STM32F3 ADC driver
- [x] Fix compilation of STM32 DMA channel `clearInterruptFlags()`

### DIFF
--- a/tools/build_script_generator/common.py
+++ b/tools/build_script_generator/common.py
@@ -213,6 +213,7 @@ def common_compiler_flags(compiler, target):
         "-Werror=maybe-uninitialized",
         "-Werror=overflow",
         "-Werror=sign-compare",
+        "-Werror=return-type",
         "-Wextra",
         "-Wlogical-op",
         "-Wpointer-arith",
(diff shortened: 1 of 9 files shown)


tools/build_script_generator/common.py
```diff
@@ -213,6 +213,7 @@ def common_compiler_flags(compiler, target):
         "-Werror=maybe-uninitialized",
         "-Werror=overflow",
         "-Werror=sign-compare",
+        "-Werror=return-type",
```
salkinium, line 216: Wat, how is this not a C++ default?!?!

salkinium (approved):
Thanks!

PR #1117, opened 2024-01-04
Add board support for STM32F401 Discovery

cajt:
Added a BSP for the F401 Discovery Board.

Notes:
- Completed basic testing with the included examples, including the USART, Gyro and Accelerometer.
- Board is obsolete according to ST.

### DIFF
--- a/examples/stm32f401_discovery/uart/main.cpp
+++ b/examples/stm32f401_discovery/uart/main.cpp
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2011, Georgi Grinshpun
+ * Copyright (c) 2011-2012, Fabian Greif
+ * Copyright (c) 2012, 2014, Sascha Schade
+ * Copyright (c) 2013, Kevin Läufer
+ * Copyright (c) 2013, 2015-2017, Niklas Hauser
+ * Copyright (c) 2024, Carl Treudler
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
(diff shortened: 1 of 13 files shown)


rleh:
The CI complains about two things:
- Please add the new example folder to the appropriate CI job in `.github/workflows/linux.yml`.
- Please run `python3 tools/scripts/synchronize_docs.py` to synchronize the documentation.

examples/stm32f401_discovery/uart/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#include <modm/board.hpp>
+
+// ----------------------------------------------------------------------------
+/**
+ * Very basic example of USART usage.
+ * The ASCII sequence 'A', 'B', 'C, ... , 'Z', 'A', 'B', 'C', ...
```
rleh:
```suggestion
 * The ASCII sequence 'A', 'B', 'C', ... , 'Z', 'A', 'B', 'C', ...
```

rleh (approved):
We really need to take care of deduplicating the examples somehow, but that shouldn't stop this pull request.

examples/stm32f401_discovery/uart/main.cpp
cajt: done.

cajt:
> We really need to take care of deduplicating the examples somehow, but that shouldn't stop this pull request.

From the abstraction side, the boards looked almost identical and could run mostly unmodified code (only difference was number of LEDs and their names). 

The way I used the examples was basically a unit test, to check is the BSP is working. For use as code examples the level of abstraction is fine imho, and kinda what I would expect. To ease dealing with testing, a more abstract and generalized way of writing these might help. Especially for more elaborate tests. But that things can't do double duty as example and unit-test anymore (if that was ever intended).

salkinium (approved):
Thanks!

(I'll squash your commits and merge it later)

PR #1122, opened 2024-01-21
Refactor IWDG driver and add initialize function for prescaler with downcounter

salkinium:
Ok, this escalated a little. I was just looking through the code and saw that the Iwdg class was not inlined, so I quickly refactored it and then noticed that the prescaler was manually computed and then I had to program a constexpr algorithm for it.

- [x] Inline IWDG functions
- [x] Move into `modm::platform` namespace.
- [x] Add prescaler + downcounter algorithm.
- [x] Add unittest for prescaler + downcounter algorithm.
- [x] Add `Rcc::Lsi` and `Rcc::Hsi` clock.
- [x] Add `modm.digsep` filter for nicely formatting long integers.
- [x] Add `SystemClock::Iwdg` to every board.
- [x] Add a generic `assertBaudrateInTolerance` implementation.
- [x] Add documentation for prescaler calculators.
- [x] Simplify tolerance assertion functions.
- [x] Added `seconds_t`, `milliseconds_t`, `microseconds_t` for template parameters.
- [x] `percent_t` can be float since C++20.
- [x] Fix minimum prescaler value for STM32 UART (8 or 16).
- [x] Implement iterator correctly so that the warnings go away. Help!


examples/nucleo_f072rb/independend_watchdog/main.cpp
```diff
@@ -25,7 +26,7 @@ main()
 	Board::initialize();
 	LedD13::setOutput();
 	// set the watchdog timeout to 4 seconds
-	Iwdg::initialize(Iwdg::Prescaler::Div32, 0x0FFFu);
+	Iwdg::initialize<SystemClock, 4.0>();
```
salkinium: Sadly `std::chrono::microseconds` are not allowed as a template type.

src/modm/platform/iwdg/stm32/iwdg.hpp
```diff
@@ -29,28 +37,66 @@ class Iwdg
...
-	trigger();
-	static Status
-	getStatus();
+	initialize()
+	{
+		constexpr double frequency = 1.0 / timeout_seconds;
+		constexpr auto result = modm::GenericPrescalerCounter<double>::from_power(
+			SystemClock::Iwdg, frequency, 1ul << 12, 4, 256);
+		assertBaudrateInTolerance< baudrate_t(result.frequency * 1e6),
+								   baudrate_t(frequency * 1e6), tolerance >();
```
salkinium: Still need to add a generic implementation of this.

examples/nucleo_f072rb/independend_watchdog/main.cpp
chris-durand: If you really wanted to, you could define a custom microsecond struct with a non-explicit converting constructor that takes a `std::chrono::microseconds` parameter. 
That could be used as the template argument type and `std::chrono::microseconds` would implicitly convert to it. A bit ugly, but it does the job.

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,190 @@
...
+		constexpr Result(T input_frequency, T desired_frequency,
+						 uint32_t index, uint32_t prescaler, uint32_t counter) :
+			index{index}, prescaler{prescaler}, counter{counter},
+			frequency{input_frequency / (prescaler * counter)},
+			input_frequency{input_frequency},
+			desired_frequency{desired_frequency},
+			error{1.f - float(input_frequency) / float(prescaler * counter * desired_frequency)}
+		{}
+		/// Zero-indexed prescaler result
+		const uint32_t index;
```
chris-durand, line 44: `const` data members are generally not recommended because it makes types non-copyable and non-moveable, so it's not possible to pass the result to a function. Do we really want to enforce that?

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,190 @@
...
+		/// Relative Frequency Error
+		const float error;
+	};
+
+public:
+	/// From any iterable container.
+	/// @note container must have `begin()`, `end()` and `size()` function!
+	template<class Iterator>
+	static constexpr Result
+	from_iterator(T input_frequency, T desired_frequency, uint32_t max_counter, Iterator prescalers)
```
chris-durand, line 64: The `from_iterator` function does not actually take any iterator arguments, it is rather something like a container or a range. I would change the input to a pair of iterators or rename it and have it take a [`std::ranges::forward_range`](https://en.cppreference.com/w/cpp/ranges/forward_range) by forwarding reference. That would be the most generic version that could work with containers, views, C-arrays, etc. You would have to replace calls to `.begin()` / `.end()` by the non-member version to support plain C-arrays.

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,190 @@
...
+			Function modifier;
+			bool reversed;
+
+			constexpr value_type operator*() const { return modifier(value); }
+			constexpr linear_range_iterator& operator++()
+			{ state++; reversed ? value-- : value++; return *this; }
+			constexpr linear_range_iterator& operator=(const linear_range_iterator&o)
+			{ state = o.state; value = o.value; reversed = o.reversed; modifier = o.modifier; return *this; }
+			constexpr bool operator==(const linear_range_iterator &o) const { return state == o.state; }
+			constexpr bool operator!=(const linear_range_iterator &o) const { return state != o.state; }
```
chris-durand, line 141: There is no need to define `operator !=` anymore in C++20. The compiler will automatically synthesize it from `operator==`.
```suggestion
```

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,190 @@
...
+			difference_type operator-(const linear_range_iterator &o) { return state - o.state; }
```
chris-durand, line 142: This is not `constexpr`

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,190 @@
...
+			value_type state;
+			value_type value;
+			Function modifier;
+			bool reversed;
+
+			constexpr value_type operator*() const { return modifier(value); }
+			constexpr linear_range_iterator& operator++()
+			{ state++; reversed ? value-- : value++; return *this; }
+			constexpr linear_range_iterator& operator=(const linear_range_iterator&o)
+			{ state = o.state; value = o.value; reversed = o.reversed; modifier = o.modifier; return *this; }
```
chris-durand, line 139:
```suggestion
```
The compiler generated version does the same thing, no need to define the assignment operator manually. This is probably also why you get the `-Wdeprecated-copy` because there is no user-defined copy-constructor to match the assignment operator.

chris-durand (commented):
Very nice, thanks!

salkinium:
Great, I messed up and merged this accidentally while rebasing. Fantastic…

salkinium:
There's no way to open a merged PR, unlike closed PRs, so I'll address the review in a separate PR on the weekend. Damn it.

src/modm/math/algorithm/prescaler_counter.hpp
salkinium: The problem is that this then requires a default constructor and I cannot default construct the `Function`, cos it's a lambda. This iterator design does not work anyways for closures, but I cannot use `modm::inplace_function` cos not constexpr.

chris-durand: Hm, when I remove the lines as proposed and the `#pragma` ones the code just works fine for me and compiles warning-free. This shouldn't do anything with default constructors.

chris-durand: To make it work with closures you can store a pointer to the function object in the iterator instead of the function itself. There is no lifetime issue because we only return the `Result` and `linear_range` remains valid during the call to `from_iterator`.

src/modm/math/algorithm/prescaler_counter.hpp
salkinium: Yup, this worked great, thanks for the tip! Ranges are actually kinda cool!

PR #1123, opened 2024-01-22
[ci] Update CI for faster default runners

salkinium:
[GitHub doubled the default runner resources](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/), let's check if we can do more faster.

- [x] Instrument the Workflows
- [ ] Optimized number of jobs

### DIFF
--- a/.github/workflows/compile-all.yml
+++ b/.github/workflows/compile-all.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     types: [labeled]
 
+env:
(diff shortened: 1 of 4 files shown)


salkinium:
There doesn't seem to be a difference when using `SCONSFLAGS="-j4"` (the current settings) vs. `SCONSFLAGS="-j8"`. However, the hosted unittests are compiled with only one core.

salkinium:
Meh, not worth it.

PR #1127, opened 2024-02-03
Refactor IWDG driver and add initialize function for prescaler with downcounter

salkinium:
Re-opening this since I prematurely merged the previous PR #1122 due to pushing the wrong branch (PEBCAK). Addresses the last review. Original description:

Ok, this escalated a little. I was just looking through the code and saw that the Iwdg class was not inlined, so I quickly refactored it and then noticed that the prescaler was manually computed and then I had to program a constexpr algorithm for it.

- [x] Inline IWDG functions
- [x] Move into `modm::platform` namespace.
- [x] Add prescaler + downcounter algorithm.
- [x] Add unittest for prescaler + downcounter algorithm.
- [x] Add `Rcc::Lsi` and `Rcc::Hsi` clock.
- [x] Add `modm.digsep` filter for nicely formatting long integers.
- [x] Add `SystemClock::Iwdg` to every board.
- [x] Add a generic `assertBaudrateInTolerance` implementation.
- [x] Add documentation for prescaler calculators.
- [x] Simplify tolerance assertion functions.
- [x] Added `seconds_t`, `milliseconds_t`, `microseconds_t` for template parameters.
- [x] `percent_t` can be float since C++20.
- [x] Fix minimum prescaler value for STM32 UART (8 or 16).
- [x] Rename `from_range` to `from_linear`.
- [x] Rename `from_iterator` to `from_range` and implement it to be compatible with `std::ranges::forward_range`.


src/modm/board/nucleo_g071rb/board.hpp
```diff
@@ -70,7 +70,7 @@ struct SystemClock
 	static constexpr uint32_t Usart3	= Apb;
 	static constexpr uint32_t Usart2	= Apb;
 	static constexpr uint32_t Spi2		= Apb;
-	static constexpr uint32_t Iwdg		= Apb;
+	static constexpr uint32_t Iwdg		= Rcc::LsiFrequency;
```
chris-durand, line 73: You wanted to make the same change to the G070 Nucleo as well. Alternatively you could merge the headers.

src/modm/math/algorithm/prescaler.hpp
```diff
@@ -29,50 +27,47 @@ namespace modm
...
 		/// Relative Frequency Error
-		const float error;
+		float error;
 	};
 
 public:
-	/// From any iterable container.
-	/// @note container must have `begin()`, `end()` and `size()` function!
-	template<class Iterator>
+	/// From any forward range.
```
chris-durand, line 60: The function also requires the forward range to be sized because it calls `std::ranges::size()`. This wouldn't work in cases where computing the size is O(n) like for a `std::forward_list` or with certain range adaptors. `std::ranges::distance()` would be the more generic function to compute the size even if it's O(n). Just a little nitpick, the function is probably fine as is.

Furthermore this function has undefined behavior if an empty range is passed in, but this case doesn't make sense. Not sure if we can do anything about it except documenting it. When used in a compile-time context it simply wouldn't compile.

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+public:
+	/// From any forward range.
+	static constexpr Result
+	from_range(T input_frequency, T desired_frequency, uint32_t max_counter,
+			   std::ranges::forward_range auto&& prescalers)
+	{
+		const double desired = double(input_frequency) / desired_frequency;
+		const auto [min, max] = std::minmax_element(
+				std::ranges::begin(prescalers), std::ranges::end(prescalers));
```
chris-durand:
```suggestion
		const auto [min, max] = std::ranges::minmax_element(prescalers));
```

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,186 @@
...
+		{
+			const double err = std::abs(desired - pre * cnt);
+			if (err < out_error) {
+				out_error = err;
+				out_prescaler = pre;
+				out_counter = cnt;
+				out_index = idx;
+			}
+		};
+		for (uint32_t idx{0}; const auto prescaler : prescalers)
```
chris-durand: Maybe
```suggestion
		for (const auto [idx, prescaler] : enumerate(prescalers))
```
?

src/modm/math/algorithm/prescaler_counter.hpp
```diff
@@ -0,0 +1,186 @@
...
+			value_type state;
+			value_type value;
+			Function *modifier;
+			bool reversed;
+
+			constexpr value_type operator*() const { return (*modifier)(value); }
+			constexpr linear_range_iterator& operator++() { state++; reversed ? value-- : value++; return *this; }
+			constexpr linear_range_iterator operator++(int) { auto old(*this); ++*this; return old; }
+			constexpr bool operator==(const linear_range_iterator &o) const { return state == o.state; }
+			constexpr difference_type operator-(const linear_range_iterator &o) { return state - o.state; }
```
chris-durand:
```suggestion
			constexpr difference_type operator-(const linear_range_iterator &o) const { return state - o.state; }
```

src/modm/math/algorithm/prescaler.hpp
```diff
@@ -91,54 +86,62 @@ GenericPrescaler
...
-			const Function modifier;
-			constexpr uint32_t operator*() const { return modifier(state); }
+			Function *modifier;
+
+			constexpr value_type operator*() const { return (*modifier)(state); }
 			constexpr linear_range_iterator& operator++() { state++; return *this; }
-			constexpr bool operator!=(const linear_range_iterator &o) const { return state != o.state; }
+			constexpr linear_range_iterator operator++(int) { auto old(*this); ++*this; return old; }
+			constexpr bool operator==(const linear_range_iterator &o) const { return state == o.state; }
+			constexpr difference_type operator-(const linear_range_iterator &o) { return state - o.state; }
```
chris-durand:
```suggestion
			constexpr difference_type operator-(const linear_range_iterator &o) const { return state - o.state; }
```

chris-durand (approved):
Thanks!

src/modm/math/algorithm/prescaler_counter.hpp
salkinium: oh, [`std::ranges::enumerate_view`](https://en.cppreference.com/w/cpp/ranges/enumerate_view) exists, so we can alias this to `modm::enumerate`.

chris-durand: :+1:, but that requires gcc 13. Mac OS hosted and AVR currently have 12. Let's upgrade the compilers for the next release and do it afterwards.

salkinium: We need to find a workaround for the Protothread macro issue to upgrade to GCC13…

chris-durand: I don't think this is fixable. The whole construct relies on jumping into a statement expression with a case label. According to the gcc documentation this was never valid but is now properly diagnosed since gcc 13. It just happened to work often enough ...

The only acceptable solution I see is abandoning resumable functions.

salkinium: This doesn't seem to work due to an (for me) incomprehensible compiler error about `unsigned int&` lvalue vs `unsigned int` rvalue… help (see CI for issue)

chris-durand: It's a limitation in `modm::enumerate`. The iterator types you have defined (`linear_range`, etc.) return r-values, here plain integers, not l-value references when you dereference them and the implementation can't deal with that. I'll push a fix tonight.

chris-durand: It's fixed now. I've also pushed a fixup for a typo in the Nucleo G070 `module.lb` file.

salkinium: Thank you!

PR #1128, opened 2024-02-23
Fix float casting 

victorandrehc:
Correctly casts to float before calling std::round in delay function


salkinium:
Did you close this intentionally?

PR #1134, opened 2024-02-26
[stm32] Extend Timer features: PWM modes, Break Input and Comparator

victorandrehc:
1. Add PWM Modes
2. Add output configure channels on timer
3. Fixes the register field size for the repetition counter register
4. Break Input on Timers
5. Implementation for comparator in the g0

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ Please [discover modm's peripheral drivers for your specific device][discover].
 <td align="center">✅</td>
 <td align="center">✕</td>
 <td align="center">✕</td>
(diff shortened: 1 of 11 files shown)


salkinium (commented):
Yes from me, but I can only judge syntax, since I don't know much about the Timer hardware.

chris-durand (approved):
Thanks

rleh:
~~@salkinium What happened to this commits? Both commits somehow disappeared from develop...?~~
Nevermind, I was on the wrong repo 🤦🏽

PR #1135, opened 2024-02-26
Add disco_f411ve

TomSaw:
Quite an old board i found behind the fridge. Very similar to disco_f407vg despite different clocks and PLL factors.

It's the only board in the modm collection featuring a **stm32f411vet6**, hence the clock and PLL constants maybe worth being stored.

### DIFF
--- a/src/modm/board/disco_f411ve/board.hpp
+++ b/src/modm/board/disco_f411ve/board.hpp
@@ -1,0 +1,231 @@
+/*
+ * Copyright (c) 2024, Thomas Sommer
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+#pragma once
+
+#include <modm/architecture/interface/clock.hpp>
+#include <modm/driver/inertial/lis3dsh.hpp>
+#include <modm/platform.hpp>
+
+using namespace modm::platform;
+
+namespace Board
+{
+/// @ingroup modm_board_disco_f411ve
+/// @{
+using namespace modm::literals;
+
+/// STM32F411 running at 96MHz generated from the external 8MHz crystal
+struct SystemClock
+{
+	static constexpr uint32_t Frequency = 96_MHz;
+	static constexpr uint32_t Ahb = Frequency;
+	static constexpr uint32_t Apb1 = Frequency / 2;
+	static constexpr uint32_t Apb2 = Frequency;
+
+	static constexpr uint32_t Adc = Apb2;
+
+	static constexpr uint32_t Spi1 = Apb2;
+	static constexpr uint32_t Spi2 = Apb1;
+	static constexpr uint32_t Spi3 = Apb1;
+	static constexpr uint32_t Spi4 = Apb2;
+	static constexpr uint32_t Spi5 = Apb2;
+
+	static constexpr uint32_t Usart1 = Apb2;
+	static constexpr uint32_t Usart2 = Apb1;
+	static constexpr uint32_t Usart3 = Apb1;
+	static constexpr uint32_t Uart4 = Apb1;
+	static constexpr uint32_t Uart5 = Apb1;
+	static constexpr uint32_t Usart6 = Apb2;
+	static constexpr uint32_t Uart7 = Apb1;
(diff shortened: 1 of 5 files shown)


src/modm/board/disco_f411ve/board.hpp
```diff
@@ -0,0 +1,230 @@
...
+	static constexpr uint32_t Timer1 = Apb2Timer;
+	static constexpr uint32_t Timer2 = Apb1Timer;
+	static constexpr uint32_t Timer3 = Apb1Timer;
+	static constexpr uint32_t Timer4 = Apb1Timer;
+	static constexpr uint32_t Timer5 = Apb1Timer;
+	static constexpr uint32_t Timer9 = Apb2Timer;
+	static constexpr uint32_t Timer10 = Apb2Timer;
+	static constexpr uint32_t Timer11 = Apb2Timer;
+
+	static constexpr uint32_t Usb = 48_MHz;
```
salkinium, line 65: Is this correct?

salkinium (commented):
Thanks! Do you want to try the USB example on this one? Should be almost identical to [the F407 disco](https://github.com/modm-io/modm/blob/develop/examples/generic/usb/project.xml#L7).

src/modm/board/disco_f411ve/board.hpp
TomSaw: I've setup a PWM on Timer10 and Timer11, confirmed the output frequency and looks fine.

But could not make USB / TinyUSB working.
Please give me some advice!

Observations:
- Power on without Usb-Cable, Leds blink period = 2.5s -> Seems to be the initial default
- Connecting an Usb-Cable, the Leds blink period remains at 2.5s -> `tud_mount_cb()` not invoked.
- However, when disconnecting the Usb-Cable, the blink period changes to 250ms -> `tud_umount_cb()` invoked!
- Reconnecting the cable again, changes nothing. The blink period remains at 250ms.
- Only a reset gets me back to 2.5s right now.

salkinium: That's cos the USB is running at 50MHz and not 48MHz, hence my question about whether this clock is correct ;-P It probably works for a short time, but then gets out of sync.

CubeMX shows this solution running the CPU at 96MHz and USB at 48MHz:

salkinium: Actually, the /Q prescaler isn't set at all, so the USB probably runs at 100MHz or something completely wrong.

TomSaw: Ok, i have not really tried to iunderstand the clock system yet. So i got a copy of CubeMX as it looks like an usefull reference. Have seen this graphical configurator before but your screenshot finally convinced me to give it a closer look, thanks!

USB now works 😄

src/modm/board/disco_f411ve/board.hpp
```diff
@@ -0,0 +1,231 @@
...
```
TomSaw, line 65: The PLL statements in all board.hpp could be moved above the `SystemClock` struct to calculate 
`SystemClock::Frequency` and `SystemClock::Usb`

👉🏾 This would document the causality and the "...Usb requires 48Mhz..." assertion somewhere else in modm would have catched me!

Something like this:
```
static constexpr uint32_t CrystalHSE = 8;
const Rcc::PllFactors pllFactors {
   ...
}

static constexpr uint32_t MainPllClock = CrystalHSE / PllFactors.pllM * PllFactors.pllN;

static constexpr uint32_t SystemClock::Frequency = MainPllClock / PllFactors.pllP;
static constexpr uint32_t SystemClock::Usb = MainPllClock / PllFactors.pllQ;
```

salkinium: Ohhhhhhh, that's an **excellent** idea!!! I'll merge this first, then prototype it. We can probably also put the AHB/APB prescalers in there.

TomSaw: Something easy, to build up 'momentum of success' will also support my motivational chemistry to complete the old graphic related PRs.

TomSaw: I've quickly sketched this in #1137

src/modm/board/disco_f411ve/board.hpp
salkinium: I swapped the prescalers to align with the CubeMX solution, so the CPU now runs at 96MHz and the USB still at 48MHz.

salkinium (approved):

PR #1138, opened 2024-03-06
[driver] Adding AS5047 driver and example

hshose:
- [x] driver
- [x] example
- [x] tested on hardware

### DIFF
--- a/src/modm/driver/encoder/as5047.hpp
+++ b/src/modm/driver/encoder/as5047.hpp
@@ -1,0 +1,157 @@
+// coding: utf-8
+// ----------------------------------------------------------------------------
+/*
+ * Copyright (c) 2024, Henrik Hose
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef MODM_AS5047_HPP
+#define MODM_AS5047_HPP
+
+#include <array>
+#include <modm/architecture/interface/register.hpp>
+#include <modm/architecture/interface/spi_device.hpp>
+#include <modm/processing/resumable.hpp>
+#include <modm/processing/timer.hpp>
+#include <numbers>
+
+namespace modm
+{
+
+/// @cond
+namespace detail
+{
+constexpr uint16_t
+as5047_setMsbToEvenParity(const uint16_t num)
+{
+	uint16_t par = 0x7fff & num;
+	par ^= par >> 8;
+	par ^= par >> 4;
+	par ^= par >> 2;
+	par ^= par >> 1;
+	return ((par & 1) << 15) | (0x7fff & num);
+}
+}  // namespace detail
+/// @endcond
+
+/// @ingroup modm_driver_as5047
+struct as5047
+{
+	enum class Errorfl : uint16_t
+	{
+		Parerr = Bit2,
+		Invcomm = Bit1,
+		Frerr = Bit0,
+	};
+	MODM_FLAGS16(Errorfl)
+
+	enum class Prog : uint16_t
+	{
+		Progver = Bit6,
+		Progotp = Bit3,
+		Otpref = Bit2,
+		Progen = Bit0,
+	};
+	MODM_FLAGS16(Prog)
+
+	enum class Diaagc : uint16_t
+	{
+		Magl = Bit11,
+		Magh = Bit10,
+		Cof = Bit9,
+		Lf = Bit8,
+	};
+	MODM_FLAGS16(Diaagc)
+
+	enum class Register : uint16_t
+	{
+		ReadNop = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x0000)),
+		ReadErrfl = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x0001)),
+		ReadProg = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x0003)),
+		ReadDiaagc = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x3FFC)),
+		ReadMag = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x3FFD)),
+		ReadAngleunc = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x3FFE)),
+		ReadAnglecom = detail::as5047_setMsbToEvenParity(((1 << 14) | 0x3FFF)),
+
+	};
+
+	struct modm_packed Data
+	{
+		/// @return
+		constexpr float
+		getAngleRad() const
+		{
+			const uint16_t angle = data & 0x3fff;
+			return static_cast<float>(angle) / 16383.f * 2.f * std::numbers::pi_v<float>;
+		}
+
+		/// @return
+		constexpr float
+		getAngleDeg() const
+		{
+			const uint16_t angle = data & 0x3fff;
+			return static_cast<float>(angle) / 16383.f * 360.f;
+		}
+
+		/// @return
+		constexpr uint16_t
+		getAngleRaw() const
+		{
+			const uint16_t angle = data & 0x3fff;
+			return angle;
+		}
+
+		uint16_t data;
+	};
+};  // struct as5047
+
+/**
+ * @tparam SpiMaster
+ * @tparam Cs
+ *
+ * @author  Henrik Hose
+ * @ingroup modm_driver_as5047
+ */
+template<typename SpiMaster, typename Cs>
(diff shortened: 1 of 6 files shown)


src/modm/driver/encoder/as5047.hpp
```diff
@@ -0,0 +1,152 @@
...
+constexpr uint16_t
+setMsbToEvenParity(const uint16_t num)
+{
+	uint16_t par = 0x7fff & num;
+	par ^= par >> 8;
+	par ^= par >> 4;
+	par ^= par >> 2;
+	par ^= par >> 1;
+	return ((par & 1) << 15) | (0x7fff & num);
+}
```
salkinium, line 39: This is the modm namespace which is quite generic. Would be good to move to `struct as5047` or put into the relevant `:math` module.

src/modm/driver/encoder/as5047.lb
```diff
@@ -0,0 +1,32 @@
...
+#- * - coding : utf - 8 - * -
+#
+#Copyright(c) 2024, Henrik Hose
+#
+#This file is part of the modm project.
+#
+#This Source Code Form is subject to the terms of the Mozilla Public
+#License, v.2.0. If a copy of the MPL was not distributed with this
+#file, You can obtain one at http:  // mozilla.org/MPL/2.0/.
+#-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -
```
salkinium: some weird whitespace in this header?

src/modm/driver/encoder/as5047.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+	enum class Register : uint16_t
+	{
+		ReadNop = setMsbToEvenParity(((1 << 14) | 0x0000)),
+		ReadErrfl = setMsbToEvenParity(((1 << 14) | 0x0001)),
+		ReadProg = setMsbToEvenParity(((1 << 14) | 0x0003)),
+		ReadDiaagc = setMsbToEvenParity(((1 << 14) | 0x3FFC)),
+		ReadMag = setMsbToEvenParity(((1 << 14) | 0x3FFD)),
+		ReadAngleunc = setMsbToEvenParity(((1 << 14) | 0x3FFE)),
+		ReadAnglecom = setMsbToEvenParity(((1 << 14) | 0x3FFF)),
```
salkinium: You could move the `(1 << 14)` into the function.

src/modm/driver/encoder/as5047_impl.hpp
```diff
@@ -0,0 +1,58 @@
...
+modm::ResumableResult<void>
+As5047<SpiMaster, Cs>::readout()
+{
+	RF_BEGIN();
+
+	RF_WAIT_UNTIL(this->acquireMaster());
+
+	Cs::reset();
+	inBuffer[0] = 0;
+	inBuffer[1] = 0;
```
salkinium: Don't think you need to zero the read buffer.

src/modm/driver/encoder/as5047_impl.hpp
```diff
@@ -0,0 +1,58 @@
...
+	outBuffer[1] = static_cast<uint8_t>(Register::ReadAngleunc);
+	outBuffer[0] = static_cast<uint8_t>(static_cast<uint16_t>(Register::ReadAngleunc) >> 8);
+	RF_CALL(SpiMaster::transfer(outBuffer, inBuffer, 2));
+	Cs::set();
+
+	modm::delay_us(1);
+
+	Cs::reset();
+	inBuffer[0] = 0;
+	inBuffer[1] = 0;
```
salkinium: same.

salkinium (changes requested):
Nice, just some nitpicks.

src/modm/driver/encoder/as5047.hpp
hshose: I don't think so, see [Figure 13 from the Datasheet](https://ams.com/documents/20143/36005/AS5047P_DS000324_3-00.pdf), bit 14 is the "READ/WRITE" bit, since these registers are only ever read, its unrelated to the parity bit.
I thought its more expressive this way, since it closely models what the datasheet says.

image

src/modm/driver/encoder/as5047.hpp
hshose: I'm not sure, how to move it into the struct?

<details>
<summary>
If I declare `setMsbToEvenParity` as `static constexpr` in the `as5047` struct,
I get an error saying `error: 'static constexpr uint16_t modm::as5047::setMsbToEvenParity(uint16_t)' used before its definition`
</summary>

```
struct as5047
{
	enum class Errorfl : uint16_t
	{
		Parerr = Bit2,
		Invcomm = Bit1,
		Frerr = Bit0,
	};
	MODM_FLAGS16(Errorfl)

	enum class Prog : uint16_t
	{
		Progver = Bit6,
		Progotp = Bit3,
		Otpref = Bit2,
		Progen = Bit0,
	};
	MODM_FLAGS16(Prog)

	enum class Diaagc : uint16_t
	{
		Magl = Bit11,
		Magh = Bit10,
		Cof = Bit9,
		Lf = Bit8,
	};
	MODM_FLAGS16(Diaagc)

	static constexpr uint16_t
	setMsbToEvenParity(const uint16_t num)
	{
		uint16_t par = 0x7fff & num;
		par ^= par >> 8;
		par ^= par >> 4;
		par ^= par >> 2;
		par ^= par >> 1;
		return ((par & 1) << 15) | (0x7fff & num);
	}

	enum class Register : uint16_t
	{
		ReadNop = setMsbToEvenParity(((1 << 14) | 0x0000)),
		ReadErrfl = setMsbToEvenParity(((1 << 14) | 0x0001)),
		ReadProg = setMsbToEvenParity(((1 << 14) | 0x0003)),
		ReadDiaagc = setMsbToEvenParity(((1 << 14) | 0x3FFC)),
		ReadMag = setMsbToEvenParity(((1 << 14) | 0x3FFD)),
		ReadAngleunc = setMsbToEvenParity(((1 << 14) | 0x3FFE)),
		ReadAnglecom = setMsbToEvenParity(((1 << 14) | 0x3FFF)),

	};

	struct modm_packed Data
	{
		/// @return
		constexpr float
		getAngleRad() const
		{
			const uint16_t angle = data & 0x3fff;
			return static_cast<float>(angle) / 16383.f * 2.f * std::numbers::pi_v<float>;
		}

		/// @return
		constexpr float
		getAngleDeg() const
		{
			const uint16_t angle = data & 0x3fff;
			return static_cast<float>(angle) / 16383.f * 360.f;
		}

		/// @return
		constexpr uint16_t
		getAngleRaw() const
		{
			const uint16_t angle = data & 0x3fff;
			return angle;
		}

		uint16_t data;
	};
};  // struct as5047
```
</details>

I think this might be related to [this](https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#1626): 
> The Standard should make clear that a constexpr member function cannot be used in a constant expression until its class is complete.

Do you have any suggestions?
Should I put everything into an `as5047` namespace?

I don't think moving the function as is to the `math:` is appropriate, since the function specifically calculates and sets bit 15 of a `uint16_t`, if moved to `math:`, a more generic version might be more appropriate. Also, other modm modules like `etlcpp` have their own functions to calculate parity bits...

src/modm/driver/encoder/as5047.lb
hshose: Don't know how these ended up there :cactus: should be fixed now.

src/modm/driver/encoder/as5047_impl.hpp
hshose: Fixed.

src/modm/driver/encoder/as5047_impl.hpp
hshose: Fixed.

src/modm/driver/encoder/as5047.hpp
salkinium: Ugh, C++ amiright!1!!

You can put it into a `modm::details` namespace then and use `/// @cond` and `/// @endcond` to block doxypress from indexing it ([Examples](https://github.com/search?q=repo%3Amodm-io%2Fmodm+%22namespace+detail%22+path%3A%2F%5Esrc%5C%2Fmodm%5C%2F%2F&type=code)). Prolly prefix the function with `as5047_` to avoid collisions if you wanna be super careful.

src/modm/driver/encoder/as5047.hpp
salkinium: Yup that makes sense.

src/modm/driver/encoder/as5047.hpp
hshose: > Ugh, C++ amiright!1!!

:cat2: No we are super careful :cat:

salkinium (approved):
Thanks!

PR #1139, opened 2024-03-08
[stm32] add advanced initialize on iwdg

victorandrehc:
This PR reinstates the ability to directly set the prescale and reload values on iwdg without letting modm to calculate this value directly.

### DIFF
--- a/src/modm/platform/iwdg/stm32/iwdg.hpp
+++ b/src/modm/platform/iwdg/stm32/iwdg.hpp
@@ -46,6 +46,15 @@ public:
 	};
 
 public:
+	static inline void
+	initialize(Prescaler prescaler, uint16_t reload)
+	{
+		writeKey(writeCommand);
+		IWDG->PR = uint32_t(prescaler);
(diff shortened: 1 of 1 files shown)


rleh (approved):

src/modm/platform/iwdg/stm32/iwdg.hpp
```diff
@@ -46,6 +46,11 @@ class Iwdg : public ::modm::PeripheralDriver
 	};
 
 public:
+	static inline void
+	initialize(Prescaler prescaler, uint16_t reload)
+	{
+		configure(prescaler, reload);
+	}
```
salkinium, line 56: Let's instead rename `configure` to `initialize` and move it here to avoid the unnecessary forwarding.

victorandrehc: Just pushed a version without the forwarding. Please also don't forget the PR #1136.

salkinium (approved):
Thanks!

PR #1141, opened 2024-03-08
Also adjust SpiMaster clock of F411-Disco in Board::initializeLis3(...)

TomSaw:
In PR `009bb0` the SystemClock::Frequency has been adjusted from 84 to 96 MHz without also adjusting the SpiMaster clock in `Board::initializeLis3(...)` from 5.25 to 6 MHz which leads to a failing static_assert.


salkinium (approved):
oh no, oh yes

PR #1142, opened 2024-03-10
[fix] Remove unused template parameter from GeneralPurposeTimer::signalToChannel()

TomSaw:
Looks like @chris-durand left some dev artifacts in there.

### DIFF
--- a/src/modm/platform/timer/stm32/advanced.hpp.in
+++ b/src/modm/platform/timer/stm32/advanced.hpp.in
@@ -338,7 +338,7 @@ public:
 	static void
 	setOutputIdleState(OutputIdleState idle, OutputIdleState idle_n = OutputIdleState::Reset)
 	{
-		constexpr auto channel = signalToChannel<Peripheral::Tim{{ id }}, Signal>();
+		constexpr auto channel = signalToChannel<Signal>();
 		setOutputIdleState(channel, idle, idle_n);
 	}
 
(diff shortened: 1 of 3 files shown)


salkinium (approved):
Thanks!

chris-durand (changes requested):
The code removed here actually serves an important purpose. It prevents using a signal that belongs to a different timer.

For instance, trying to write `Timer4::setCompareValue<GpioB7::Ch4>(0);` on an STM32G474 is now correctly rejected with

```
modm/src/modm/platform/gpio/data.hpp:1410:26: error: static assertion failed: GpioB7::Ch4 only connects to Tim3!
 1410 |         static_assert((p == Peripheral::Tim3),"GpioB7::Ch4 only connects to Tim3!"); };
      |                       ~~~^~~~~~~~~~~~~~~~~~~~
modm/src/modm/platform/gpio/data.hpp:1410:26: note: '(modm::platform::Peripheral::Tim4 == modm::platform::Peripheral::Tim3)' evaluates to false
```

The signal `GpioB7::Ch4` is channel 4 of another timer. With the change proposed here this would silently change the settings of another channel unrelated to that respective GPIO for the timer in question.

The static assert is triggered on construction of the empty `SignalConnection` object that is removed in this PR. No runtime code is emitted anyway. As the code works as expected, I'd recommend closing this PR.

salkinium:
Good catch! I completely forgot about how `SignalConnection` works…

PR #1147, opened 2024-03-12
[stm32] Fix flash page size for large F103, F105 and F107 devices

chris-durand:
Fix page size in flash driver for F103 devices with 256 kB flash or more, F105, and F107.

flash_f1


salkinium (approved):
Yes

PR #1148, opened 2024-03-14
[stm32/sam_x7x] Allow I2C interrupt priority to be set

rleh:
Allow the I2C interrupt priority to be set, for the other I2C driver not taken care of in #1143.
I just noticed it and wanted to prevent the interfaces from diverging.

cc @victorandrehc


chris-durand (approved):
Thanks!

salkinium (approved):
Thanks, pls merge I'm currently afk.

PR #1149, opened 2024-03-14
Fix Python syntax

rleh:
Since Python 3.6 backslash-character pairs that are not valid escape sequences are [deprecated](https://docs.python.org/dev/whatsnew/3.6.html#deprecated-python-behavior) and generate a `SyntaxWarning: invalid escape sequence` warning. This is fixed by using raw string.

The same problem is also fixed in modm-devices submodule, see modm-io/modm-devices#109.


salkinium (approved):
Yes

PR #1152, opened 2024-03-26
Couple of fixes for BlackMagicProbe

kapacuk:
- Added a missing import to `modm_tools/bmp.py`
- Changed `guess_serial_port()` to use `/dev/ttyBmpGdb` when guessing a BMP port


tools/modm_tools/bmp.py
rleh, line 1: Nice catch to change the file permissions to _755_ 👍🏽

tools/modm_tools/utils.py
rleh, line 1: Since this script can not be used directly, I would prefer to keep the _644_ permissions here.

rleh (commented):
Hi @kapacuk and welcome!

Thank you very much for your contribution. Looks fine, just two minor comments (see below).

tools/modm_tools/utils.py
kapacuk: You're right, I reverted it back to 644

rleh (approved):

salkinium (approved):
Awesome, thanks! (Please squash into one commit.)

salkinium:
Never mind, just remembered I can squash and merge from the GitHub UI. 🤦‍♂️

PR #1173, opened 2024-06-29
[docs] Add a reference to "debug mode" in assert.md

TomSaw:


rleh (commented):
Your link links to the scons bud system documentation, but the debug mode is build system independent.
There is this page (https://modm.io/reference/module/modm-build/), but it does not explain how to use the debug/release build modes.
I'm not sure what the best solution is.

salkinium:
Yeah, good point.

1. This should probably be "debug profile", to avoid confusion with using the debugger (eg. `scons debug`).
2. There should be a section above "Compiler Options" in https://modm.io/reference/module/modm-build/ that says we have two profiles: release and debug and that they can be selected with the `profile=release` argument to make and scons (CMake has their own definitions).

TomSaw:
> 2\. There should be a section above "Compiler Options" in https://modm.io/reference/module/modm-build/ that says we have two profiles: release and debug and that they can be selected with the `profile=release` argument to make and scons (CMake has their own definitions).

There's even an empty promise in the [scons](https://modm.io/reference/module/modm-build-scons/#scons) and [make](https://modm.io/reference/module/modm-build-make/#make) docs: 

> For a description of the `release` and `debug` profiles, see the `modm:build` module documentation.

However, I found clarification in the specific  scons and make docs and think this reference is good enough.

salkinium (approved):
Thank you!

PR #1177, opened 2024-07-01
[ext] Update submodules

salkinium:
We should really automate this.


rleh (approved):

salkinium:
LVGL still needs to be updated from v8 to v9 separately.

PR #1189, opened 2024-07-13
Fix python syntax in modm_tools

rleh:
With recent python versions there are a lot of warnings (`DeprecationWarning` until python 3.11, `SyntaxWarning` from version 3.12 onwards) generated from the modm_tools python code:
```bash
$ scons
/...blink/modm/modm_tools/bmp.py:13: SyntaxWarning: invalid escape sequence '\*'
  """
/.../blink/modm/modm_tools/build_id.py:13: SyntaxWarning: invalid escape sequence '\*'
[...]
```

Is it necessary to adapt the way this doc strings are extracted form the python files, @salkinium? I don't know where exactly that happens...

### DIFF
--- a/tools/modm_tools/avrdude.py
+++ b/tools/modm_tools/avrdude.py
@@ -10,7 +10,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # -----------------------------------------------------------------------------
 
-"""
+r"""
 ### AvrDude
 
 This tool simply wraps the `avrdude` command to provide two features:
--- a/tools/modm_tools/bmp.py
+++ b/tools/modm_tools/bmp.py
@@ -10,7 +10,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # -----------------------------------------------------------------------------
 
-"""
+r"""
 ### Black Magic Probe
 
 This tool wraps GDB to program an ELF file onto a target connected to a BMP.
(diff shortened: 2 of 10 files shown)


salkinium:
These docs are assembled here: https://github.com/modm-io/modm/blob/23f64ca4dda59484f73df8128ca701e7ea5503d5/tools/build_script_generator/module.lb#L20-L43

~~Best check with `lbuild discover -n :build` if the \* star is still displayed correctly.~~ I forgot that doesn't do any formatting, so it'll display `\*`. 🤦

salkinium:
Hm, I don't know, the [homepage built test](https://github.com/modm-io/modm/actions/runs/9918539479) shows that the \* star is also rendered `\*`.

salkinium (changes requested):
It makes the website show `\*` instead of `*`. [Do we want to use this one instead?](https://en.wikipedia.org/wiki/Asterisk#Encodings) ✱

calebchalmers:
Could you simply remove the slashes and put `(* *only ...*)` in the doc strings? I believe the first asterisk will be preserved in markdown because of the space after it.

rleh:
Using [raw string](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) seems to work fine:

<details>

<summary>Comparison screenshots 🖼️ 🖼️</summary>

image
image

</details>

salkinium (approved):
Awesome, thanks!

PR #1192, opened 2024-07-13
Add option to specify which repositories to generate SConscript files for

calebchalmers:
Currently, the `modm:scons:build` module will generate an SConscript for itself and every other repository included in the build process. This causes issues if another repository has its own method of generating an SConscript, since modm will overwrite it.

This PR changes the behaviour to only generate SConscript files for modm and any additional repos specified in `modm:scons:build:include_repos` (comma-separated list of repository names).

Note: this is a minor breaking change since *no additional repositories* will be included by default, whereas previously they would have been.

### DIFF
--- a/test/config/lbuild.xml
+++ b/test/config/lbuild.xml
@@ -6,7 +6,6 @@
   </repositories>
   <options>
     <option name="modm:build:scons:include_sconstruct">True</option>
-    <option name="modm:build:scons:include_repos">modm-test</option>
     <option name="modm:build:info.build">True</option>
     <option name="modm:build:info.git">Info+Status</option>
     <option name="modm:build:scons:cache_dir">$cache</option>
--- a/test/config/lbuild.xml
+++ b/test/config/lbuild.xml
@@ -6,6 +6,7 @@
   </repositories>
   <options>
     <option name="modm:build:scons:include_sconstruct">True</option>
+    <option name="modm:build:scons:include_repos">modm-test</option>
     <option name="modm:build:info.build">True</option>
     <option name="modm:build:info.git">Info+Status</option>
     <option name="modm:build:scons:cache_dir">$cache</option>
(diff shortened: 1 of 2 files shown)


calebchalmers:
Tests are failing due to:
```
  File "/__w/modm/modm/repo.lb", line 55, in <module>
    import modm_devices.parser
ModuleNotFoundError: No module named 'modm_devices'
```

Any ideas what the issue is here?

ext/mpaland/printf
salkinium, line 1: This is causing the git checkout to fail

calebchalmers: Ah, thanks. I'm not sure why that was in there

salkinium:
Good idea, but I don't like this implementation as it's not backward compatible and the option is not user friendly (although it could be defaulted I guess).

I'd rather you check the buildlog if any of the repos already generated a SConscript file in their output path, and then not generate ours (but still call it of course).

calebchalmers:
> I'd rather you check the buildlog if any of the repos already generated a SConscript file in their output path, and then not generate ours (but still call it of course).

Yeah I was trying to think of the best way to go about this. I like your idea to check for existing files, but would we need to worry about build order?

What if another repo generates its SConscript in `post_build`?

salkinium:
We need a `post_post_build` step!1!! Lol no, there's an undocumented `module.order :  int = 0` that you can set in the `init()` to enforce module order. Here we could set it to 10000 to enforce it's build after all other post_build steps.

salkinium (approved):
Could you check if my simplification still works correctly for your use case?

rleh (approved):
Nice 👍🏽 

Thank you a lot!

calebchalmers:
> Could you check if my simplification still works correctly for your use case?

Yep, works for me :+1:

PR #1193, opened 2024-07-14
[driver] rewrite adns9800

TomSaw:
The Adns9800-driver was made of pure blocking SPI calls - fixed that, added some features and eventually were directed to some clean, fibers-only implementation.

Writing a driver without the RF limitations was a pleasure! <strike>By this chance I wanna ask, if there's a (solid) concept/thoughts how we transition drivers to fibers?</strike>

- [ ] test `Adns9800::captureFrame()` in hardware
- [x] finish update of `examples/blue_pill_f103/adns9800`
- [x] cleanup `adns9800_data.hpp`
- [ ] fix Resumable Functions variant


src/modm/driver/motion/adns9800.fiber.hpp
```diff
@@ -0,0 +1,430 @@
...
+    uint8_t buffer[D::length];
+
+    readTransaction([this, &buffer]() {
+      SpiMaster::transfer(static_cast<uint8_t>(Register::Motion_Burst));
+      modm::this_fiber::sleep_for(shutter_config.getOneFrameTime());
+      SpiMaster::transfer(nullptr, buffer, sizeof(buffer));
+    });
+
+    return D(buffer);
+  }
```
TomSaw, line 331: Contrary to modm's standard,  `Adns9800::read<>()` **returns** adns9800::Data*.
This way, i could implement `read<>()` for different variants of Data flat and simple.

I also thougt, modm's default driver design - using a Data& property  - may have snapped due to the limitations of RFs !?

src/modm/driver/motion/adns9800.fiber.hpp
```diff
@@ -0,0 +1,430 @@
...
+  /**
+   * @tparam Cpi    Counts per inch
+   *                Allowed values: 200 to 8200 in steps of 200
+   */
+  template<int Cpi>
+  requires (Cpi >= 200) and (!Cpi % 200 == 0) and (Cpi <= 8200)
+  struct Resolution: public modm::Register8 {
+    Resolution() : modm::Register8(Cpi / 200) {};
+  };
+  using Resolution_t = modm::Register8;
```
TomSaw, line 86: Instead of writing all the options from 200, 400, 600, 800, ... , 8200 i came up with this.

Should be typesafe as it is, no?

src/modm/driver/motion/adns9800.fiber.hpp
```diff
@@ -0,0 +1,430 @@
...
+  /**
+   * @tparam Cpi    Counts per inch
+   *                Allowed values: 200 to 8200 in steps of 200
+   */
+  template<int Cpi>
+  requires (Cpi >= 200) and (Cpi % 200 == 0) and (Cpi <= 8200)
+  struct Resolution: public modm::Register8 {
+    Resolution() : modm::Register8(Cpi / 200) {};
+  };
+  using Resolution_t = modm::Register8;
```
TomSaw: Instead of writing all the options from 200, 400, 600, 800, ... , 8200 i came up with this.

It constraints allowed values without overloading the corresponding `Adns9800::set(...)`.

Alternatively, something with `constexpr` would allow both, compile time and runtime scaling by 1/200 of the argument. let's see...

src/modm/driver/motion/adns9800.fiber.hpp
```diff
@@ -0,0 +1,430 @@
...
+  read() {
+    uint8_t buffer[D::length];
+
+    readTransaction([this, &buffer]() {
+      SpiMaster::transfer(static_cast<uint8_t>(Register::Motion_Burst));
+      modm::this_fiber::sleep_for(shutter_config.getOneFrameTime());
+      SpiMaster::transfer(nullptr, buffer, sizeof(buffer));
+    });
+
+    return D(buffer);
```
TomSaw: EDITED AGAIN:
I've crossed modm's default here to add support for multiple `Data` types.
Problem is, the binding of `Data&` to the driver class limits the programmer to one type or multiple, symmetric driver instances 😖. (Is this some kind of RF relict?)

#### Handling `Data` exclusively via the emitting method `Device::read(...)` feels better:

Either a fresh constructed Data is returned from Device::read(...) or - as a compromise to the current default - a `Data&` is passed to `Device::read(...)`.

With the "`Data&` in `Driver`"-design, responsibility for the integrity of `Driver::Data&`s is on `Driver` due to encapsulation principles -> A mutex on `Driver::Data& data` within `Driver::read(...)` might become fundamentally at some point!?

In contrast, when returning newly constructed `Data`, the API user could take appropriate measures. Returning `Data` also facilitates RVO / Copy Elision in typical situations, f.e. when forwarding `Data` samples from `Driver::read(...)` into a `Container` for further processing.

TomSaw:
Not intended to be a dizz, but i've stroken Sascha Schade from the copyright: There's nothing left from the previous driver-port but the lbuild firmware select option.

src/modm/driver/motion/adns9800.lb
```diff
@@ -27,15 +28,28 @@ def prepare(module, options):
             enumeration=["a4", "a4b", "a5", "a6"],
             description="Select the firmware version to initialize the device with.",
             default="a6"))
+    module.add_option(
+        BooleanOption(
+            name="use_fiber",
+            description="Use an advanced driver without backwards compatibilty for protothreads.",
+            default="false"))
```
TomSaw: Borrowed this opt-in option from `processing:protothreads` to pick the pure fibers driver implementation.
For consistency, this option would be added globally for each `modm:driver:*` but this also floods the output of `lbuild discover`.

In hope for a better idea.

salkinium: nah, just depend on the `:processing:fiber` module directly and ignore backward compatibility. GCC13 refuses to compile the RF_YIELD macro, cos it's undefined behavior, so we cannot support resumables anymore. The proper way to do this would be to use C++23 coroutines, but unfortunately the APIs are different so that it cannot be done without significant code changes. Let's just use fibers, it will solve \*a lot\* of problems in the future.

src/modm/driver/motion/adns9800.hpp
```diff
@@ -9,95 +9,426 @@
...
-	static void
-	writeReg(Register const reg, uint8_t const data);
+  /**
+   * @brief      powerUp() resets the device's internal state after power-loss or user invoked shutdown().
+   */
+	void
+  powerUp() {
+    /// @see power-up sequence, datasheet page 20
+    Cs::reset();
+    modm::this_fiber::sleep_for(10ns);
```
salkinium: Use `modm::delay(10ns)`. The fiber sleep is only implemented for microseconds, since a context switch takes around 1us. (for comparison one cycle at 100MHz = 10ns<sup>-1</sup>.) (even on 550MHz STM32H7 it takes ~250ns).

src/modm/driver/motion/adns9800.hpp
```diff
@@ -9,95 +9,426 @@
...
+
+  /**
+   * @brief     Verify presence of the correct device by reading and checking it's product and revision id.
+   */
+  bool
+  verifyIdentity() {
+    bool isValid = true;
+
+    const uint8_t product_id = readRegister(Register::Product_ID);
+    isValid &= modm_assert_continue_fail(product_id == uint8_t(0x33), "adns9800.id", "Unexpected Product Id", product_id);
```
salkinium: return an error enum instead? perhaps castable to bool or something like that

src/modm/driver/motion/adns9800.hpp
```diff
@@ -9,95 +9,426 @@
...
+    while(not ready_to_read.isExpired()) modm::this_fiber::yield();
+    while(not this->acquireMaster()) modm::this_fiber::yield();
+    Cs::reset();
+
+    closure();
+
+    ready_to_write.restart(20us);      // tSRW: [t]ime [S]pi between [R]rite and [W]rite
+    ready_to_read.restart(20us);       // tSRR: [t]ime [S]pi between [R]rite and [R]ead
+
+    modm::this_fiber::sleep_for(120ns); // tSCLK_NCS_read: Cs persistance after last SCLK for read
```
salkinium: Hm ok, the use case is perfectly valid, but I think this is currently cast to 0us and then returns immediately. 😬 oops

I think the sleep functions should just call `modm::delay_ns` if the unit is `ns`.

src/modm/driver/motion/adns9800.hpp
```diff
@@ -9,95 +9,426 @@
...
+  }
+
+  template<class Callable>
+  void
+  writeTransaction(Callable&& closure) {
+    while(not ready_to_write.isExpired()) modm::this_fiber::yield();
+    while(not this->acquireMaster()) modm::this_fiber::yield();
+    Cs::reset();
+
+    closure();
```
salkinium:
```suggestion
    std::forward<Closure>(closure)();
```
#CosCppBullshitAboutPerfectForwarding

salkinium (commented):
I'll review some more next week.

src/modm/driver/motion/adns9800.hpp
TomSaw: Thought of just some registers to rotate on context switches but 1us(!) sounds like there's more 😅. Assembler is not (yet) in my repertoire but now I feel the need to digg deeper into fibers gearwork.

- 
To fix this pitfall `sleep_for(...)` could branch to `delay(...)` for durations below an appropriate threshold. With the MCUs clock by hand, this could even be scaled as needed.

salkinium: Yes, I agree those details should be all taken care of by the function itself. 
Have to think about this some more, for example, on AVRs, a context switch takes 10us, so if you sleep for just 1us in a loop, it'll never yield…

(The [context switch code is here](https://github.com/modm-io/modm/blob/develop/src/modm/processing/fiber/context_arm_m.cpp.in#L15-L68) btw)

src/modm/driver/motion/adns9800.hpp
TomSaw: The lambda capture counts as argument the perfect forwarding is required for, right?

TomSaw: aiiieee 🙈

src/modm/driver/motion/adns9800.fiber.hpp
salkinium: Historically, this was optimizaed for placing all data objects into a big struct that was then serialized directly over a comms network. However this is not how you would do it today anymoe.

Feel free to completely refactor it to your liking.

TomSaw:
Just wanna tell: I'll finish this in the beginning of September.

PR #1195, opened 2024-07-21
Implement some modm containers as wrappers of standard library types

chris-durand:
Subset of #1020 

- [x] Implement modm::Pair using std::pair
- [x] Implement modm::*List using std::list
- [x] Implement modm::DynamicArray using std::vector

There is no direct equivalent of `modm::LinkedList` in the standard library. Thus, it is also implemented as a doubly-linked list. `std::forward_list` is not directly usable because it can't append to both ends of the list in O(1) and doesn't track the size.


salkinium (approved):
Awesome!

PR #1188, opened 2024-07-12
Add STM32C0x

Apehaenger:
Add STM32C0x Summary:

- [x] RCC
- [x] Timer
- [x] ADC
- [x] Device driver https://github.com/modm-io/modm-devices/pull/112
- [x] USART
- [x] Examples

---

In relation to #1187, I tried to add STM32C011F6P6 support for a custom PCB.

Well, somehow hard for an ungifted to jump into the internals, who's intention was to use modm for his project ;-)
However, can't become more dumb by trying :-)

Unfortunately I got stuck after a couple of lines and don't know if I missed something to hack in modm or if I made something wrong with my minimal blink example :-/
If I chose a supported MCU it compiles well, but all supported MCU's I found, do also have a board, so... 

This is what I get now:
```
jebeling@lt-je:~/Development/OpenMowerProject/xESC_YF_rev4-adapter/firmware/ext/modm/examples/stm32c0x1/blink$ lbuild build
>> modm: Recomputing device cache...

ERROR: In 'modm:build':

Traceback (most recent call last):
  File "/home/jebeling/Development/OpenMowerProject/xESC_YF_rev4-adapter/firmware/ext/modm/tools/build_script_generator/module.lb", line 251, in post_build
    all_rams = [m for m in linkerscript.get("memories") if "w" in m["access"]]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

Any hints are welcome ;-)

PR #1197, opened 2024-08-12
ADC sequence of G0 (and C0) devicec have  wrong channel shift?

Apehaenger:
While still working on my STM32C0x implementation #1188, I had a head-cruncher by building a ADC-Sequence example.
For me it looks like the implementation is wrong as it shifts the ADC channel only by 1 bit instead of 4.
So in my opinion ADC sequences can't work for ~~F0,~~ G0 as well as C0 devices?!


ext/modm-devices
Apehaenger, line 1: Sorry, the updated submodule is only related to my C0x implementation

chris-durand (commented):
Thanks for fixing this!

Register layout from the G0 manual for reference:
Screenshot_20240812_225400

The unrelated submodule update is still there. Could you remove it?

Apehaenger:
The register layout for C0 look equal.

Regarding the unrelated submodule:
~~I would be happy if I could, but I'm unsure how ;-)
Rewind my modm-device submodule to standard and commit again?~~
Done

salkinium (approved):
Thanks!

Issue #1201, opened 2024-08-19
CMSIS DSP error when targeting hosted platforms


calebchalmers:
When building a project with the CMSIS DSP module for a hosted target, I get the following error:

`modm/ext/cmsis/dsp/arm_math.h:34:10: fatal error: modm/platform/device.hpp: No such file or directory`

This is due to my recent PR about including `device.hpp` in `arm_math.h` to provide configuration; however, for hosted platforms it seems like modm doesn't generate a `device.hpp` file.

salkinium:
Ah interesting, I never thought about building CMSIS-DSP for hosted but according to the docs that is supported. Probably also a good idea for testing etc.

Hosted does not generate a `device.hpp` mostly because hosted doesn't have a specific hardware. I would prefer to not include it in the `arm_math.h` for hosted builds. I hope the CMSIS-DSP has sensible defaults for hosted.

salkinium:
I'm closing this as there are probably better ways to run CMSIS-DSP on Hosted than via the rather minimal modm support.
There's also an official [Python Wrapper with NumPy support](https://pypi.org/project/cmsisdsp/) which seems more friendly to quick development anyways.

PR #1203, opened 2024-08-20
[ext] Check for device.hpp before including in CMSIS-DSP

calebchalmers:
Checks for the existence of `modm/platform/device.hpp` before including it in `arm_math.h`. This fixes CMSIS-DSP not compiling when running on a hosted platform (#1201).

### DIFF
--- a/ext/arm/arm_math.h.in
+++ b/ext/arm/arm_math.h.in
@@ -31,7 +31,9 @@ extern "C"
 {
 #endif
 
+#if __has_include(<modm/platform/device.hpp>)
 #include <modm/platform/device.hpp>
+#endif
 
 /* Local configuration file */
 #if __has_include(<arm_math_local.h>)


salkinium:
Could you add an example for Linux using CMSIS-DSP? Otherwise future refactoring will break this again.

calebchalmers:
Okay so after trying to add an example for this, I realized that the CMSIS DSP module isn't even available when using `hosted-linux`... So something seems wrong on our project's end, since it's including those DSP headers somehow.

salkinium:
Any update on this? I still find this a valuable idea!

calebchalmers:
I've been busy with classes lately, so I haven't had a chance to dig through our project to find what was causing this issue yet :(

If you want this to be added anyway, we would also have to also allow the DSP module to be used in hosted environments. Removing the check for that shouldn't be hard, but idk if it would simply work or require more effort?

salkinium:
I'm closing this as there are probably better ways to run CMSIS-DSP on Hosted than via the rather minimal modm support.
There's also an official [Python Wrapper with NumPy support](https://pypi.org/project/cmsisdsp/) which seems more friendly to quick development anyways.

PR #1210, opened 2024-10-01
[driver] DW3110 UWB radio

Javask:
Reopening this under my name on request by @rleh.

Driver and example are working and have been tested on real hardware. 
Most of the important features should be implemented.

### DIFF
--- a/examples/nucleo_f401re/dw3110-communication/main.cpp
+++ b/examples/nucleo_f401re/dw3110-communication/main.cpp
@@ -1,0 +1,170 @@
+/*
+ * Copyright (c) 2024, Elias H.
+ * Copyright (c) 2024, Raphael Lehmann
+ * Copyright (c) 2024, Michael Jossen
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <modm/board.hpp>
+#include <modm/debug/logger.hpp>
+#include <modm/driver/radio/dw3110/dw3110_phy.hpp>
+#include <modm/processing/protothread.hpp>
+#include <modm/processing/timer.hpp>
+
+using namespace Board;
+using namespace std::chrono_literals;
+
+using MySpiMaster = modm::platform::SpiMaster1;
+using MyDw3110_a = modm::Dw3110Phy<MySpiMaster, GpioB6>;
+using MyDw3110_b = modm::Dw3110Phy<MySpiMaster, GpioA10>;
+
+class TXThread : public modm::pt::Protothread
+{
+public:
+	bool
+	init()
+	{
+		auto ret = RF_CALL_BLOCKING(radio.initialize(
+			modm::Dw3110::Channel::Channel9, modm::Dw3110::PreambleCode::Code_64Mhz_9,
+			modm::Dw3110::PreambleLength::Preamble_128,
+			modm::Dw3110::StartFrameDelimiter::Decawave_8));
+		RF_CALL_BLOCKING(radio.setEnableLongFrames(true));
+		return ret;
+	}
+
+	bool
+	run()
+	{
+		PT_BEGIN();
+		while (true)
+		{
+
+			txdata[txdata.size() - 1]++;
+			timeout.restart(Button::read() ? 500ms : 10ms);
+			PT_WAIT_UNTIL(timeout.execute());
+			if (PT_CALL(radio.transmit(txdata, true)) == MyDw3110_b::Error::None)
+			{
+				sentCount++;
+			} else
+			{
+				MODM_LOG_DEBUG << "[TX] Failed to trasmit!" << modm::endl;
+			}
+		}
+		PT_END();
+	}
+
+	size_t
+	getCount()
+	{
+		return sentCount;
+	}
+
+private:
+	MyDw3110_b radio{};
+	std::array<uint8_t, 5> txdata = {0xBA, 0xDE, 0xAF, 0xFE, 0x00};
+	modm::Timeout timeout{10ms};
+	size_t sentCount{0};
+};
+
+class RXThread : public modm::pt::Protothread
+{
+public:
+	bool
+	init()
+	{
+		auto ret = RF_CALL_BLOCKING(radio.initialize(
+			modm::Dw3110::Channel::Channel9, modm::Dw3110::PreambleCode::Code_64Mhz_9,
+			modm::Dw3110::PreambleLength::Preamble_128,
+			modm::Dw3110::StartFrameDelimiter::Decawave_8));
+		RF_CALL_BLOCKING(radio.setEnableLongFrames(true));
+		return ret;
+	}
+
+	bool
+	run()
+	{
+		PT_BEGIN();
+		while (true)
+		{
+			while (!PT_CALL(radio.packetReady()))
+			{
+				if (!PT_CALL(radio.isReceiving()))
+				{
+					// KEEP ON SEPERATE LINE
+					PT_CALL(radio.startReceive());
+				}
+				PT_YIELD();
+			}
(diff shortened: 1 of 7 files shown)


salkinium (approved):
Thanks!

cc @rleh I assume you already reviewed this in the RCA, so I can just merge it?

chris-durand:
@salkinium I'm taking a look at this tomorrow evening. Wanted to review it since some time but never got to it. Can we wait one more day?

examples/nucleo_f401re/dw3110-communication/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	RXThread rx;
+	if (!rx.init())
+	{
+		MODM_LOG_ERROR << "Failed to initialize TR Device!" << modm::endl;
+		success = false;
+	}
+	if (!success)
+		while (true) {}
```
chris-durand: An endless loop without side effects is undefined behaviour in C++. It works with gcc in most cases but clang, for example, is very aggressive at [removing them](https://gcc.godbolt.org/z/Mf1q7MY79). You could add `__NOP()` or something similar here.

src/modm/driver/radio/dw3110/dw3110_definitions.hpp
```diff
@@ -0,0 +1,452 @@
...
+	static constexpr Register CHAN_CTRL{.bank = GEN_CFG_AES_1, .offset = 0x14, .length = 2};
+	static constexpr Register LA_PEND_01{.bank = GEN_CFG_AES_1, .offset = 0x18, .length = 4};
+	static constexpr Register LA_PEND_23{.bank = GEN_CFG_AES_1, .offset = 0x1C, .length = 4};
+	static constexpr Register SPI_COLLISION{.bank = GEN_CFG_AES_1, .offset = 0x20, .length = 1};
+	static constexpr Register RDB_STATUS{.bank = GEN_CFG_AES_1, .offset = 0x24, .length = 1};
+	static constexpr Register RDB_DIAG{.bank = GEN_CFG_AES_1, .offset = 0x28, .length = 1};
+	static constexpr Register AES_CFG{.bank = GEN_CFG_AES_1, .offset = 0x30, .length = 2};
+	static constexpr Register AES_IV0{.bank = GEN_CFG_AES_1, .offset = 0x34, .length = 4};
+	static constexpr Register AES_IV1{.bank = GEN_CFG_AES_1, .offset = 0x38, .length = 4};
+	static constexpr Register AES_IV2{.bank = GEN_CFG_AES_1, .offset = 0x3C, .length = 2};
```
chris-durand: Screenshot_20241104_233305

This says length = 4 and there is another register AES_IV4 that's missing.

I know AES is unsupported, so it doesn't matter much.

src/modm/driver/radio/dw3110/dw3110_phy.hpp
```diff
@@ -0,0 +1,472 @@
...
+#include <modm/processing/timer/timeout.hpp>
+#include <span>
+
+#include "dw3110_definitions.hpp"
+
+namespace modm
+{
+
+/**
+ * Hardware abstraction layer for DW3110 \n
```
chris-durand, line 30: Is the `\n` here on purpose for formatting the doxygen output or is it a mistake?

src/modm/driver/radio/dw3110/dw3110_phy_impl.hpp
```diff
@@ -0,0 +1,1415 @@
...
+modm::Dw3110Phy<SpiMaster, Cs>::checkResult(std::span<const uint8_t, Len> expected,
+											std::span<const uint8_t, Len> got)
+{
+	bool result = true;
+	for (size_t i = 0; i < Len; i++)
+	{
+		if (expected[i] != got[i]) { result = false; }
+	}
+	return result;
+}
```
chris-durand: Isn't this [`std::ranges::equal()`](https://en.cppreference.com/w/cpp/algorithm/ranges/equal) ? It works on `span`s, like `std::ranges::equal(span1, span2)`. There should be no need to have this member function.

src/modm/driver/radio/dw3110/dw3110_phy.hpp
```diff
@@ -0,0 +1,472 @@
...
+	modm::ResumableResult<bool>
+	fetchPacket(std::span<uint8_t> payload, size_t &payload_len);
+
+	/// Transmit a given package using the current configuration using a given transmission mode
+	/// @param payload Span to the desired payload
+	/// @param ranging Decides whether or not to set the ranging bit in the header
+	/// @param fast Decides if the data portion is sent at 850kbps or 6.8Mbps
+	template<TXMode tmode = TXMode::Default>
+	modm::ResumableResult<Error>
+	transmit(const std::span<const uint8_t> payload, bool ranging = true, bool fast = true);
```
chris-durand, line 262: Just fyi, the first `const` here does nothing. `const`-qualifying a pass-by-value argument in a function declaration has no effect. In the definition, though, it makes the span itself `const`, e.g. you can't reassign what the span points to.
It is sufficient to have `std::span<const uint8_t>` to make the data `const`. The way it's used here looks unintentional because it's not consistent with all the other value arguments which aren't spans being non-`const`.

src/modm/driver/radio/dw3110/dw3110_phy_impl.hpp
```diff
@@ -0,0 +1,1415 @@
...
+	}
+
+	// Loop until valid state
+	timeout.restart(1ms);
+	while (chip_state != Dw3110::SystemState::IDLE_RC &&
+		   chip_state != Dw3110::SystemState::IDLE_PLL)
+	{
+		if (timeout.execute())
+		{
+			MODM_LOG_ERROR << "Timeout waiting for IDLE State!" << (int)chip_state << modm::endl;
```
chris-durand, line 56: `std::to_underlying(chip_state)` is less error-prone than the C-style cast. It's impossible to get the type wrong.

chris-durand (commented):
Very nice, thanks!

src/modm/driver/radio/dw3110/dw3110_phy_impl.hpp
Javask: to_underlying would resolve to uint8_t here which would be printed as char not as int, so casting to int was desired.

src/modm/driver/radio/dw3110/dw3110_phy.hpp
Javask: This was for formatting. Doxygen put everything on a single line by default.

src/modm/driver/radio/dw3110/dw3110_definitions.hpp
Javask: Good catch! I must have missed AES_IV4 since it is not present in the overview.

examples/nucleo_f401re/dw3110-communication/main.cpp
salkinium: Or use `modm_assert`

Javask:
Fixed the register definition, removed the const specifier, removed the checkResult member function and replaced with std::ranges::equal and added __NOP() to the while loop.
Also made a small change in the example to change the tx buffer length, requested by @rleh since the buffer might contain uninitialized values that get transmitted.
I don't have hardware on hand until this evening, so the changes have not yet been run on hardware. I will retest the example this evening just to make sure.

Javask:
Still works.

src/modm/driver/radio/dw3110/dw3110_phy_impl.hpp
chris-durand: The modm streams only print `char` as a character but `unsigned char` (`uint8_t`) and `signed char` (`int8_t`) as integers. 

```c++
MODM_LOG_ERROR << uint8_t(100) << ", " << int8_t(100) << ", " << char(100);
```
prints `100, 100, d`.

You can leave the code as is. It's not critical here.

chris-durand (approved):
Thanks!

PR #1212, opened 2024-10-23
WIP: RP2040 PIO support

andryblack:
Initial support of RP2040 PIO hardware.
Compile time C++ assembler with syntax closed to original.

todo:
- [x] write all opcodes arguments
- [ ] work with dma

### DIFF
--- a/examples/rp_pico/pio_ws2812/main.cpp
+++ b/examples/rp_pico/pio_ws2812/main.cpp
@@ -1,0 +1,144 @@
+/*
+ * Copyright (c) 2016, Sascha Schade
+ * Copyright (c) 2017, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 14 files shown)


salkinium:
Very, very awesome btw! I also looked into this and wanted to be able to run the existing pio programs, so I wanted to create a pioasm template for modm, which I started work on here: https://github.com/modm-ext/pico-sdk-partial/tree/pioasm. 
I never finished because I realized that I needed the ASM part too, which you're working on now. Perhaps you can continue that if you want.

andryblack:
> Very, very awesome btw! I also looked into this and wanted to be able to run the existing pio programs, so I wanted to create a pioasm template for modm, which I started work on here: https://github.com/modm-ext/pico-sdk-partial/tree/pioasm. I never finished because I realized that I needed the ASM part too, which you're working on now. Perhaps you can continue that if you want.

The main idea is not using tool for generating PIO code. Existing PIO programs need adapted in any case, because constaint framework specific initialisation code.  Programs code not big in size and ported as one to one. I updated syntax: removed excess braces.

PR #1213, opened 2024-10-23
[tools] fix with Python 3.13

andryblack:
With Python 3.13 telnetlib is removed from base modules

### DIFF
--- a/tools/modm_tools/jlink.py
+++ b/tools/modm_tools/jlink.py
@@ -39,7 +39,6 @@ import os
 import time
 import signal
 import platform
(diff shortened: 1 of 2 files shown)


salkinium (approved):
I've seen this, but I'm still struggling with pushing an update to PyPi for include telnetlib3 in [the `modm` package](https://github.com/modm-ext/modm-cli). That's why the CI is still failing…

salkinium (approved):
Finally uploaded the new `modm` package with the `telnetlib3` dependency.

PR #1217, opened 2024-10-26
[rp] spi: add missing include

andryblack:


salkinium (approved):
Yes thanks, I added the lbuild dependency.

PR #1220, opened 2024-11-05
[linux] Enable SocketCAN to receive FDCan frames

Javask:
Enables socketcan to receive FDCan frames and send FDCan frames if the payload length is bigger than 8.


salkinium (approved):
Thanks!

salkinium:
I assume you tested this in hardware?

Javask:
I have tested it using a vcan interface, FD and normal can frames both work fine. I havent used in on a usb can adapter yet but that should work the same.

PR #1229, opened 2024-11-20
[timer] Fix Timeout::wait implementation

### LINKED ISSUE #1227: this_fiber::poll() condition polarity inconsistent
Regarding the implementation of `this_fiber::poll()`:

https://github.com/modm-io/modm/blob/7b571a67ac525addbe6698f1b15963d4af0563a3/src/modm/processing/fiber/functions.hpp#L56-L65

My reading of the implementation is that it will block while the condition returns false. However, the docstring says it blocks while the condition returns true.

This also indicates that the following usages are incorrect:
1. The fiber tests use trivial "return true" conditions which I believe are intended to block indefinitely but actually return immediately. Although, I don't really understand what these tests are intended to do. They aren't actually running in a fiber as far as I can tell. https://github.com/modm-io/modm/blob/7b571a67ac525addbe6698f1b15963d4af0563a3/test/modm/processing/fiber/fiber_test.cpp#L104-L120
2. The `Timeout::wait` function returns immediately. https://github.com/modm-io/modm/blob/7b571a67ac525addbe6698f1b15963d4af0563a3/src/modm/processing/timer/timeout_impl.hpp#L129-L134

I did a grep for `this_fiber::poll` and inspected each call site in the repo to convince myself that they're correct. I _believe_ these are the only two buggy ones.

I'm not sure if the desired change is to update the documentation and two mentioned call sites or change the function behavior, so just posting an issue for now.

salkinium:
Fixes #1227.

Thanks for finding this!
Added a unit test to catch any regression in future.

cc @WasabiFan

### DIFF
--- a/src/modm/io/iostream.hpp.in
+++ b/src/modm/io/iostream.hpp.in
@@ -306,10 +306,19 @@ inline IOStream&
 flush(IOStream& ios)
 { return ios.flush(); }
 
-//// Write a newline. **DOES NOT FLUSH THE STREAM!**
+/// Write a newline. **DOES NOT FLUSH THE STREAM!**
 inline IOStream&
 endl(IOStream& ios)
+%% if is_hosted
+{
+	ios.endl();
+	ios.flush();
+	return ios;
+}
+%% else
 { return ios.endl(); }
+%% endif
+
 
 /// set the output mode to binary style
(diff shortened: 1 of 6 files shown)


WasabiFan (approved):
Thanks!

PR #1231, opened 2024-11-22
fixed lvgl example for nucleo-l452re

### LINKED ISSUE #1228: nucleo-l452-re lvgl example wrong display buffering mode
The example for rendering on the ili9341 via lvgl in examples/nucleo_l452re/lvgl/main.cpp uses the direct draw mode for d´te display, but the buffer is not screen sized. Quote from the [documentation](https://docs.lvgl.io/9.2/porting/display.html): 
> " LV_DISPLAY_RENDER_MODE_DIRECT: The buffer(s) has to be screen sized"

I think the mode should be partial for this buffer as a screen sized buffer could be to large for onboard memory.
Also had an issue with the buffer being not aligned, adding modm_aligned(4) as attribute helped.

frnktank:
`modm/examples/nucleo_l452re/lvgl/main.cpp`:
Changed buffering mode in lvgl example to "partial" to accomodate for the smaller than screen-sized buffer. Also aligned the buffer in memory because of lvgl requiring a aligned buffer.

### DIFF
--- a/examples/nucleo_l452re/lvgl/main.cpp
+++ b/examples/nucleo_l452re/lvgl/main.cpp
@@ -63,7 +63,7 @@ modm::Touch2046<touch::Spi, touch::Cs> touchController;
 
 
 static constexpr size_t buf_size = LV_HOR_RES_MAX * LV_VER_RES_MAX / 8;
-static lv_color_t buf[buf_size];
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Thanks!

PR #1234, opened 2024-11-24
[iwdg] Fix reversed prescaler enum index for STM32

### LINKED ISSUE #1233: `modm::platform::Iwdt` chooses wrong prescaler index
With the reversal of valid prescaler values (256, 4), one also needs to adjust the resulting index that determines the `Prescaler `from the corresponding `enum`.

I can see two ways of fixing that: 

- either go through the possible prescaler values in ascending order [4, 256] 
- or correct the resulting index by `static_cast<uint8_t>(Prescaler::Div256)-result.index`

salkinium:
Fixes #1233.

cc @jgmess-dlr can you test this please?


jgmess-dlr:
Works for me on STM32H753. Thanks for the effort!

jgmess-dlr (approved):
Fixes the issue as proposed.

PR #1236, opened 2024-12-01
[stm32] Use SCB_CleanInvalidateDCache() on startup

### LINKED ISSUE #1235: [stm32] `SCB_InvalidateDCache()` might crash on startup
In case the DCache was already enabled before startup, for instance because the application is not starting from the internal STM bootloader but something like u-boot, that already uses the cache, a simple `SCB_InvalidateDCache()` might cause the application to crash.
Instead, a call to 'SCB_CleanInvalidateDCache()' mitigates the problem and has the same overall effect.

jgmess-dlr:
Closes #1235

### DIFF
--- a/src/modm/platform/core/cortex/startup.c.in
+++ b/src/modm/platform/core/cortex/startup.c.in
@@ -106,7 +106,7 @@ void __modm_startup(void)
 %% endif
 %% if with_dcache
 	// Enable data cache with default WBWA policy
-	SCB_InvalidateDCache();
+	SCB_CleanInvalidateDCache();
 	SCB_EnableDCache();
 %% endif
 %#


salkinium (approved):
Thanks!

For my future self: The I-Cache does not have a `SCB_CleanInvalidateICache()` function.

jgmess-dlr:
> Thanks!
> 
> For my future self: The I-Cache does not have a `SCB_CleanInvalidateICache()` function.

Yeah, I tried that at first, too ;)

salkinium:
FYI: This doesn't work if the D-Cache hasn't been enabled before.
On my DISCO-F746NG board, the `SCB_CleanInvalidateDCache()` function leads to a hardfault.
But when I first enable the D-cache and then clean/invalidate it, it works fine:

```diff
-SCB_CleanInvalidateDCache();
+SCB_EnableDCache();
+SCB_CleanInvalidateDCache();
```

I've added a fix to https://github.com/modm-io/modm/pull/1247

PR #1252, opened 2025-01-05
[stm32] add missing HSI48-related functionality

### LINKED ISSUE #1250: Missing `Rcc::enableInternalClockMHz48` Function for Some STM32 Microcontrollers
I am using the STM32G431KBT6 microcontroller and would like to enable HSI48 for the USB peripheral. However, I found that only the function `Rcc::setClock48Source(Rcc::Clock48Source::Hsi48)` is available, but there is no corresponding function to enable HSI48.  

I noticed that the library does include the function `Rcc::enableInternalClockMHz48`, but it is not generated in my code.

https://github.com/modm-io/modm/blob/ab7c1ca29a2bcc47e02ba1a991b6e138549d5303/src/modm/platform/clock/stm32/rcc.cpp.in#L44

dokee39:
Fixes #1250

- add `Rcc::enableInternalClockMHz48` function for stm32 g0 g4 l0 h7 (h5 u0)
- make changes to `PllSource`, `SystemClockSource`, `Clock48Source` and `UsbClockSource`

| family | hsi48_cr |
|:-------|:---------|
| F0 (42, 48, 71,72,78, 91, 98) | CR2 |
| G0 (B1 C1) | CR |
| G4 | CRRCR |
| L0 (x2) | CRRCR |
| L4 (`target["name"][0] not in ["7", "8"]`) | CRRCR |
| U0 | CRRCR |
| L5 | CRRCR |
| U5 | CR |
| H5 | CR |
| H7 | CR |

### DIFF
--- a/src/modm/platform/clock/stm32/module.lb
+++ b/src/modm/platform/clock/stm32/module.lb
@@ -69,15 +69,16 @@ def build(env):
     properties["pllprediv2"] = False    # FIXME: not sure what value this should have
     properties["pll_hse_prediv2"] = target["family"] == "f1" and target["name"] in ["01", "02", "03"]
     properties["hsi48"] = \
+        (target["family"] in ["g4", "h5", "h7", "l5", "u0", "u5"]) or \
         (target["family"] == "f0" and target["name"] in ["42", "48", "71", "72", "78", "91", "98"]) or \
-        (target["family"] == "l4" and target["name"][0] not in ["7", "8"]) or \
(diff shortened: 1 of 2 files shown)


dokee39:
I overlooked that some models do not have USB peripherals, which caused the code check to fail. It seems I need to revise and resubmit the PR (×_×;)"

salkinium (approved):
Awesome, thank you very much!

PR #1255, opened 2025-01-17
[cortex-m7] Enable I/D-Cache via lbuild

salkinium:
Only enable the D-Cache *after* the memory operations, so that all data is committed to the memory when required (for example for the VTOR in RAM!!!)

Related to https://github.com/modm-io/modm/pull/1236.


salkinium:
@jgmess-dlr Could you check if this change still works for your use case? I discovered stability issues when uploading using a JLINK. I think the main issue is that VTOR is set to RAM by default on Cortex-M4, but the VTOR may not actually have been copied to RAM and instead is still in DCache.

PR #1261, opened 2025-04-02
Fix Dw3110 initialization function failing on STM32 G474

Javask:
Corrects a bug in the DW3110 driver that causes the initialization function to fail on faster chips. On the F401 it initializes successfully because the device reaches IDLE_PLL in the time it takes to check it once. On G474 it always times out on first power up since the state that is checked against was not updated correctly.


salkinium (approved):
Yes, I understood some of those words ;-P
Thanks!

PR #1265, opened 2025-05-17
[ext] Update FreeRTOS to 202411.00

salkinium:
- [ ] Standard scheduler examples
- [ ] Fix FreeRTOS+TCP examples
- [ ] Tested in hardware


salkinium:
@mikewolfram Could you have a look at upgrading the FreeRTOS+TCP examples?

PR #1273, opened 2025-06-28
Fix cmake cmp0175 warnings on cmake 3.31 and higher

Tecnologic:
I use cmake 4.0 and gcc 14.2 on my project. the hole cmake build is based on the cmake example project.

What does thid fix?:
1. [CMake starting from 3.31](https://cmake.org/cmake/help/latest/policy/CMP0175.html) have the cmp0175 policy which deprecateds the USE_TERMINAL flag for custom output commands. To fix that i changed the Comands generated in ModmConfiguration.cmake to the approach cmake suggests. currently it works for the build but i have done very little testing on the commands if everything works as expected. 
2. My CMake root and the modm root are differnent folders. To make that possible i needed to change how project_source_dir is filled and that it is used for the linker script as well. Also the python path needed to change to the "CMAKE_CURRENT_SOURCE_DIR" because thats inside the generated modm folder.

### DIFF
--- a/tools/build_script_generator/cmake/resources/ModmConfiguration.cmake.in
+++ b/tools/build_script_generator/cmake/resources/ModmConfiguration.cmake.in
@@ -175,132 +175,140 @@ function(modm_targets_create project_name)
 %% if core.startswith("cortex-m")
   add_custom_command(TARGET ${project_name}
     POST_BUILD
-    COMMAND cmake -E env PYTHONPATH=${PROJECT_SOURCE_DIR}/modm ${Python3_EXECUTABLE} -m modm_tools.size ${project_name}.elf \"{{ memories }}\")
+    COMMAND cmake -E env PYTHONPATH=${CMAKE_CURRENT_SOURCE_DIR}/modm ${Python3_EXECUTABLE} -m modm_tools.size ${project_name}.elf \"{{ memories }}\")
 %% endif
 
-  add_custom_target(size DEPENDS ${project_name}.elf)
-  add_custom_command(TARGET size
+  add_custom_target(TARGET size ALL DEPENDS size.stamp)
+  add_custom_command(OUTPUT size.stamp
     USES_TERMINAL
-    COMMAND cmake -E env PYTHONPATH=modm ${Python3_EXECUTABLE} -m modm_tools.size ${PROJECT_BINARY_DIR}/${project_name}.elf \"{{ memories }}\"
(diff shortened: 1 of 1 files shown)


Tecnologic:
as expected most of the problems arise from folder structure.
my linker script location is "{project_source_dir}/link/linkerscript.ld" as opposed to "modm/link/linkerscript.ld". The variable {project_source_dir} seems unappropriate here.

Tecnologic:
To get away with the linkerscript on cortex m platforms i use <option name="modm:platform:cortex-m:linkerscript.override">link/linkerscript.ld</option> That fixes my compile issue in my project while keeping the examples working.

Tecnologic:
I'll skip the directory independance for this pull request and stick to the override option for the linkerskript.

salkinium:
So Point 1 makes sense to me, thanks for updating our CMake to the latest standard!

Point 2 is a bit unclear to me. Could you provide the folder structure, maybe using an simplified `tree` command? Why is the linkerscript location so special? Shouldn't other paths also break when moving files around? There's also a [`modm:platform:cortex-m:linkerscript.override` option](https://modm.io/reference/module/modm-platform-cortex-m/#linkerscriptoverride) in case you want to modify the existing linkerscript.

lbuild/modm has the option to relocate the output path using the `<outpath>...</outpath>` xml option:
https://github.com/modm-io/modm/blob/464c35f5336367d2a8bbedc7bd693582df007847/examples/nucleo_f103rb/blink/project.xml#L7
It's a fairly niche feature that was [introduced in this PR](https://github.com/modm-io/modm/pull/645/files). That was before the CMake generator existed, so not sure if that is properly supported.

Tecnologic:
I switched to using [modm:platform:cortex-m:linkerscript.override option](https://modm.io/reference/module/modm-platform-cortex-m/#linkerscriptoverride) option in my project so this contains no path releated stuff anymore accept from the cmake specific stuff.

as recomended in the other pr i squashed the commits and rebased on the current release. 

I hope this is fine for u.

salkinium (approved):
I was able to reproduce this behaviour with cmake v4.0.3 and your commit fixes it!

Thank you!

salkinium:
Could you fix your commit message to `[cmake] Fix CMP0175 warnings for CMake ≥v3.31`. The current squashed one is very non-descriptive. I unfortunately cannot push to your branch.

Tecnologic:
I struggled a bit, but i think this i gona work

PR #1295, opened 2025-11-06
Enhance documentation for Raspberry Pi debugging setup

daixtrose:
For people new to the topic, it is not evident that a combination of hardware and software is meant. This changeset adds links to both. 

Also, I added some words and rephrased a sentence that lacked a verb.

### DIFF
--- a/src/modm/board/rp_pico/module.lb
+++ b/src/modm/board/rp_pico/module.lb
@@ -32,16 +32,14 @@ to the mounted virtual disk.
 
 ### OpenOCD
 
-Another option is to use Picoprobe and custom version of OpenOCD provided by RPi.
+Another option is to use the [Raspberry Pi Debug Probe](https://www.raspberrypi.com/products/debug-probe/) together with the [Raspberry Pi Picoprobe software](https://github.com/raspberrypi/debugprobe) and a custom version of OpenOCD provided by RPi.
 
 You would need to build and program one of the boards with the Picoprobe firmware,
 configure and build RPi OpenOCD, and connect the boards via SWD.
-The wiring diagram and detailed description is available in the
+The wiring diagram and detailed description are available in the
 [Pico Documentation](https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf#Appendix%20A:%20Using%20Picoprobe).
 
-The how-to shows steps for building and installing the RPi OpenOCD into your
-system globally, however, if you still want to be able to work with other OpenOCD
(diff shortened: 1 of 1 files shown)


src/modm/board/rp_pico/module.lb
```diff
@@ -32,16 +32,14 @@ to the mounted virtual disk.
 
 ### OpenOCD
 
-Another option is to use Picoprobe and custom version of OpenOCD provided by RPi.
+Another option is to use the [Raspberry Pi Debug Probe](https://www.raspberrypi.com/products/debug-probe/) together with the [Raspberry Pi Picoprobe software](https://github.com/raspberrypi/debugprobe) and a custom version of OpenOCD provided by RPi.
```
rleh, line 35: As far as I understand, https://github.com/raspberrypi/debugprobe is just the firmware that runs on the Picoprobe hardware, right?
Then I don't think it makes sense to link to it here, because the user does not interact with this software directly.

rleh: Also, I believe that a "custom version" of OpenOCD is no longer necessary since the release of OpenOCD 0.12.0 in February 2023.

The following command works fine on my fedora installation with default openocd package:
```bash
openocd -f interface/cmsis-dap.cfg -f target/rp2040.cfg
```` 

see https://www.raspberrypi.com/documentation/microcontrollers/debug-probe.html#install-tools

File: src/modm/board/rp_pico/module.lb [hunk 2]
```diff
@@ -32,16 +32,14 @@ to the mounted virtual disk.
...
 You would need to build and program one of the boards with the Picoprobe firmware,
 configure and build RPi OpenOCD, and connect the boards via SWD.
-The wiring diagram and detailed description is available in the
+The wiring diagram and detailed description are available in the
 [Pico Documentation](https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf#Appendix%20A:%20Using%20Picoprobe).
 
-The how-to shows steps for building and installing the RPi OpenOCD into your
-system globally, however, if you still want to be able to work with other OpenOCD
-binaries, you should only install it locally:
+The how-to explains how to build and install the RPi OpenOCD globally, but if you want to keep using other OpenOCD binaries, you should install it only locally:
```
rleh, line 42: I would argue to remove the entire following how-to-build-openocd block (which was written in 2022 before the release of OpenOCD 0.12.0) as it is no longer necessary with OpenOCD 0.12.0.

PR #1296, opened 2025-11-10
[driver] driver for ST vl53l5, vl53l7, vl53l8

hshose:
## STs Ultra Light Driver for VL53L5 (I2C), VL53L7 (I2C), and VL53L8 (I2C, SPI) 8x8 Pixel Time-of-Flight Sensors

As the interaction with the sensor is pretty complicated (many undocumented registers and huge binary blobs), I use the ultra light drivers (ULD) for the sensors as provided by ST.
If you think this is a good way of integrating support for these sensors, I suggest to add these ULDs as a submodule under ext (see temp commit here for proposed structure). Does this work for y'all?
The VL53L5 driver exists undre STs namespace on github, the VL53LMZ driver for VL53L7 and VL53L8 only on STs website, I thus decided on STs website as datasorce for all drivers; I hope this is fine?
The `update.py` script in the future submodule automatically extracts the files, fixes some typos and bugs in their C-API, prefixes global functions with "VL53_", and unifies the use of the `VL53_Platform` struct among both drivers and their examples.

## modm transport
I added the modm transport SPI and I2C classes in the `ext/vl53` folder, is this the right place or should they go in the `src/modm/driver/position` folder?

## Examples
The VL53L8 sensor supports SPI and I2C, the VL53L7 sensor (same driver as the VL53L8 but wider field of view) supports only I2C, the VL53L5 supports only I2C.
Examples should cover these cases. 

## Todos:
- [x] Test VL53L8 SPI in hardware
- [x]  Test VL53L5 I2C in hardware
- [x] Test VL53L7/VL53L8 I2C in hardware

## Future plans:

Add support for the upcoming [VL53L9](https://www.st.com/en/imaging-and-photonics-solutions/vl53l9cx.html) when its released

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -52,3 +52,6 @@
 [submodule "ext/segger/rtt"]
 	path = ext/segger/rtt
 	url = https://github.com/modm-ext/segger-rtt.git
+[submodule "ext/vl53/vl53"]
+	path = ext/vl53/vl53
+	url = https://github.com/modm-ext/vl53-partial.git
--- a/examples/nucleo_h723zg/vl53/vl53l5cx_i2c/project.xml
+++ b/examples/nucleo_h723zg/vl53/vl53l5cx_i2c/project.xml
@@ -1,0 +1,12 @@
+<library>
+  <extends>modm:nucleo-h723zg</extends>
+  <options>
+    <option name="modm:build:build.path">../../../../build/nucleo_h723zg/vl53l5cx_i2c</option>
+    <option name="modm:platform:cortex-m:main_stack_size">8Ki</option>
+  </options>
+  <modules>
+    <module>modm:build:scons</module>
+    <module>modm:platform:i2c:1</module>
+    <module>modm:vl53:vl53l5cx</module>
+  </modules>
+</library>
--- a/ext/vl53/vl53.lb
+++ b/ext/vl53/vl53.lb
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025, Henrik Hose
+#
+# This file is part of the modm project.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -----------------------------------------------------------------------------
+
+class Vl53l5cx(Module):
+    def init(self, module):
+        module.name = "vl53l5cx"
+        module.description = "Driver for ST VL53L5CX Multi-Zone Time-of-Flight Sensor"
+
+    def prepare(self, module, options):
+        return True
+
+    def build(self, env):
+        env.outbasepath = "modm/ext"
+        env.copy("vl53/vl53l5cx/src")
+        env.copy("vl53/vl53l5cx/inc")
+        env.copy("vl53/vl53l5cx/LICENSE.txt")
+        env.collect(":build:path.include", "modm/ext/vl53/vl53l5cx/inc")
+
+class Vl53lmz(Module):
+    def init(self, module):
+        module.name = "vl53lmz"
+        module.description = "Driver for ST VL53L7 and VL53L8 Multi-Zone Time-of-Flight Sensors"
+
+    def prepare(self, module, options):
+        return True
+
+    def build(self, env):
+        env.outbasepath = "modm/ext"
+        env.copy("vl53/vl53lmz/src")
+        env.copy("vl53/vl53lmz/inc")
+        env.copy("vl53/vl53lmz/LICENSE.txt")
+        env.collect(":build:path.include", "modm/ext/vl53/vl53lmz/inc")
+
+def init(module):
+    module.name = ":vl53"
+    module.description = """\
+# ST's VL53 Time-of-Flight ULD Drivers
+
+Currently supported sensors:
+ - With VL53LMZ driver: VL53L7CX, VL53L7CH, VL53L8CX, VL53L8CH
+ - VL53L5CX driver
+
+"""
+
+def prepare(module, options):
+    module.add_submodule(Vl53l5cx())
+    module.add_submodule(Vl53lmz())
+    module.depends(
+        ":architecture:spi.device",
+        ":architecture:i2c.device",
+        ":architecture:fiber",
+        ":driver:i2c.eeprom")
+    return True
+
+def build(env):
+    env.outbasepath = "modm/src/modm/driver/vl53"
+    env.copy("vl53.cpp")
+    env.copy("vl53_transport.hpp")
+    env.copy("vl53_transport_impl.hpp")
+
+    env.outbasepath = "modm/ext/vl53"
+    env.collect(":build:path.include", "modm/ext/vl53")
+    env.copy("vl53_platform.h")
--- a/ext/vl53/vl53_platform.h
+++ b/ext/vl53/vl53_platform.h
@@ -1,0 +1,223 @@
+/**
+ *
+ * Copyright (c) 2021 STMicroelectronics.
+ * All rights reserved.
+ *
+ *
+ * This software component is provided to you as part of a software package and
+ * applicable license terms are in the  Package_license file. If you received this
+ * software component outside of a package or without applicable license terms,
+ * the terms of the BSD-3-Clause license shall apply.
+ * You may obtain a copy of the BSD-3-Clause at:
+ * https://opensource.org/licenses/BSD-3-Clause
+ *
+ ******************************************************************************
+ */
+
+#ifndef VL53_PLATFORM_H_
+#define VL53_PLATFORM_H_
+#pragma once
+
+#if __has_include("modm_config_vl53.h")
+#include "modm_config_vl53.h"
+#endif
+
+#include <stdint.h>
+#include <string.h>
+
+/*
+ * @brief The macro below is used to define the number of target per zone sent
+ * through I2C. This value can be changed by user, in order to tune I2C
+ * transaction, and also the total memory size (a lower number of target per
+ * zone means a lower RAM). The value must be between 1 and 4.
+ */
+
+#define VL53L5CX_NB_TARGET_PER_ZONE 1U
+
+/*
+ * @brief The macro below can be used to avoid data conversion into the driver.
+ * By default there is a conversion between firmware and user data. Using this macro
+ * allows to use the firmware format instead of user format. The firmware format allows
+ * an increased precision.
+ */
+
+// #define 	VL53L5CX_USE_RAW_FORMAT
+
+/*
+ * @brief All macro below are used to configure the sensor output. User can
+ * define some macros if he wants to disable selected output, in order to reduce
+ * I2C access.
+ */
+
+// #define VL53L5CX_DISABLE_AMBIENT_PER_SPAD
+// #define VL53L5CX_DISABLE_NB_SPADS_ENABLED
+// #define VL53L5CX_DISABLE_NB_TARGET_DETECTED
+// #define VL53L5CX_DISABLE_SIGNAL_PER_SPAD
+// #define VL53L5CX_DISABLE_RANGE_SIGMA_MM
+// #define VL53L5CX_DISABLE_DISTANCE_MM
+// #define VL53L5CX_DISABLE_REFLECTANCE_PERCENT
+// #define VL53L5CX_DISABLE_TARGET_STATUS
+// #define VL53L5CX_DISABLE_MOTION_INDICATOR
+
+/*
+ * @brief The macro below is used to define the number of target per zone sent
+ * through I2C. This value can be changed by user, in order to tune I2C
+ * transaction, and also the total memory size (a lower number of target per
+ * zone means a lower RAM usage).
+ */
+
+#define VL53LMZ_NB_TARGET_PER_ZONE 1
+
+/*
+ * @brief All macro below are used to configure the sensor output. User can
+ * define some macros if he wants to disable selected output, in order to reduce
+ * I2C access.
+ */
+
+// #define VL53LMZ_DISABLE_AMBIENT_PER_SPAD
+// #define VL53LMZ_DISABLE_NB_SPADS_ENABLED
+// #define VL53LMZ_DISABLE_NB_TARGET_DETECTED
+// #define VL53LMZ_DISABLE_SIGNAL_PER_SPAD
+// #define VL53LMZ_DISABLE_RANGE_SIGMA_MM
+// #define VL53LMZ_DISABLE_DISTANCE_MM
+// #define VL53LMZ_DISABLE_REFLECTANCE_PERCENT
+// #define VL53LMZ_DISABLE_TARGET_STATUS
+// #define VL53LMZ_DISABLE_SCI
+
+/**
+ * @brief Structure VL53_Platform needs to be filled by the customer,
+ * depending on his platform. At least, it contains the VL53 I2C address.
+ * Some additional fields can be added, as descriptors, or platform
+ * dependencies. Anything added into this structure is visible into the platform
+ * layer.
+ */
+
+typedef struct
+{
+	/* To be filled with customer's platform. At least an I2C address/descriptor
+	 * needs to be added */
+	/* Example for most standard platform : I2C address of sensor */
+	uint16_t address;
+
+	void *transport;
+
+} VL53_Platform;
+
+/*
+ * @brief The macro below is used to define the number of target per zone sent
+ * through I2C. This value can be changed by user, in order to tune I2C
+ * transaction, and also the total memory size (a lower number of target per
+ * zone means a lower RAM). The value must be between 1 and 4.
+ */
+
+#define VL53_NB_TARGET_PER_ZONE 1U
+
+/*
+ * @brief The macro below can be used to avoid data conversion into the driver.
+ * By default there is a conversion between firmware and user data. Using this macro
+ * allows to use the firmware format instead of user format. The firmware format allows
+ * an increased precision.
+ */
+
+// #define 	VL53_USE_RAW_FORMAT
+
+/*
+ * @brief All macro below are used to configure the sensor output. User can
+ * define some macros if he wants to disable selected output, in order to reduce
+ * I2C access.
+ */
+
+// #define VL53_DISABLE_AMBIENT_PER_SPAD
+// #define VL53_DISABLE_NB_SPADS_ENABLED
+// #define VL53_DISABLE_NB_TARGET_DETECTED
+// #define VL53_DISABLE_SIGNAL_PER_SPAD
+// #define VL53_DISABLE_RANGE_SIGMA_MM
+// #define VL53_DISABLE_DISTANCE_MM
+// #define VL53_DISABLE_REFLECTANCE_PERCENT
+// #define VL53_DISABLE_TARGET_STATUS
+// #define VL53_DISABLE_MOTION_INDICATOR
+
+/**
+ * @param (VL53_Platform*) p_platform : Pointer of VL53 platform
+ * structure.
+ * @param (uint16_t) Address : I2C location of value to read.
+ * @param (uint8_t) *p_values : Pointer of value to read.
+ * @return (uint8_t) status : 0 if OK
+ */
+
+uint8_t
+VL53_RdByte(VL53_Platform *p_platform, uint16_t RegisterAdress, uint8_t *p_value);
+
+/**
+ * @brief Mandatory function used to write one single byte.
+ * @param (VL53_Platform*) p_platform : Pointer of VL53 platform
+ * structure.
+ * @param (uint16_t) Address : I2C location of value to read.
+ * @param (uint8_t) value : Pointer of value to write.
+ * @return (uint8_t) status : 0 if OK
+ */
+
+uint8_t
+VL53_WrByte(VL53_Platform *p_platform, uint16_t RegisterAdress, uint8_t value);
+
+/**
+ * @brief Mandatory function used to read multiples bytes.
+ * @param (VL53_Platform*) p_platform : Pointer of VL53 platform
+ * structure.
+ * @param (uint16_t) Address : I2C location of values to read.
+ * @param (uint8_t) *p_values : Buffer of bytes to read.
+ * @param (uint32_t) size : Size of *p_values buffer.
+ * @return (uint8_t) status : 0 if OK
+ */
+
+uint8_t
+VL53_RdMulti(VL53_Platform *p_platform, uint16_t RegisterAdress, uint8_t *p_values, uint32_t size);
+
+/**
+ * @brief Mandatory function used to write multiples bytes.
+ * @param (VL53_Platform*) p_platform : Pointer of VL53 platform
+ * structure.
+ * @param (uint16_t) Address : I2C location of values to write.
+ * @param (uint8_t) *p_values : Buffer of bytes to write.
+ * @param (uint32_t) size : Size of *p_values buffer.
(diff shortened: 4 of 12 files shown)


ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,140 @@
...
+namespace modm
+{
+
+class Vl53TransportBase
+{
+public:
+	virtual ~Vl53TransportBase() = default;
+
+	virtual uint8_t
+	writeMulti(uint16_t reg, uint8_t *values, uint32_t size) = 0;
```
chris-durand:
```suggestion
	writeMulti(uint16_t reg, const uint8_t *values, uint32_t size) = 0;
```

It doesn't hurt to make it `const`-correct even if the calling code doesn't ask for it.

ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,140 @@
...
+	virtual uint8_t
```
chris-durand, line 31: I presume the return values should correspond to the C API? Your I2C implementation will return 1 on success, but the SPI transport always returns 0. That doesn't match ;)

ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,140 @@
...
+	Vl53SpiTransport &
+	operator=(const Vl53SpiTransport &) = delete;
+
+	uint8_t
+	writeMulti(uint16_t register_address, uint8_t *p_values, uint32_t size) override;
+
+	uint8_t
+	readMulti(uint16_t register_address, uint8_t *p_values, uint32_t size) override;
+
+	inline uint8_t
```
chris-durand:
```suggestion
	uint8_t
```
No explicit `inline` required here. It's implicitly `inline` because it's a definition inside the class declaration body.

ext/vl53/vl53_transport_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+	return status;
+}
+
+template<typename SpiMaster, typename Cs, typename Lpn, size_t VL53LMZ_COMMS_CHUNK_SIZE>
+uint8_t
+Vl53SpiTransport<SpiMaster, Cs, Lpn, VL53LMZ_COMMS_CHUNK_SIZE>::resetSensor()
+{
+	Lpn::reset();
+	modm::delay_ms(100);
```
chris-durand: This should use some fiber wait mechanism instead of a blocking delay.

ext/vl53/vl53_transport_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+template<typename SpiMaster, typename Cs, typename Lpn, size_t VL53LMZ_COMMS_CHUNK_SIZE>
+uint8_t
+Vl53SpiTransport<SpiMaster, Cs, Lpn, VL53LMZ_COMMS_CHUNK_SIZE>::readMulti(uint16_t RegisterAddress,
+																		  uint8_t *p_values,
+																		  uint32_t size)
+{
+	uint8_t status = 0;
+	uint32_t position = 0;
+	uint32_t data_size = 0;
+	uint16_t temp;
```
chris-durand: I'm guessing this was adapted from some C89 code? `position` can go into to the `for` statement and `temp` can be defined in the body for better readability.

chris-durand (commented):
Very nice! I don't have the time for a full review because I'll be on a dive boat in the Andaman sea in 2 hours ;)

Putting ST's code into ext is the right way to go. Getting it from the website is also fine because it's BSD licensed and you are providing the automated update script. @salkinium or @rleh can probably help with making a repo in modm-ext.

ext/vl53/vl53_transport_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+		modm::this_fiber::poll([&] { return this->acquireMaster(); });
+		Cs::reset();
+		SpiMaster::transfer(&tx_buffer[0], &rx_buffer[0], data_size + 2);
+		// SpiMaster::transfer(nullptr, p_values + position, data_size);
+		if (this->releaseMaster()) { Cs::set(); }
+
+		std::memcpy(p_values + position, &rx_buffer[2], data_size);
+	}
+
+	return status;
```
chris-durand: `status` is always 0

ext/vl53/vl53_transport_impl.hpp
hshose: Yes. I fixed it as you suggested in both SPI functions.

ext/vl53/vl53_transport_impl.hpp
hshose: Not sure if SPI has a status that is returned. I now just always return 0 directly and avoid the unused status variable.

ext/vl53/vl53_transport_impl.hpp
hshose: fixed.

ext/vl53/vl53_transport.hpp
hshose: fixed

ext/vl53/vl53_transport.hpp
hshose: whuuuuupsiii :-D You're absolutely right!

ext/vl53/vl53_transport.hpp
hshose: fixed

hshose:
All examples work in HW now. If you set up the ext repo, I can add the submodule and move the ST files there...

rleh:
I created https://github.com/modm-ext/vl53-partial and granted you access.

hshose:
Adding submodule done.

.gitmodules
```diff
@@ -52,3 +52,6 @@
 [submodule "ext/segger/rtt"]
 	path = ext/segger/rtt
 	url = git@github.com:modm-ext/segger-rtt.git
+[submodule "ext/vl53/vl53"]
+	path = ext/vl53/vl53
+	url = git@github.com:modm-ext/vl53-partial.git
```
salkinium: Could you change this URL to the HTTPS method? Otherwise it will fail for people without a GitHub SSH key setup. See https://github.com/modm-io/modm/pull/1297.

examples/nucleo_h723zg/vl53/vl53l5cx_i2c/project.xml
```diff
@@ -0,0 +1,15 @@
+<library>
+  <extends>modm:nucleo-h723zg</extends>
+  <options>
+    <option name="modm:build:build.path">../../../../build/nucleo_h723zg/vl53l5cx_i2c</option>
+    <option name="modm:vl53:vl53l5cx:examples">yes</option>
+    <option name="modm:platform:cortex-m:main_stack_size">64Ki</option>
```
salkinium: WAT? is this a left over of the development, or do the examples really use 64kB of stack?

ext/vl53/vl53.hpp
```diff
@@ -0,0 +1,73 @@
...
+		tmp = (buffer[i] << 24) | (buffer[i + 1] << 16) | (buffer[i + 2] << 8) | (buffer[i + 3]);
+
+		memcpy(&(buffer[i]), &tmp, 4);
+	}
+}
+
+uint8_t
+VL53_WaitMs(VL53_Platform *p_platform, uint32_t TimeMs)
+{
+	modm::delay_ms(TimeMs);
```
salkinium:
```suggestion
	modm::this_fiber::sleep_for(std::chrono::milliseconds(TimeMs));
```

ext/vl53/vl53.lb
```diff
@@ -0,0 +1,86 @@
...
+        )
+        return True
+
+    def build(self, env):
+        env.outbasepath = "modm/ext"
+        env.copy("vl53/vl53l5cx/src")
+        env.copy("vl53/vl53l5cx/inc")
+        env.copy("vl53/vl53l5cx/LICENSE.txt")
+        if env[":vl53:vl53l5cx:examples"]:
+            env.copy("vl53/vl53l5cx/examples")
```
salkinium: I weakly feel like this shouldn't be an lbuild option.
At least it's not consistent the the CMSIS-DSP examples, which are [instead included from the source path](https://github.com/modm-io/modm/blob/develop/examples/nucleo_f429zi/cmsis_dsp/signal_converge/main.cpp).

ext/vl53/vl53.lb
```diff
@@ -0,0 +1,86 @@
...
+
+def prepare(module, options):
+    module.add_submodule(Vl53l5cx())
+    module.add_submodule(Vl53lmz())
+    module.depends(":architecture:spi.device")
+    module.depends(":architecture:i2c.device")
+    return True
+
+def build(env):
+    env.outbasepath = "modm/ext/vl53"
```
salkinium:
```suggestion
    env.outbasepath = "modm/src/modm/driver/vl53"
```
To separate the modm part from the ext part.

ext/vl53/vl53_platform.h
```diff
@@ -0,0 +1,219 @@
...
+ * You may obtain a copy of the BSD-3-Clause at:
+ * https://opensource.org/licenses/BSD-3-Clause
+ *
+ ******************************************************************************
+ */
+
+#ifndef VL53_PLATFORM_H_
+#define VL53_PLATFORM_H_
+#pragma once
+
```
salkinium, line 20:
```suggestion
#if __has_include("modm_config_vl53.h")
#include "modm_config_vl53.h"
#endif
```
To allow user configuration without modifying this library file.

salkinium (changes requested):
Thanks, just some minor improvements needed.

.gitmodules
hshose: sure, fixed this.

ext/vl53/vl53.hpp
hshose: fixed

hshose: fixed

ext/vl53/vl53.lb
hshose: done

ext/vl53/vl53_platform.h
hshose: done

ext/vl53/vl53.lb
hshose: Really don't have any strong opinion on this.
I thought the point of lbuild was to copy the files you need from the top-level modm directory to your local project directory and then using macros in the examples to access ext files seems to contradict that paradigm.

Anyways, I changed it to be consistent with CMSIS DSP.

examples/nucleo_h723zg/vl53/vl53l5cx_i2c/project.xml
hshose: The I2C version needs the 64kB (or at least it stops working if I set 32kB stack, not really sure how to debug stack usage in modm.
The SPI version needs only 16kB (depending on the size of the buffer for the SPI read/write).

The firmware needs a bunch of large buffers for the transfers and also has pretty large firmware blobs (`cost uint8_t[LARGE]`), not sure where this breaks with smaller stack.

hshose:
Thank you @salkinium for the detailed review, I tried to address most of the issues, only the stack and example include questions remain. I left the stack as is because the I2C examples seem to need this, I changed the SPI examples stack to 16kB.
Changed the examples as separate commit in case you don't want to have this.

examples/nucleo_h723zg/vl53/vl53l5cx_i2c/project.xml
salkinium: It's because everything in the example functions is allocated on the stack.

- sizeof(VL53L5CX_ResultsData) = 1360
- sizeof(VL53L5CX_Configuration) = 2744

But that's just 4KB together, where does the other 60kB come from?
I'm worried there's some kind of recursion inside the drivers that the port is causing.
Did you test the driver outside of the examples yet?

salkinium: Ah, found it. You allocate the I2C write buffer on the stack here:
```cpp
template<typename I2cMaster, typename Lpn>
uint8_t
Vl53I2cTransport<I2cMaster, Lpn>::writeMulti(uint16_t RegisterAddress, const uint8_t *p_values,
											 uint32_t size)
{
	uint8_t data_write[size + 2];
	data_write[0] = (RegisterAddress >> 8) & 0xFF;
	data_write[1] = RegisterAddress & 0xFF;
```
and this gets called in the `vl53l5cx_init` function with size = 0x8000, which is 32kB.
```cpp
	status |= VL53_WrMulti(&(p_dev->platform),0,
		(uint8_t*)&VL53L5CX_FIRMWARE[0x8000],0x8000);
```

You should instead inherit privately from [the `modm::I2cEeprom` class](https://github.com/modm-io/modm/blob/develop/src/modm/driver/storage/i2c_eeprom.hpp#L70), which allows for copy-less prefixing of a 2-byte address.

ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+	Vl53I2cTransport(const Vl53I2cTransport &) = delete;
+
+	Vl53I2cTransport &
+	operator=(const Vl53I2cTransport &) = delete;
+
+	inline void
+	SetAddress(const uint8_t address)
+	{
+		this->address_ = address;
```
salkinium: Isn't this functionality already [part of `modm::I2cDevice`](https://github.com/modm-io/modm/blob/develop/src/modm/architecture/interface/i2c_device.hpp#L58)?

ext/vl53/vl53_transport_impl.hpp
```diff
@@ -0,0 +1,182 @@
...
+		tx_buffer[0] = SPI_WRITE_MASK(temp) >> 8;
+		tx_buffer[1] = SPI_WRITE_MASK(temp) & 0xFF;
+
+		for (uint32_t i = 0; i < data_size; i++) { tx_buffer[i + 2] = p_values[position + i]; }
+
+		data_size += 2;
+
+		modm::this_fiber::poll([&] { return this->acquireMaster(); });
+		Cs::reset();
+		SpiMaster::transfer(&tx_buffer[0], nullptr, data_size);
```
salkinium: Why not do it in two transfers?
```cpp
uint8_t addr_buffer[2];
addr_buffer[0] = SPI_WRITE_MASK(temp) >> 8;
addr_buffer[1] = SPI_WRITE_MASK(temp) & 0xFF;
SpiMaster::transfer(addr_buffer, nullptr, sizeof(addr_buffer));
SpiMaster::transfer(p_values, nullptr, size);
```
that saves the memcpy and the scratch buffer

ext/vl53/vl53_transport.hpp
hshose: whups, okay yes!

examples/nucleo_h723zg/vl53/vl53l5cx_i2c/project.xml
hshose: This is nice! Didn't know to look in the storage drivers for this. Gets the example Stack down to like 12kB.

ext/vl53/vl53_transport_impl.hpp
hshose: I used to do this, and it works with the `SpiMaster2`.
However, with the `SpiMaster2_Dma` it doesn't work and I don't really know why. Any ideas?

When I have the separate TX Buffer and memcopy the write content into that Buffer, it also works with the `SpiMaster2_Dma`.

ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+	static uint16_t
+	SPI_READ_MASK(uint16_t x) noexcept
+	{
+		return static_cast<uint16_t>(x & static_cast<uint16_t>(~0x8000));
+	}
+};
+
+template<typename I2cMaster, typename Lpn>
+class Vl53I2cTransport : public modm::I2cDevice<I2cMaster>, public Vl53TransportBase, private I2cEeprom<I2cMaster, 2>
```
salkinium:
```suggestion
class Vl53I2cTransport : public modm::I2cEeprom<I2cMaster, 2>, public Vl53TransportBase
```

ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,138 @@
...
+{
+public:
+	Vl53I2cTransport(const uint8_t address) : I2cDevice<I2cMaster>{address}, I2cEeprom<I2cMaster,2>{address} {};
```
salkinium:
```suggestion
	Vl53I2cTransport(const uint8_t address) : I2cEeprom<I2cMaster,2>{address} {};
```

ext/vl53/vl53_transport.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+	Vl53I2cTransport &
+	operator=(const Vl53I2cTransport &) = delete;
+
+	void
+	setAddress(const uint8_t address)
+	{
+		I2cEeprom<I2cMaster,2>::transaction.setAddress(address);
+		I2cDevice<I2cMaster>::transaction.setAddress(address);
+	};
```
salkinium: this function is already implemented by I2cDevice

ext/vl53/vl53_transport_impl.hpp
```diff
@@ -0,0 +1,174 @@
...
+uint8_t
+Vl53I2cTransport<I2cMaster, Lpn>::readByte(uint16_t RegisterAddress, uint8_t *p_value)
+{
+	uint8_t data_write[2];
+	uint8_t data_read[1];
+
+	data_write[0] = (RegisterAddress >> 8) & 0xFF;
+	data_write[1] = RegisterAddress & 0xFF;
+
+	const auto status = I2cDevice<I2cMaster>::writeRead(&data_write[0], 2, &data_read[0], 1);
```
salkinium:
```suggestion
	const auto status = I2cDevice<I2cMaster>::writeRead(&data_write[0], 2, p_value, 1);
```

ext/vl53/vl53_transport_impl.hpp
salkinium: Not sure, but the DMA buffers need to be 4B aligned. If you make the `addr_buffer[0]` a member it should be 4B aligned.

hshose: Making addr_buffer a member doesn't help. How would i know that `p_values` is aligned properly?

chris-durand: > Not sure, but the DMA buffers need to be 4B aligned. If you make the `addr_buffer[0]` a member it should be 4B aligned.

Is that really required? The SPI DMA only performs byte accesses and it should even support unaligned access on normal memory. The only references on DMA alignment requirements I could find were on device and strongly-ordered memory which need to be properly aligned.

However, when you use cacheable memory DMA buffers need to be aligned to the cache line size for cache maintenance operations to sensibly work. modm doesn't support that anyway. @hshose Just to make sure, you didn't manually enable the data cache, right?

ext/vl53/vl53_transport.hpp
hshose: wasn't accessible before with the private I2CEEprom... but is now with the other changes.

ext/vl53/vl53_transport_impl.hpp
hshose: I wouldn't even know how to turn this on :smile: Also making the `addr_buffer` a class member doesn't help.

salkinium: I pushed a workaround to a common cache bug on Cortex-M7. It probably won't fix the DMA issue, but it's something we need to fix in modm anyways.

salkinium:
Yes, the macOS CI is super slow (~2h), I'm trying to fix it over at #1301.

ext/vl53/vl53_transport_impl.hpp
chris-durand: Erratum 1259864 only applies if a memory region is configured in write-through mode. In the default memory map all internal cacheable RAM and almost all external RAM is write-back write-allocate. At least in the 3 H7 manuals I had on my laptop. Only external RAM connected via QUAD/OCTOSPI would be write-through by default. We don't support that, so we should be fine.

Screenshot_20251121_173045

salkinium: Yup I know, but somehow this fixed several issues I was seeing on a H743/r1p1. Maybe this helps here. Otherwise it's a bug in the SPI DMA implementation.

chris-durand: Would be really interesting if it made any difference here because the data cache should be off and the examples are on an H723/r1p2 which shouldn't have this issue.

chris-durand: @hshose Could you check with a scope or logic analyzer what goes wrong if you split the transfers with DMA?

salkinium: > are on an H723/r1p2

Oh, then my "fix" is indeed useless. 😞

hshose: I am not entirely sure what if or what changed but now that I'm looking with a logic analyzer it also works with the DMA :facepalm: I should go home and do something else with my life :crying_cat_face:

hshose: Maybe it was the alignement issue and I forgot to invoke lbuild on when I tested :see_no_evil: 
Oh well.

hshose: image

Also the testing setup is a bit of a mess.

ext/vl53/vl53_transport_impl.hpp
```diff
@@ -0,0 +1,164 @@
...
+		const uint16_t temp = RegisterAddress + position;
+
+		addr_buffer[0] = SPI_WRITE_MASK(temp) >> 8;
+		addr_buffer[1] = SPI_WRITE_MASK(temp) & 0xFF;
+
+		modm::this_fiber::poll([&] { return this->acquireMaster(); });
+		Cs::reset();
+		// SpiMaster::transfer(&tx_buffer[0], nullptr, data_size);
+		SpiMaster::transfer(addr_buffer, nullptr, sizeof(addr_buffer));
+		SpiMaster::transfer(&p_values[position], nullptr, data_size);
```
salkinium: Dumb question: Why are we chunking this here? Is there a limit on the SPI transfer length? Cos the I2C just dumps 32kB of data into the sensor in one transfer, if I understand this correctly.

salkinium (commented):
I like this, thank you! I will have a look at the submodule until Monday and check if I can get it to update automatically by going past the login wall like we do for the CubeMX database.

ext/vl53/vl53_transport_impl.hpp
hshose: The ST ULD implementation does this and due to the lack of documentation for the sensors besides the provided drivers, I just did the same.
I can test tomorrow if we can just write the entire firmware blob at once.

salkinium: Ping? I didn't find a good way to download the libraries quickly. But I can fix it later, it shouldn't block this PR.

hshose:
```
64 bytes from hshose (1234:5678:7654:321::2003): icmp_seq=1 ttl=117 time=54345123.19 ms
```
Sorry, my dog ate my homework :poodle: 

It also works in hardware without the chunking (in both read and write).

salkinium:
I squashed your commits into two and then added two fixup commits with my changes. Mostly just moving files into the right place and reducing stack usage to 8kB. Could you test that again please and then squash the fixups if it works?

salkinium (approved):

hshose:
8k stack is too little for the VL53L7 or L8. it works with 12k.

For the VL53L5 8K is enough.

I have updated the examples accordingly

PR #1299, opened 2025-11-15
[RTT] printf include path wrong on cmake build

Tecnologic:
Hi All,

my cmake build with rtt failed due to the printf.h included in rtt_modm_port.cpp. The include was given with a full path
`#include <modm/ext/printf/printf.h>`. The cmake file generated from modm does not include the base folder, but the ext folder is included. So i changed the path accordingly. I expect that scons and other builds include the ext folder as well. So it should not be a big deal.


salkinium (approved):
Yes, this is the correct fix. It probably only accidentally worked before.

PR #1003, opened 2023-04-16
[WIP] [stm32] F469 disco hardware rev b08 display driver nt35510

hshose:
First attempt to fix #1002 by using the ST driver for nt35510.
This is for the STM32F469 example in `examples/stm32f469_discovery/display`

- [ ] make driver work (hardware)
- [x] make nt35510/otm8009a driver configurable with lbuild board revision

PR #1319, opened 2026-01-17
STM32F469-Disco: Add support for NT35510 display driver on newer revisions

Javask:
Add support for nt35510 driver for revision b-08.
Based on https://github.com/modm-io/modm/pull/1003.


salkinium (approved):
Thank you! (I only squashed it)

PR #1324, opened 2026-01-24
[rcc] Move STM32F1 RCC to new API

salkinium:
- [x] Extract RCC driver to separate file and significantly improve the API
- [x] Tested in Hardware on Blue Pill F103 and Nucleo-F103
- [x] Compile-all job passed previously


salkinium:
Windows CI still fails to download the GCCs, I've decided not to care for now.

PR #1331, opened 2026-02-11
[fix] allow cmsis dsp on hosted linux

hshose:
It can be annoyingly hard to develop numerics directly on bare-metal microcontroller.
The CMSIS DSP compiles just fine on the `hosted-linux` target.

Is there any reason to not allow for this target in the lbuild module?

### DIFF
--- a/examples/linux/cmsis_dsp/bayes/main.cpp
+++ b/examples/linux/cmsis_dsp/bayes/main.cpp
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2019, Niklas Hauser
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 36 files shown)


salkinium:
There was a previous attempt at this: https://github.com/modm-io/modm/pull/1203

I think it's fine, but modm support for hosted is pretty minimal and maybe using the official Python package is a better experience overall?

salkinium:
Please add a small example for hoested linux so that thies doesn't break in the future

hshose:
Sorry for the delay :-D I've added the same examples that run on the STM32F4.

I also needed to remove the CMake parts of the examples because the CI wouldn't run otherwise. Do we need something CMake or is just scons okay?

hshose:
update: ci seems to check out now :-)

salkinium (approved):
Thanks!

PR #1337, opened 2026-02-22
fix: move module options upwards towards parent in docs

### LINKED ISSUE #804: Incorrect module option availability generated for can buffers in documentation
At https://modm.io/reference/module/modm-platform-can-stm32-f0-f1-f2-f3-f4-f7-l4/, it is stated for the Rx and Tx buffer size option that

> This option is only available for stm32{f0,f1,f3}.

Image showing the website documentation with the aforementioned error

But lbuild discover is giving the following output:

```   
    │   ├── Module(modm:platform:can)   Controller Area Network (CAN)
    │   │   ├── Module(modm:platform:can:1)   Instance 1
    │   │   │   ├── Option(buffer.rx) = 32 in [1 .. 32 .. 65534]   
    │   │   │   ╰── Option(buffer.tx) = 32 in [1 .. 32 .. 65534]   
```

for my l4, f4, and f7 targets that have a can peripheral.

Also, @rleh showed me the following link where the option is added to all stm32 targets with a can peripheral: https://github.com/modm-io/modm/blob/develop/src/modm/platform/can/stm32/module.lb#L18-L30

So this seems to be an issue with the underlying magic:tm: that generates the documentation.

Unfortunately, the inner workings of the documentation generation are quite obscure to me, I'd guess the error origin is maybe somewhere around here: https://github.com/modm-io/modm/blob/develop/tools/scripts/generate_module_docs.py#L91-L123 

If you think that this can be fixed by someone who knows python but is not at all acquainted with the docs generation I'd give fixing it a try, otherwise, I would be pretty grateful if someone else could take a look at this.

joelsa:
Fixes https://github.com/modm-io/modm/issues/804

grafik

### DIFF
--- a/docs/module.md.in
+++ b/docs/module.md.in
@@ -33,6 +33,33 @@ Input Dependency: `{{dependency}}`{% if targets | length %} *{{targets}}*{% endi
 %% endfor
 %% endif
 
+%% if module.instance_options | length
+## Instance Options
+These options are available for each instance of this module.
+
+%% for o in module.instance_options
+#### {{o.name | replace(module.name + ":", "")}}
+
+{{o.description}}
+
+%% if o.targets | length
+*This option is only available for {{o.targets}}.*
+
+%% endif
+
+%% for value, targets in o.default.items() | sort
+Default: `{{value}}`{% if targets | length %} *{{targets}}*{% endif %}  
+%% endfor
+%% for input, targets in o.inputs.items() | sort
+Inputs: `[{{input}}]`{% if targets | length %} *{{targets}}*{% endif %}  
+%% endfor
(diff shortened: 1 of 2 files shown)


salkinium:
Thank you for fixing this, I completely forgot about this issue!

Screenshot 2026-02-22 at 17 31 28

I think the only thing missing is marking the instance options as such? Maybe something like this?

Screenshot 2026-02-22 at 17 38 42

joelsa:
> I think the only thing missing is marking the instance options as such? Maybe something like this?

Yes, good idea!

joelsa:
Implemented in 359de49

grafik

salkinium (approved):
Thank you!

PR #1339, opened 2026-03-03
[driver] bmp581 pressure sensor

joelsa:
Driver for the Bosch Sensortec BMP581 pressure and temperature sensor.

Example code outputs:
```
Temperature: 40.796 C
Pressure: 96074.19 Pa (960.74 hPa)
Altitude: 446.6 m (approx)
```

I am actually on 530 m, nearest weather station reports:

grafik

963 hPa at 515 m corresponds to 961 hPa on 530 m, which is basically what the sensor is reading. Absolute pressure accuracy is specified with ±0.5 hPa at 25 °C and ±1.0 hPa (max) over 0–65 °C. Unfortunately my PCB actually gets 40 °C hot, so that reading is correct as well.

Tested the sensor at different output data rates and in both polling and interrupt modes. Could test only in I2C mode and only on STM32 because that's how my board is wired up.

### DIFF
--- a/examples/nucleo_h723zg/bmp581/main.cpp
+++ b/examples/nucleo_h723zg/bmp581/main.cpp
@@ -35,45 +35,51 @@ bool
 initializeBaro()
 {
 	// Initialize the sensor
-	if (!baro.initialize()) {
+	if (!baro.initialize())
+	{
 		MODM_LOG_ERROR << "BMP581 initialization failed!\n";
 		return false;
 	}
 
 	// Configure oversampling: 4x for both pressure and temperature
-	if (!baro.setOversampling(Baro::Osr::X4, Baro::Osr::X4, true)) {
+	if (!baro.setOversampling(Baro::Osr::X4, Baro::Osr::X4, true))
+	{
 		MODM_LOG_ERROR << "Failed to set oversampling!\n";
 		return false;
 	}
 
 	// Configure output data rate: 50 Hz
-	if (!baro.setOdr(Baro::Odr::Hz50)) {
+	if (!baro.setOdr(Baro::Odr::Hz50))
+	{
 		MODM_LOG_ERROR << "Failed to set ODR!\n";
 		return false;
 	}
 
 	// Configure IIR filter: coefficient 3 for both
-	if (!baro.setIirFilter(Baro::IirFilter::Coef3, Baro::IirFilter::Coef3)) {
+	if (!baro.setIirFilter(Baro::IirFilter::Coef3, Baro::IirFilter::Coef3))
+	{
 		MODM_LOG_ERROR << "Failed to set IIR filter!\n";
 		return false;
 	}
 
 	// Configure interrupt: active high, push-pull, pulsed mode
-	const auto intConfig = Baro::IntConfig::Enable |
-	                       Baro::IntConfig::Polarity;  // Active high
-	if (!baro.setIntConfig(intConfig)) {
+	const auto intConfig = Baro::IntConfig::Enable | Baro::IntConfig::Polarity;  // Active high
+	if (!baro.setIntConfig(intConfig))
+	{
 		MODM_LOG_ERROR << "Failed to set interrupt config!\n";
 		return false;
 	}
 
 	// Enable data ready interrupt
-	if (!baro.setIntSource(Baro::IntSource::DataReadyEnable)) {
+	if (!baro.setIntSource(Baro::IntSource::DataReadyEnable))
+	{
 		MODM_LOG_ERROR << "Failed to set interrupt source!\n";
 		return false;
 	}
 
 	// Set power mode to normal (continuous measurement)
-	if (!baro.setPowerMode(Baro::PowerMode::Normal)) {
+	if (!baro.setPowerMode(Baro::PowerMode::Normal))
+	{
 		MODM_LOG_ERROR << "Failed to set power mode!\n";
 		return false;
 	}
@@ -97,7 +103,8 @@ main()
 	BaroInt::setInput(BaroInt::InputType::PullDown);
 
 	// Initialize sensor with retries
-	while (!initializeBaro()) {
+	while (!initializeBaro())
+	{
 		LedRed::toggle();
 		MODM_LOG_ERROR << "Retrying initialization...\n";
 		modm::delay(250ms);
@@ -106,7 +113,8 @@ main()
 	MODM_LOG_INFO << "BMP581 initialized successfully!\n\n";
 
 	// Read chip ID for verification
-	if (const auto chipId = baro.readChipId(); chipId) {
+	if (const auto chipId = baro.readChipId(); chipId)
+	{
 		MODM_LOG_INFO.printf("Chip ID: 0x%02X (expected 0x50)\n\n", *chipId);
 	}
 
@@ -122,12 +130,11 @@ main()
 	uint32_t readErrorCount = 0;
 	while (true)
 	{
-		if (!dataReady) {
-			continue;
-		}
+		if (!dataReady) { continue; }
 		dataReady = false;
 
-		if (baro.readData(data)) {
+		if (baro.readData(data))
+		{
 			const float temperature = data.getTemperature();
 			const float pressure = data.getPressure();
 			const float pressureHpa = data.getPressureHpa();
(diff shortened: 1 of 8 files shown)


examples/nucleo_h723zg/bmp581/project.xml
```diff
@@ -0,0 +1,12 @@
+<library>
+  <extends>modm:nucleo-h723zg</extends>
+  <options>
+    <option name="modm:build:build.path">../../../../build/nucleo_h723zg/bmp581</option>
```
hshose: Should be
```
<option name="modm:build:build.path">../../../build/nucleo_h723zg/bmp581</option>
```

src/modm/driver/pressure/bmp581_impl.hpp
```diff
@@ -0,0 +1,340 @@
...
+{
+	waitForCommandGap();
+
+	const uint8_t mask = uint8_t(OdrConfig::Odr0) | uint8_t(OdrConfig::Odr1) |
+	                     uint8_t(OdrConfig::Odr2) | uint8_t(OdrConfig::Odr3) |
+	                     uint8_t(OdrConfig::Odr4);
+	const uint8_t value = static_cast<uint8_t>(odr) << 2;
+
+	const bool ok = updateRegister(Register::OdrConfig, mask, value);
+	timer_.restart(std::chrono::microseconds{2});
```
hshose: Probably can make this work without the timer_ member and use modm::this_fiber::sleep_for(5ms); instead?

src/modm/driver/pressure/bmp581_impl.hpp
```diff
@@ -0,0 +1,340 @@
...
+	// Temperature IIR at [2:0], Pressure IIR at [5:3]
+	const uint8_t value = (static_cast<uint8_t>(tempIir) << 0) |
+	                      (static_cast<uint8_t>(pressIir) << 3);
+
+	const bool iirOk = writeRegister(Register::DspIir, value);
+
+	// Restore original power mode
+	const bool restoreOk = writeRegister(Register::OdrConfig, *odrConfig);
+
+	timer_.restart(std::chrono::microseconds{2});
```
hshose: Probably can make this work without the timer_ member and use modm::this_fiber::sleep_for(5us); instead?

examples/nucleo_h723zg/bmp581/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+int
+main()
+{
+	Board::initialize();
+	Leds::setOutput();
+
+	// Initialize I2C2
+	I2c::connect<Scl::Scl, Sda::Sda>(I2c::PullUps::Internal);
+	I2c::initialize<Board::SystemClock, 400_kHz>();
```
salkinium, line 113: 400kHz with only internal pullups is… a choice. 😬

examples/nucleo_h723zg/bmp581/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+		{
+			const float temperature = data.getTemperature();
+			const float pressure = data.getPressure();
+			const float pressureHpa = data.getPressureHpa();
+
+			// Calculate approximate altitude (simplified barometric formula)
+			// Using standard sea level pressure of 1013.25 hPa
+			constexpr float seaLevelPressure = 1013.25f;
+			const float altitude =
+				44330.0f * (1.0f - powf(pressureHpa / seaLevelPressure, 0.1903f));
```
salkinium, line 161: Could this be a function of `data`? `data.altitude(float seaLevelhPa)` at least for convenience.

examples/nucleo_h723zg/bmp581/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+		LedYellow::toggle();
+	});
+
+	modm::bmp581::Data data;
+	uint32_t sampleCount = 0;
+	uint32_t readErrorCount = 0;
+	while (true)
+	{
+		if (!dataReady) { continue; }
+		dataReady = false;
```
salkinium, line 149: Technically you'd need a `std::atomic_bool` or `std::atomic_flag` here…

src/modm/driver/pressure/bmp581.hpp
```diff
@@ -0,0 +1,511 @@
...
+	{
+		/// Get temperature in degrees Celsius
+		float
+		getTemperature() const
+		{
+			int32_t raw = (static_cast<int32_t>(rawTemp[2]) << 16) |
+						  (static_cast<int32_t>(rawTemp[1]) << 8) |
+						  (static_cast<int32_t>(rawTemp[0]));
+			// Sign-extend 24-bit signed value to 32-bit
+			if (raw & 0x00800000) { raw |= 0xFF000000; }
```
salkinium: 👀

src/modm/driver/pressure/bmp581.lb
```diff
@@ -0,0 +1,52 @@
...
+[Datasheet](https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmp581-ds004.pdf)
+"""
+
+def prepare(module, options):
+    module.depends(
+        ":architecture:gpio",
+        ":architecture:register",
+        ":architecture:spi.device",
+        ":architecture:i2c.device",
+        ":processing:fiber",
```
salkinium:
```suggestion
        ":architecture:fiber",
```
To support not using fibers at all for very smol microcontrollers. (without `:platform:fiber` all fiber APIs are blocking).

src/modm/driver/pressure/bmp581_transport.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * BMP581 I2C Transport Layer
+ *
+ * Implements register read/write operations over I2C.
+ *
+ * @tparam I2cMaster I2C master peripheral
+ */
+template<typename I2cMaster>
+class Bmp581I2cTransport : public modm::I2cDevice<I2cMaster, 4>
```
salkinium:
```suggestion
class Bmp581I2cTransport : public modm::I2cDevice<I2cMaster>
```
nesting levels are deprecated since fibers.

src/modm/driver/pressure/bmp581_transport_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+namespace modm
+{
+
+// ----------------------------------------------------------------------------
+// I2C Transport Implementation
+// ----------------------------------------------------------------------------
+
+template<typename I2cMaster>
+Bmp581I2cTransport<I2cMaster>::Bmp581I2cTransport(uint8_t address)
+	: I2cDevice<I2cMaster, 4>(address)
```
salkinium:
```suggestion
	: I2cDevice<I2cMaster>(address)
```
Same here and the rest of the file

src/modm/driver/pressure/bmp581_transport_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+	buffer_[0] = reg;
+	return I2cDevice<I2cMaster, 4>::writeRead(&buffer_[0], 1, data, length);
+}
+
+template<typename I2cMaster>
+bool
+Bmp581I2cTransport<I2cMaster>::write(uint8_t reg, uint8_t data)
+{
+	buffer_[0] = reg;
+	buffer_[1] = data;
```
salkinium: Yes works fine, but since fibers you can use the stack normally again. Something like this is the modern equivalent :

```cpp
	uint8_t buffer[]{reg, data};
	return I2cDevice<I2cMaster>::write(buffer, sizeof(buffer));
```

salkinium (changes requested):
Thank you! Just some minor nitpicks.

examples/nucleo_h723zg/bmp581/main.cpp
joelsa: It's certainly a choice, but the only one I have (forgot the external ones).

src/modm/driver/pressure/bmp581.hpp
joelsa: This is actually funny, the datasheet (p. 56f) states that pressure is signed 24 bit as well. I have trouble imagining a negative pressure environment, and the datasheet specifies the operating range from 30 ... 125 kPa.

Of the implementations I found online, 3 out of 4 do the sign extend for pressure as well, and one doesn't do it.

Should I do it for pressure as well? Is there a better way to do it?

salkinium: Maybe something like this? Not sure if you can shift a int32_t into the sign bit, maybe you need for first cast it all to uint32_t, then bit_cast to int32_t?
```cpp
			int32_t raw = (static_cast<int32_t>(rawTemp[2]) << 24) |
						  (static_cast<int32_t>(rawTemp[1]) << 16) |
						  (static_cast<int32_t>(rawTemp[0]) << 8);
			return static_cast<float>(raw) / (65536.0f * 256);
```
But yeah it's probably all academic, I'm also not sure if super empty vacuum exists.

joelsa: > Not sure if you can shift a int32_t into the sign bit

Found that:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1817.htm

> If you left-shift a signed number so that the sign bit is affected, the result is undefined.

But this source states it works under most architectures:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1817.htm

> maybe you need for first cast it all to uint32_t, then bit_cast to int32_t?

Yep, will do that. Will do it for pressure as well. If the datasheet says I should do so, I have no reason to do otherwise.

examples/nucleo_h723zg/bmp581/main.cpp
joelsa: I am not sure.
This is the standard formula that assumes 15 °C sea level temperature and a temperature reduction of 6.5 °C per km of height increase.
Users that have a good measurement of temperature might use that one, but others have PCBs that get quite hot (like mine) and don't want to use the BMP581s temperature measurement.
Other users are only interested in relative height differences and might use an isothermal approximation.

Is this really something that belongs in the driver?

salkinium: No, these are good points against.

src/modm/driver/pressure/bmp581.hpp
```diff
@@ -0,0 +1,509 @@
...
+	struct Data
+	{
+		/// Get temperature in degrees Celsius
+		float
+		getTemperature() const
+		{
+			const uint32_t u = (static_cast<uint32_t>(rawTemp[2]) << 24) |
+							   (static_cast<uint32_t>(rawTemp[1]) << 16) |
+							   (static_cast<uint32_t>(rawTemp[0]) << 8);
+			return static_cast<float>(static_cast<int32_t>(u)) / (65536.0f * 256.0f);
```
salkinium:
```suggestion
			return static_cast<float>(static_cast<int32_t>(u) / 256) / (65536.0f);
```
This will probably preserve more resolution, since casting high integers to float causes quantization.

salkinium (approved):
Nice!

PR #1340, opened 2026-03-05
[driver] Bosch Sensortec BMI270 IMU

hshose:
BMI270 IMU driver.

- [x] example with interrupts i2c
- [ ]  example with interrupts spi
- [x] fifo buffer example
- [x] component retrim ("CRT") calibration example. There is a bug when writing to the devices non-volatile memory (only 14-ish write cycles per datasheet anyways), so the example is set up to allow for the stm32 to retrieve the retrim values, e.g., to store them in the `flash_reserved` section. @joelsa has more experience with the calibration. I messed around with the vendors offset calibration routine (this calibrates for constant bias and is called "FOC") and applying in the imu's registers, but this was pretty buggy. I therefore recommend to do bias calibration in software (e.g., from an average over motionless values).

Also thanks and credits to @joelsa for explaining the CRT issues to me and building the breakout board I used for testing :tada:


hshose:
rebased fixup commits.

All I2C examples tested in hardware. Unfortunately, I don't have a breakout board to test the SPI example.

examples/nucleo_h723zg/bmi270/i2c/main.cpp
```diff
@@ -0,0 +1,213 @@
...
+std::atomic<int32_t> gyroY{0};
+std::atomic<int32_t> gyroZ{0};
+
+std::atomic_flag imuBusLock = ATOMIC_FLAG_INIT;
+
+bool
+tryLockImuBus()
+{
+	return !imuBusLock.test_and_set(std::memory_order_acquire);
+}
```
salkinium: We have a mutex?
https://github.com/modm-io/modm/blob/d0fcb0df5117bce0a61e25048a9b5c63ff819431/src/modm/processing/fiber/mutex.hpp#L28-L57

examples/nucleo_h723zg/bmi270/i2c/main.cpp
```diff
@@ -0,0 +1,213 @@
...
+		{
+			gyroInterruptCount.fetch_add(1, std::memory_order_relaxed);
+			gyroX.store(data->raw[0], std::memory_order_relaxed);
+			gyroY.store(data->raw[1], std::memory_order_relaxed);
+			gyroZ.store(data->raw[2], std::memory_order_relaxed);
+		}
+	}
+
+	unlockImuBus();
+}
```
salkinium: Yes this works, but blocks the entire scheduler for the duration of the interrupt, which may impact the responsiveness of your app (in this example it's not perceptible).

You could set an atomic bool in this callback and poll on it in dedicated fiber (or the main function).
But since we only support *round-robin* cooperative scheduling, there's no advantage to using an EXTI, you may as well just poll the pin directly.

In future I would like to implement the notify_one/notify_any interface so that fibers waiting on these primitives get scheduled next instead of round-robin. Then the EXTI interface makes sense and would improve latency, but you'd never get as good as an RTOS.

Until then I would simplify this to polling in the main loop tbh.

examples/nucleo_h723zg/bmi270/i2c/main.cpp
```diff
@@ -0,0 +1,213 @@
...
+	return ok;
+}
+
+int
+main()
+{
+	Board::initialize();
+	Leds::setOutput();
+	I2c::connect<Scl::Scl, Sda::Sda>(I2c::PullUps::Internal);
+	I2c::initialize<Board::SystemClock, 1_MHz, 10_pct>();
```
salkinium, line 78: You're not really doing 1MHz I2C with internal pullups?

examples/nucleo_h723zg/bmi270/i2c_fifo/main.cpp
```diff
@@ -0,0 +1,201 @@
...
+using Scl = GpioB8;  // D15
+using Sda = GpioB9;  // D14
+
+using Int1 = GpioD15;  // D8
+using Int2 = GpioF3;   // D9
+
+using Transport = modm::Bmi270I2cTransport<I2c>;
+using Imu = modm::Bmi270<Transport>;
+
+Imu imu{static_cast<uint8_t>(0x68)};
```
salkinium:
```suggestion
Imu imu{0x68u};
```
FYI: We generally have a `Imu::address(options)` functions for this where the options is typically an enum or counter for the addr pins config.

examples/nucleo_h723zg/bmi270/i2c_fifo/main.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+	Exti::connect<Int1>(Exti::Trigger::RisingEdge, [](auto) { onFifoWatermarkInterrupt(); });
+
+	std::array<uint8_t, FifoBufferSize> fifoBuffer{};
+
+	while (true)
+	{
+		if (!fifoWatermarkInterrupt.exchange(false, std::memory_order_acq_rel))
+		{
+			modm::this_fiber::sleep_for(5ms);
```
salkinium, line 140: Why use an EXTI if you then only poll it every 5ms anyways?

examples/nucleo_h723zg/bmi270/spi/main.cpp
```diff
@@ -0,0 +1,218 @@
...
+		{
+			gyroInterruptCount.fetch_add(1, std::memory_order_relaxed);
+			gyroX.store(data->raw[0], std::memory_order_relaxed);
+			gyroY.store(data->raw[1], std::memory_order_relaxed);
+			gyroZ.store(data->raw[2], std::memory_order_relaxed);
+		}
+	}
+
+	unlockImuBus();
+}
```
salkinium, line 95: same issue as above, although the SPI is much faster

src/modm/driver/inertial/bmi270_calibration_impl.hpp
```diff
@@ -0,0 +1,498 @@
...
+		if (ok) { ok &= waitStRunningComplete(); }
+		if (ok) { ok &= evaluateCrtResult(configuredBurstBytes); }
+	}
+
+	restoreOk &= setPowerControl(*savedPowerControl);
+	restoreOk &= setGyroGainEnable(*savedGyroGainEnable);
+	if (apsWasEnabled) { restoreOk &= setAdvancedPowerSave(true); }
+
+	return ok and restoreOk;
+}
```
salkinium, line 463: Yikes. At least the use of lambda functions doesn't spam the implementation full of details functions.

src/modm/driver/inertial/bmi270_transport_impl.hpp
```diff
@@ -0,0 +1,157 @@
...
+{
+	if (count > MaxRegisterSequence) { return {}; }
+
+	modm::this_fiber::poll([&] { return this->acquireMaster(); });
+	Cs::reset();
+
+	txBuffer_[0] = static_cast<uint8_t>(startReg) | ReadFlag;
+	txBuffer_[1] = 0;
+	std::fill_n(txBuffer_.begin() + 2, count, 0);
+	SpiMaster::transfer(txBuffer_.data(), rxBuffer_.data(), count + 2);
```
salkinium:
```suggestion
	SpiMaster::transfer(txBuffer_.data(), nullptr, 2);
	SpiMaster::transfer(nullptr, rxBuffer_.data(), count);
```
I feel like I suggested this before in another driver of yours ;-P

src/modm/driver/inertial/bmi270_transport_impl.hpp
```diff
@@ -0,0 +1,157 @@
...
+Bmi270SpiTransport<SpiMaster, Cs>::writeRegisters(Register startReg, std::span<const uint8_t> data)
+{
+	if (data.size() > MaxRegisterSequence) { return false; }
+
+	modm::this_fiber::poll([&] { return this->acquireMaster(); });
+	Cs::reset();
+
+	txBuffer_[0] = static_cast<uint8_t>(startReg);
+	std::copy(data.begin(), data.end(), txBuffer_.begin() + 1);
+	SpiMaster::transfer(txBuffer_.data(), nullptr, data.size() + 1);
```
salkinium: Same here

src/modm/driver/inertial/bmi270_transport_impl.hpp
```diff
@@ -0,0 +1,157 @@
...
+		uint8_t(static_cast<uint8_t>(Register::FifoData) | ReadFlag), 0u};
+	std::array<uint8_t, 2> discard{};
+	SpiMaster::transfer(header.data(), discard.data(), header.size());
+
+	std::size_t offset = 0;
+	while (offset < data.size())
+	{
+		const std::size_t chunk = std::min<std::size_t>(data.size() - offset, MaxRegisterSequence);
+		std::fill_n(txBuffer_.begin(), chunk, 0);
+		SpiMaster::transfer(txBuffer_.data(), data.data() + offset, chunk);
```
salkinium: same probably

src/modm/driver/inertial/bmi270_transport_impl.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+template<typename I2cMaster>
+bool
+Bmi270I2cTransport<I2cMaster>::writeRegisters(Register startReg, std::span<const uint8_t> data)
+{
+	if (data.size() > MaxRegisterSequence) { return false; }
+
+	buffer_[0] = static_cast<uint8_t>(startReg);
+	std::copy(data.begin(), data.end(), buffer_.begin() + 1);
+	return I2cDevice<I2cMaster>::write(buffer_.data(), data.size() + 1);
```
salkinium: for small data sizes this is ok, but we have an eeprom transaction class somewhere.
We definitely had this discussion before.

salkinium (changes requested):

src/modm/driver/inertial/bmi270_transport_impl.hpp
salkinium: Yes, here: https://github.com/modm-io/modm/pull/1296#discussion_r2539907947

examples/nucleo_h723zg/bmi270/i2c/main.cpp
hshose: image

hshose: Well, I was doing with internal pullups until turning up to 1MHz and forgetting them in the examples :-D

src/modm/driver/inertial/bmi270_calibration_impl.hpp
hshose: yeah, this calibration is ugly and i therefore put it in a separate file. If you prefer to not have it in modm, I can totally understand and keep this specific to our project where we use the IMU.

src/modm/driver/inertial/bmi270_transport_impl.hpp
hshose: Your memory is better than mine :-D I think I looked at the same existing driver again when implementing this. :-S Sorry

src/modm/driver/inertial/bmi270_transport_impl.hpp
hshose: also within the vl53 driver if I'm not mistaken.

hshose:
@salkinium thank you for the review :-)

examples/nucleo_h723zg/bmi270/i2c/main.cpp
joelsa: Even if there are external pull-ups in place, having the very weak internal ones enabled as well doesn't hurt in 99.9% of cases. 

IMO the question for examples is:
Should the internal pull-ups be activated so that users trying out the example get a config that is more likely to work out of the box, or should internal pull-ups be deactivated, so they either use external ones or need to activate the internal ones deliberately and won't get stuck having to debug a somewhat working but flaky setups.

joelsa: @hshose There are DNP'ed 0402 pull-up resistors on the PCB you are using (R114 & R115). So you could get rid of the THT ones if you want to solder something there.

examples/nucleo_h723zg/bmi270/i2c/main.cpp
hshose: Thanks! Good to know, fixed this now.

examples/nucleo_h723zg/bmi270/i2c/main.cpp
hshose: Haha yes, I saw these but didn't want to mess up the use with the Pi CM that has onboard pull-ups.

hshose: I set the pullups to external for the examples and then its explicit what happened here.

examples/nucleo_h723zg/bmi270/i2c_fifo/main.cpp
hshose: Added an I2cAddress struct to the TransportBase. Is this good enough?

src/modm/driver/inertial/bmi270_transport_impl.hpp
hshose: Done

src/modm/driver/inertial/bmi270_transport_impl.hpp
hshose: Done

src/modm/driver/inertial/bmi270_transport_impl.hpp
hshose: Done

src/modm/driver/inertial/bmi270_transport_impl.hpp
hshose: Done

examples/nucleo_h723zg/bmi270/i2c/main.cpp
hshose: Done.

examples/nucleo_h723zg/bmi270/i2c_fifo/main.cpp
hshose: Agreed, here the Exti is useless and I can just poll. Fixed accordingly.

examples/nucleo_h723zg/bmi270/i2c/main.cpp
hshose: ignore all my previous comments on this, the lower update rate without exti that I reported was due to the flimsy jumpers where the interrupt line didn't make good contact.

Changed to simple main loop polling and everything works.

examples/nucleo_h723zg/bmi270/i2c/main.cpp
hshose: see comment below, only polling in the example now.

salkinium (approved):
Awesome, thanks!

PR #1350, opened 2026-04-23
[docs] Revert back to Doxygen for API docs generation

salkinium:
Doxypress is unmaintained and gets stuck on both macos and Linux. In the CI we have to use the ancient version for this reason.
We'll have to switch this soon too.


rleh (approved):
Thanks!

chris-durand (approved):
Thank you!

PR #1351, opened 2026-04-24
fix: stop-token not resetted on start

marinauterion:
after stopping a fiber with stoken
and then starting it again, the stop-token was still set


salkinium (approved):
Thanks!

PR #1353, opened 2026-05-11
Fix BMI088 driver getting stuck + GenericTimeout::wait() behavior

### LINKED ISSUE #1352: BMI088 driver getting stuck during initialisation
The last change made to the driver was a fiber refactoring:

```
commit 99c69b519467e6b8704e41d7d5ee7ec713ccd478
Author: Niklas Hauser <niklas.hauser@rwth-aachen.de>
Date:   Sat Aug 23 20:31:21 2025 +0200

    [driver] Replace Timeout with this_fiber::sleep_for
```
The use of the following function was replaced by `modm::Timeout::wait()`:

```cpp
template<Bmi088Transport Transport>
void
Bmi088<Transport>::timerWait()
{
   while (timer_.isArmed()) {
       modm::this_fiber::yield();
   }
}
```

I could determine with the debugger that the code always got stuck here:

```cpp
template<Bmi088Transport Transport>
bool
Bmi088<Transport>::setAccRange(AccRange range)
{
	timer_.wait();
```

The culprit is the definition of the `wait()` method:
```cpp
template< class Clock, class Duration >
void
modm::GenericTimeout<Clock, Duration>::wait()
{
	modm::this_fiber::poll([this]{ return execute(); });
}
```

It is based on `execute()` which will only return true on its first invocation. Thus, two subsequent calls to `wait()` on an expired timer will block indefinitely. The same happens if the timer hasn't been started yet.

I'd change the behaviour of `wait()` to poll for `!isArmed()`. If there is something to wait for, it'll wait or otherwise just return instead of blocking forever.

@salkinium Would you be fine with that change?

chris-durand:
Fix the BMI088 driver getting stuck during initialization. This was caused by a refactoring using `GenericTimeout::wait()`.

The timer implementation is changed to wait as long as the timer is armed instead of waiting for `execute()` to return true. This prevents several edge cases where the wait function blocks indefinitely.

Fixes #1352

### DIFF
--- a/src/modm/processing/timer/timeout_impl.hpp
+++ b/src/modm/processing/timer/timeout_impl.hpp
@@ -130,7 +130,7 @@ template< class Clock, class Duration >
 void
 modm::GenericTimeout<Clock, Duration>::wait()
 {
-	modm::this_fiber::poll([this]{ return execute(); });
+	modm::this_fiber::poll([this]{ return !isArmed(); });
 }
 
 // ----------------------------------------------------------------------------
--- a/test/modm/processing/timer/timeout_test.cpp
+++ b/test/modm/processing/timer/timeout_test.cpp
@@ -3,6 +3,7 @@
  * Copyright (c) 2009-2010, 2012, 2018, Fabian Greif
  * Copyright (c) 2012, 2014-2015, 2017, 2020, Niklas Hauser
  * Copyright (c) 2013, 2016, Sascha Schade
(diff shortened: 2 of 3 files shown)


chris-durand:
I'll also add a unit test tomorrow

chris-durand:
@salkinium I've added the test

salkinium (approved):
Thank you very much!

PR #1362, opened 2026-07-02
[RP] RP2040 Flash support

andryblack:
Added access to masked ROM functions.
Added functionality for in program flash access.
Example for working with flash.

Because XIP must be disabled to perform flash operations, any code interacting with the flash cannot reside within the flash memory itself. Utilizing functions from ROM allows for a reduction in RAM consumption by this functionality.

### DIFF
--- a/src/modm/platform/core/rp/rom.hpp
+++ b/src/modm/platform/core/rp/rom.hpp
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) 2026, Andrey Kunitsyn
+ *
+ * This file is part of the modm project.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+// ----------------------------------------------------------------------------
+#pragma once
+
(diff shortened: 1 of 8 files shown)


src/modm/platform/core/rp/rom.hpp
salkinium, line 1: Instead of generating a basically empty file, we should just not generate the file on family != 20.

src/modm/platform/flash/rp/flash.cpp
```diff
@@ -0,0 +1,236 @@
...
+static bool boot2_copyout_valid = false;
+
+struct flash_hardware_save_state_t
+{
+	uint32_t qspi_pads[NUM_QSPI_GPIOS];
+};
+
+modm_ramcode static void
+xip_cache_clean_all()
+{}
```
salkinium: is this intended as a customization point ie. overwriteable by the application? Then it should be non-static `modm_weak`

salkinium (changes requested):
Very nice, thank you!

src/modm/platform/flash/rp/flash.cpp
andryblack: no, this is rp30 artifact, need remove

salkinium (approved):
Thanks!

PR #1366, opened 2026-07-07
[stm32] H7 ethernet and example

hshose:
Ethernet support with FreeRTOS for H7 devices.

Tested in hardware on NUCLEO-H723ZG.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -900,32 +900,34 @@ We have out-of-box support for many development boards including documentation.
 </tr><tr>
 <td align="center"><a href="https://modm.io/reference/config/modm-nucleo-l496zg-p">NUCLEO-L496ZG-P</a></td>
 <td align="center"><a href="https://modm.io/reference/config/modm-nucleo-l552ze-q">NUCLEO-L552ZE-Q</a></td>
+<td align="center"><a href="https://modm.io/reference/config/modm-nucleo-n657x0-q">NUCLEO-N657X0-Q</a></td>
 <td align="center"><a href="https://modm.io/reference/config/modm-nucleo-u083rc">NUCLEO-U083RC</a></td>
-<td align="center"><a href="https://modm.io/reference/config/modm-nucleo-u385rg-q">NUCLEO-U385RG-Q</a></td>
 </tr><tr>
+<td align="center"><a href="https://modm.io/reference/config/modm-nucleo-u385rg-q">NUCLEO-U385RG-Q</a></td>
 <td align="center"><a href="https://modm.io/reference/config/modm-nucleo-u575zi-q">NUCLEO-U575ZI-Q</a></td>
 <td align="center"><a href="https://modm.io/reference/config/modm-olimexino-stm32">OLIMEXINO-STM32</a></td>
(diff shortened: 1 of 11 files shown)


src/modm/platform/eth/stm32/eth.hpp
```diff
@@ -18,8 +18,20 @@
...
+	defined(STM32H730xx) || defined(STM32H730xxQ) || defined(STM32H733xx) || \
+	defined(STM32H735xx) || defined(STM32H742xx) || defined(STM32H743xx) || \
+	defined(STM32H745xx) || defined(STM32H745xG) || defined(STM32H747xx) || \
+	defined(STM32H747xG) || defined(STM32H750xx) || defined(STM32H753xx) || \
+	defined(STM32H755xx) || defined(STM32H757xx) || defined(STM32H7A3xx) || \
+	defined(STM32H7A3xxQ) || defined(STM32H7B0xx) || defined(STM32H7B0xxQ) || \
+	defined(STM32H7B3xx) || defined(STM32H7B3xxQ)
+#	define MODM_ETH_STM32H7
+#endif
+
```
salkinium: No? Just copy the file and choose it at lbuild time.

hshose: sure.

README.md
salkinium, line 1: Something is broken with this file, maybe you're using an old local modm-devices?

salkinium (commented):
I'm a bit unhappy that we're going to FreeRTOS instead of integrating lwip in fibers, because  I don't really want to maintain two HALs for cooperative and preemtive scheduling, that just seems like the worst of both worlds.

But for now, yeah ok, fine.

hshose:
Will fix the modm-devices issue that is indeed a relic from my (failed) attempt to implement support for the cursed N6.

Also fine with me if this doesn't get merged if there is a plan to support lwip soon?

I don't like the FreeRTOS.

salkinium:
Oh yeah, I looked at the N6 and just said nope.

I'm currently [working on a board](https://chaos.social/@salkinium/116806033771936525) that also has a LAN8742A chip but connected to a STM32H5, so there's at least a use case for me to integrate lwip. But it's not going to be soon, maybe by the end of the year.

I would merge this as is as it solves a problem, and with lbuild it's not in the way of anything else.

hshose:
Split example and driver, rebased current develop

WasabiFan:
Just came across the thread and figured I'd note... I have a lwip wrapper implementation of Ethernet for the H7 that I've been using locally. Doesn't have a dependency on FreeRTOS. Invokes the standard lwip periodic callbacks in a fiber. I pulled in lwip as a submodule. I have only been using UDP and TCP, no higher network layers. UDP can saturate the 100Mbit link full duplex although I've been recently working on reducing copy overhead. Tested only on H753+LAN8742A, haven't implemented any other PHYs. I've kept it private rather than opening a PR because there's some AI-generated code that I don't want to burden maintainers with until I properly review it myself. But I would be happy to spend some time reviewing/cleaning it up and post a PR if folks are interested.

salkinium:
Yes please! You seem to not be the only one, @rleh has created a partial repo of lwip already here: https://github.com/modm-ext/lwip-partial

PR #1369, opened 2026-07-21
[stm32] Fix STM32H5 32-bit timer counter size

maschull106:
TIM2 and TIM5 are 32-bit timers on STM32H5, but `general_purpose.hpp.in` omitted `"h5"` from the family whitelist used to select `TimerN::Value`.

- [x] Tested on NUCLEO-H563ZI


salkinium (approved):
Thanks!

PR #1374, opened 2026-08-07
[build] Fix STM32 build with GCC15

chris-durand:
Allow compilation of STM32 GPIO module with GCC15 by fixing ODR violation due to missing `inline` on variable

GCC 15.2 is crashing trying to build CMSIS-DSP, but the issue is fixed in 15.3.


salkinium (approved):
Thanks!

PR #1376, opened 2026-08-25
add H5 variant to H7 spi driver

hajue-schmitz:
add spi support for STM32H5
add sai clock setting for H5
make changes for I2cClockSource because the H503 uses PLL2R other H5 uses PLL3R


salkinium:
Did you not want to continue this PR? This sounds very useful.

hajue-schmitz:
Yes. I'll fix some merge failure and add a unit test.
I'm also finish testing the adc-driver for the H5 and also working on the GDMA driver.
This works actually with the adc in circular mode.

